PR #3, opened 2015-09-29
Verbosify error messages in databroker.py

ericdill:
Also hit all the error conditions in the tests


danielballan:
This is helpful. +1 for more test coverage!

PR #4, opened 2015-09-29
WIP: Improving test coverage, wheee

ericdill:


tacaswell:
Don't we need `databroker/utils/console.py` for the colored text in `summarize`?

danielballan:
I thought I remembered that but @ericdill says it's nowhere in the repo or other repos.

tacaswell:
It was deleted here: https://github.com/NSLS-II/dataportal/pull/206/files#diff-a99b236f5625bb5a279fbd68f5993fccL399 with the verbal promise it would come back someday.

danielballan:
Ah, indeed. Now that promise is a written one. :- ) I think it might belong in a different repo anyway.

PR #5, opened 2015-09-29
Add 'scan_type' to the RunStart document.

ericdill:
Same as NSLS-II/metadatastore#190

### DIFF
--- a/databroker/examples/sample_data/image_and_scalar.py
+++ b/databroker/examples/sample_data/image_and_scalar.py
@@ -13,9 +13,9 @@ import argparse
 
 
 # number of motor positions to fake
-num1 = 20
+num_motor_positions = 50
 # number of temperatures to record per motor position
-num2 = 10
+num_temps_per_motor_position = 10
 
 # These are imported from MDS and cannot be imported as things.
 noisy = common.noisy
@@ -23,13 +23,17 @@ example = common.example
 
 # This section sets up what the simulated images will look like.
 
-img_size = (500, 500)
+# never use a square image for test purposes!
+img_size = (150, 175)
+
 period = 150
 
 
 def I_func_sin(count):
     return (1 + .5*np.sin(2 * count * np.pi / period))
-center = 25
+
+# never put the center of the test pattern in the center of the data set!
+center = num_motor_positions / 2.5
 sigma = center / 4
 
 
@@ -80,7 +84,7 @@ def run(run_start_uid=None, sleep=0):
         uid=str(uuid.uuid4()))
 
     events = []
-    for idx1, i in enumerate(range(num1)):
+    for idx1, i in enumerate(range(num_motor_positions)):
         img = next(frame_generator)
         img_sum = float(img.sum())
         img_sum_x = img.sum(axis=0)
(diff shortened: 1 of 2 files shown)


tacaswell:
closed/opened to re-start travis based on master now, not when it was created

```python
======================================================================
ERROR: databroker.tests.test_misc.test_pims_images_old_api
----------------------------------------------------------------------
  File "/home/travis/mc/envs/testenv/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/build/NSLS-II/databroker/databroker/tests/test_misc.py", line 21, in test_pims_images_old_api
    assert_equal(len(images), image_and_scalar.num1)
AttributeError: module 'databroker.examples.sample_data.image_and_scalar' has no attribute 'num1'
======================================================================
ERROR: databroker.tests.test_misc.test_pims_images
----------------------------------------------------------------------
  File "/home/travis/mc/envs/testenv/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/build/NSLS-II/databroker/databroker/tests/test_misc.py", line 30, in test_pims_images
    assert_equal(len(images), image_and_scalar.num1)
AttributeError: module 'databroker.examples.sample_data.image_and_scalar' has no attribute 'num1'
```

databroker/examples/sample_data/image_and_scalar.py
```diff
@@ -13,9 +13,9 @@
 
 
 # number of motor positions to fake
-num1 = 20
```
tacaswell: this is referenced in the tests.  Either update the tests or put an alias back in.

danielballan:
There is a rename coming on scan_type. Moving to the word plan, with input from @klauer -- and the idea that each run involves scanning something is not true across the board.

danielballan:
@ericdill If I were you I would just go with using bluesky to write the test data, but if you'd rather see this merged change `scan_type` -> `plan_type`.

ericdill:
@danielballan I like the idea of using bluesky to write test data.  Any objections to moving (and simplifying) the example code for generating data into bluesky from metadatastore, bluesky and filestore? I'm all for separation of concerns, but it seems like the example generation code is just always lagging behind

danielballan:
Off the top of my head, I think that sounds fine. To be clear, that would make bluesky a test-only dependency of these projects. We have been trying the alternative -- imperfectly imitating bluesky's output by hand -- for awhile now and I think it's clear that it is worse. But I would want a group consensus on that....

cc @klauer @tacaswell @dchabot

klauer:
Seems to make more sense generating it with bluesky. But what package dependency implications would that have?

danielballan:
@ericdill I think you are using the word "test dependency" to mean "test dependency listed in the conda package." I was using it to mean "package you need to install if you want to run the tests." So I want to make sure that everyone is OK that bluesky would become *de facto* required for any developer working on metadatastore.

ericdill:
> @ericdill I think you are using the word "test dependency" to mean "test dependency listed in the conda package." 

Yes, I was attempting to answer the question of 'what would `conda install metadatastore` do', but I guess I'm not entirely sure that *was* the question that @klauer was even asking. Oh well.

> I was using it to mean "package you need to install if you want to run the tests." So I want to make sure that everyone is OK that bluesky would become de facto required for any developer working on metadatastore.

Yes, I understand.  To further hammer on this idea, since the bluesky RunEngine is currently the only way to get experimental data into the databroker stack, I am like one million :+1: on making bluesky the only way to get test data into the testing databroker stack so we don't have to keep twiddling the data generation code in mds/db.

danielballan:
Ah, I missed @klauer's question. Developers working on metadatastore/filestore/databroker would have to install bluesky and its dependencies -- which, thanks to the callbacks, are pretty sprawling right now. But they would not need bluesky's *test* dependencies (like ophyd).

ericdill:
closing old prs

PR #6, opened 2015-09-30
MNT: Install pymongo2.6.3 on travis

ericdill:
- if travis passes, this is good to go


ericdill:
This should be pinned to pymongo = 2.9 for now

PR #8, opened 2015-10-13
BLD: Specify install_requires

danielballan:


ericdill:
We probably need to get filestore/metadatastore up on pypi before this gets merged.

`conda build conda-recipe` will fail unless you change the build.sh to be `$PYTHON setup.py install --single-version-externally-managed --record=/dev/null`.  That is discussed: https://groups.google.com/a/continuum.io/forum/#!topic/conda/ZKdP5BujriA

PR #9, opened 2015-10-13
FIX: Time column had mismatched index.

danielballan:
attn @ericdill


ericdill:
Is this the sort of bug fix that should also get a test associated with it?

danielballan:
Fair point. Test added. Previous Travis build failed on a mongo connection error. Presumably fixed by your recent pymongo work, I think. We'll see.

ericdill:
Awesome, thanks for the test.

PR #10, opened 2015-10-13
Use pymongo 2.9 on travis

ericdill:
This will fix the errors that travis is spewing


ericdill:
cc @danielballan

databroker/examples/sample_data/frame_generators.py
```diff
@@ -59,7 +59,7 @@ def func(step, count):
         raise ValueError("image shape must be 2 dimensional "
                          "you passed in {}".format(im_shape))
 
-    cur_position = np.asarray(im_shape) // 2
+    cur_position = np.asarray(im_shape) / 2
```
danielballan: For posterity and @tacaswell: this fixes a numpy 1.10 issue, but there are more issues in the dependencies themselves (looking at you, pandas) so we are pinning Travis for now.

PR #12, opened 2015-11-04
ENH: first pass at pivoting event -> event stream

tacaswell:
This converts an event stream of timeseries to an event-stream of events
with the timeseries entries 'pivoted' up to each be an event.

### DIFF
--- a/databroker/pivot.py
+++ b/databroker/pivot.py
@@ -1,0 +1,93 @@
+from itertools import chain
+import uuid
+import time as ttime
+
+def pivot_timeseries(events, pivot_keys, static_keys=None):
+    """Pivot sequence of events with sequences as data to single sequence
+
+    For example a sequence of 5 events with 100 frames of 2D data per datum
+    will turn into a sequence of 500 events which each have a single frame.
+
+    Any keys not in ``pivot_keys`` or ``static_keys`` will be dropped.
+
+    This will return new event documents with a new Descriptor which is
+    not in the database.
+
+
+    Parameters
+    ----------
+    events : sequence
+        A sequence of events all with the same Descriptor
+
+    pivot_keys : sequence
+        A list of keys from the events to pivot out
+
+    static_keys : sequence, optional
+        A list of keys from the original events to propagate
+        to the new events
+
+    Yields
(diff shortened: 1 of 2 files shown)


databroker/pivot.py
```diff
@@ -0,0 +1,204 @@
...
+        raise RuntimeError("Pivot and static_keys overlap")
+
+    events = iter(events)
+    ev0 = next(events)
+    orig_desc = ev0['descriptor']
+    new_desc = {'uid': str(uuid.uuid4()),
+                'data_keys': dict(),
+                'run_start': orig_desc['run_start'],
+                'time': ttime.time()}
+    new_desc['data_keys']['fr_no'] = {'shape': (),
```
danielballan: I don't love `fr_no`.

databroker/pivot.py
```diff
@@ -0,0 +1,204 @@
...
+            inner_data['fr_no'] = fr_no
+            inner_ts = dict(static_ts)
+            inner_ts['fr_no'] = ttime.time()
+            new_ev['data'] = inner_data
+            new_ev['timestamps'] = inner_ts
+            yield new_ev
+            seq_no += 1
+
+
+def zip_events(*evs, **kwargs):
```
danielballan: Should `**kwargs` be `lazy=True`?

databroker/pivot.py
```diff
@@ -0,0 +1,204 @@
...
+    Parameters
+    ----------
+    *evs
+        sequences that yield events
+
+    lazy : bool, optional
+        If False, listify event streams first for validation
+        defaults to True
+    """
+    # BEGIN LPy
```
danielballan: LPy -> Legacy Python -> Python2 is the @tacaswell equivalent of Cockney rhyming slang.

databroker/pivot.py
tacaswell: No, because in Lpy you can not do *args and then a keyword argument.

tacaswell: See 25b3137a74883d5c9371408251958f2d1d7629f1 and the failure on the commit before it.

PR #13, opened 2015-11-04
Use single dispatch for databroker getitem

ericdill:
This provides more flexibility at no cost.  We can, at any time, extend
the functionality of the databroker to handle new argument types to the
db[] notation by using singledispatch

Blocking Issues
----------------
- [x] Test conda-recipe on python 2.7

### DIFF
--- a/databroker/databroker.py
+++ b/databroker/databroker.py
@@ -25,7 +25,11 @@ except ImportError:
         )
 
 import numbers
-from collections import abc
+try:
+    from collections.abc import MutableSequence
+except ImportError:
+    # This will error on python < 3.3
+    from collections import MutableSequence
 
 
 logger = logging.getLogger(__name__)
@@ -118,7 +122,7 @@ def _(key):
 
 @search.register(set)
 @search.register(tuple)
-@search.register(abc.MutableSequence)
+@search.register(MutableSequence)
 def _(key):
     return [search(k) for k in key]
 
--- a/databroker/databroker.py
+++ b/databroker/databroker.py
@@ -11,7 +11,18 @@ import metadatastore.doc as doc
 import metadatastore.commands as mc
 import filestore.api as fs
 import logging
-from functools import singledispatch
+try:
+    from functools import singledispatch
+except ImportError:
+    try:
+        # We are running on Python 2.6, 2.7, or 3.3
+        import singledispatch
+    except ImportError:
+        print("Please install singledispatch from PyPI"
+              "\n\n   pip install single dispatch"
+              "\n\nThen run your program again.")
+        raise
+
 import numbers
 from collections import abc
 
--- a/databroker/databroker.py
+++ b/databroker/databroker.py
@@ -11,96 +11,115 @@ import metadatastore.doc as doc
 import metadatastore.commands as mc
 import filestore.api as fs
 import logging
+from functools import singledispatch
+import numbers
+from collections import abc
 
 
 logger = logging.getLogger(__name__)
 TZ = str(tzlocal.get_localzone())
 
 
+@singledispatch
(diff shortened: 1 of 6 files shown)


danielballan:
:party: :balloon:

ericdill:
singledispatch was added in python 3.4, but there is a backport on pypi that makes this functionality available on python 2.6-3.3.  If this PR is accepted, I can update the meta.yaml and travis.yml to install singledispatch on those platforms

danielballan:
Just discovered that myself, so here's the link: https://pypi.python.org/pypi/singledispatch

databroker/databroker.py
```diff
@@ -11,96 +11,127 @@
...
 import logging
+try:
+    from functools import singledispatch
+except ImportError:
+    try:
+        # We are running on Python 2.6, 2.7, or 3.3
+        import singledispatch
+    except ImportError:
+        print("Please install singledispatch from PyPI"
+              "\n\n   pip install single dispatch"
```
tacaswell: extra space in singledispatch

tacaswell:
nifty!

databroker/databroker.py
ericdill: Ah good catch

ericdill:
Note that I have not yet tested my modifications to the conda recipe on Python 2.x or 3.3 yet, so I am not sure that it **actually** works on those platforms.

tacaswell:
I would only worry about 2.7.  IPython no longer supports 2.6 and mpl no longer support 3.3

ericdill:
Huh. I totally forgot about this PR.  I'll resurrect it tomorrow

ericdill:
Some light testing on my mac suggests that the updates to the conda-recipe work just fine.  It would be nice to have mds/fs support pymongo 3+ so that we do not need to pin to less than 3 though.  I can start looking in to this, unless @tacaswell wants to sort this out?

ericdill:
Either way, this PR is good to go. Can I get someone's eyes and a final approval?

databroker/databroker.py
```diff
@@ -11,96 +11,133 @@
...
+            try:
+                result = next(gen)
+            except StopIteration:
+                raise IndexError(
+                    "There are only {0} runs.".format(i))
+        header = Header.from_run_start(result)
+    return header
+
+
+@search.register(str)
```
ericdill: Should this be `six.text_types` or `six.text_type` instead of `str`?

danielballan: I *think* this is fine as is. @tacaswell?

tacaswell: that will fail on unicode in LPy

ericdill:
Other than codecov being mad at me, this should be ready to go.

databroker/databroker.py
```diff
@@ -11,96 +11,144 @@
...
+                         "support slicing infinitely into the past; "
+                         "the size of the result is non-deterministic "
+                         "and could become too large.")
+    start = -key.start
+    result = list(find_last(start))[stop::key.step]
+    header = [Header.from_run_start(h) for h in result]
+    return header
+
+
+@search.register(numbers.Integral)
```
danielballan: What's the difference between this and `int`? Just curious -- haven't seen this before.

ericdill: It is a pattern that was suggested in [Fluent Python] (http://shop.oreilly.com/product/0636920032519.do) and seems to be reasonably well explained [here] (http://stackoverflow.com/questions/4281955/whats-the-relationship-between-numbers-integral-and-int-in-builtins-module-in-p).

Basically, numbers.Integral is the interface that defines the `int` builtin

PR #15, opened 2015-12-04
Cool new explorer widget

ericdill:
cc @danielballan @tacaswell

### DIFF
--- a/databroker/widgets/explorer.py
+++ b/databroker/widgets/explorer.py
@@ -3,7 +3,11 @@ from ipywidgets import (interact, Text, Dropdown, Checkbox, VBox, HBox,
                         SelectMultiple, ToggleButtons, Select)
 import matplotlib.pyplot as plt
 from collections import namedtuple
-from traitlets import Tuple, Unicode, HasTraits, Set, Int, observe, link
+from traitlets import Tuple, Unicode, HasTraits, Set, Int, link
+
+import logging
+logger = logging.getLogger(__name__)
+
 
 
 def explorer():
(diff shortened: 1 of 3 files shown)


danielballan:
Yeah, this looks good.

danielballan:
I'll let @tacaswell sign off in case there are mpl subtleties I don't see.

ericdill:
How should I test this thing? I guess just poke the traits that back thr widgets and make sure they respond correctly...?  Is there any guidelines for CI testing of ipywidgets?

danielballan:
Yeah, that's what I would do.

databroker/widgets/explorer.py
```diff
@@ -0,0 +1,186 @@
...
+    key_select.on_trait_change(scan_submit, 'value')
+
+    widgets = {'x_dropdown': x_dropdown,
+               'y_select': y_select,
+               'scan_select': scan_select,
+               'fill_checkbox': fill_checkbox,
+               'key_select': key_select,
+               'scan_text': scan_text}
+    box = VBox([fill_checkbox, HBox([scan_text, key_select]), scan_select, x_dropdown, y_select])
+    return namedtuple('explorer_return_values', ['widgets', 'display'])(widgets, box)
```
tacaswell: Probably worth creating the named tuple once rather than every time.

tacaswell:
Sorry, started to review this and ended up down a rabbit hole which may result in a PR to pandas....

danielballan:
I am taking the "WIP" commit to mean, "stand by, this isn't quite ready more review." Tag me when this is ready. :- )

danielballan:
Let's get this in. What remains to be done, @ericdill?

ericdill:
There are no tests on the backend logic of the widget.  That being said, I'm not going to be able to write said tests for a while.  There are other, higher priority issues that need to be dealt with (like vertical integration testing!)  Let me rebase it (the conflict is likely in the meta.yaml) and then make sure it still runs.  Have you tried to run it on your local machine?

PR #19, opened 2016-01-23
ENH: Implement a drop-in replacement for RunEngine Document Dispatcher.

danielballan:
We documented this but never actually wrote it.

Having trouble running tests locally due to pip problems + pymongo thrashing.

### DIFF
--- a/databroker/databroker.py
+++ b/databroker/databroker.py
@@ -423,3 +423,45 @@ def get_table(headers, fields=None, fill=True, convert_times=True):
     else:
         # edge case: no data
         return pd.DataFrame()
+
+
+def stream(headers, fields=None, fill=True):
+    """
+    Get all Documents from given run(s).
+
+
+    Parameters
+    ----------
+    headers : Header or iterable of Headers
+        The headers to fetch the events for
+    fields : list, optional
+        whitelist of field names of interest; if None, all are returned
+    fill : bool, optional
+        Whether externally-stored data should be filled in. Defaults to True
+
+    Yields
+    ------
+    name, doc : tuple
+        string name of the Document type and the Document itself.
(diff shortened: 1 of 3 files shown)


databroker/databroker.py
```diff
@@ -423,3 +423,45 @@ def get_table(headers, fields=None, fill=True, convert_times=True):
...
+    headers : Header or iterable of Headers
+        The headers to fetch the events for
+    fields : list, optional
+        whitelist of field names of interest; if None, all are returned
+    fill : bool, optional
+        Whether externally-stored data should be filled in. Defaults to True
+
+    Yields
+    ------
+    doc : Document
```
tacaswell: I think this should `yield name, doc` as that is the signature that the call backs expect.

danielballan: Oh, duh. Fully agree!

databroker/tests/test_broker.py
```diff
@@ -242,3 +242,22 @@ def test_raise_conditions():
     ]
     for raiser in raising_keys:
         yield _raiser_helper, raiser
+
+
+def test_stream():
+    s = stream(db[-1])
```
tacaswell: This isn't safe for testing.  We really need to sort out py.test paramaterized tests.

ericdill: There is a module level `setup()` function that gets run first, so the broker will always have access to a header

tacaswell: This assumes a one descriptor event stream

danielballan: > We really need to sort out py.test paramaterized tests.

... in this PR?

tacaswell: Or just insert a data stream and retrieve it for this test.

PR #20, opened 2016-01-23
DOC: overhaul

danielballan:
- Remove usage examples that are more focused on DataMuxer usage.
- Add simple examples to index page.
- Split "API docs" into more obviously-titled "Searching for Data" and "Fetching Data".
- Add section explaining scan ID vs. uid (yesterday I learned that even CSX is still confused about this, so we docs on this we can point people to).


danielballan:
Built docs are already pushed [here](http://nsls-ii.github.io/databroker/)

doc/source/index.rst
```diff
@@ -13,11 +13,44 @@ built-in Python data types and numpy arrays.
...
+
+If they are not already imported, import ``DataBroker`` and its convenience
+fucntions. (This step may already be done as part of an
+`IPython profile <http://nsls-ii.github.io/beamline-configuration.html#ipython-profiles>`_.)
+
+.. code-block:: python
+
+    from databroker import DataBroker as db, get_table, get_images, get_events
+
+Example 1: Load scalar data in a table.
```
ericdill: I appreciate the simplicity of the example.  Do you want to mention the `fill` kwarg which defaults to filling in external data? Maybe change the example to `get_table(last_run, fill=False)`?

ericdill:
Just a minor quibble about the `scalar` language in the first example.  Otherwise LGTM!

danielballan:
Good suggestion, @ericdill. Revised and self-merged.

PR #21, opened 2016-01-23
Update badges in readme

ericdill:


danielballan:
While we're here, can you tack on a commit to link to the docs?

ericdill:
Ah, good idea.  Added a link to the broker-specific docs and the overall docs

PR #23, opened 2016-01-30
ENH: KeyError to catch probably spelling mistake

danielballan:
`get_images(header, 'nonexistent_key')` dies somewhere down in filestore/pims with an error saying something about iteration. This confused me for a couple minutes, so I guess we ought to provide a better error.

This PR raises a KeyError if the internal call to `get_events(header, ['nonexistent_key'])` comes back empty.


danielballan:
Superceded by #24

PR #24, opened 2016-01-31
MNT/API: raise if field does not exist

tacaswell:
If a given field does not exist in at least one destriptor for every
header, raise.

### DIFF
--- a/databroker/databroker.py
+++ b/databroker/databroker.py
@@ -347,7 +347,7 @@ def get_events(headers, fields=None, fill=True):
                                  format(k, header['run_star']))
 
     for header in headers:
-        for descriptor in header.desriptors:
+        for descriptor in header.descriptors:
             all_fields = set(descriptor['data_keys'])
             if fields:
                 discard_fields = all_fields - fields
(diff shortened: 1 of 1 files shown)


danielballan:
Hmm. I disagree that this is desired behavior. Better to catch it downstream (or not) as in https://github.com/NSLS-II/databroker/pull/23/files

danielballan:
Example of why this does more harm than good: if I poll during partial data collection, now I have to catch this error.

tacaswell:
Can change it to a KeyError, but the semantics seem wrong.  In #23 you still have to catch an exception, here it is just descriptive and prompt.

danielballan:
In #23 an empty result set becomes an error because PIMS doesn't know how to handle a sequence of length 0. (This is because it inspects the first element to glean shape, dtype.) Other downstream consumers might not have this limitation. In general, an empty collection is often a valid input.

If you are unmoved by that argument, I'll concede this one.

danielballan:
In an offline conversation, I am basically convinced that this PR is the lesser evil.

PR #26, opened 2016-02-04
BLD: Use .get() instead of directly accessing env

ericdill:
conda build master is now using a two-pass jinja system which means
that environ should be accessed via .get() instead of [] if you want
to access GIT_* variables.


conda-recipe/meta.yaml
```diff
@@ -1,6 +1,6 @@
 package:
   name: databroker
-  version: {{ environ['GIT_DESCRIBE_TAG'] }}.post{{ environ['GIT_DESCRIBE_NUMBER'] }}
+  version: {{ environ.get('GIT_DESCRIBE_TAG', 'GIT_STUB') }}.post{{ environ.get('GIT_DESCRIBE_NUMBER', 'inf') }}
```
tacaswell: This needs [1:] to strip the leading 'v'

PR #28, opened 2016-02-10
Add get_fields convenience function

danielballan:
@tacaswell and I separately needed something like this today, so it's probably a useful thing to ship with databroker.

I also added some belated docs from the `stream` function that was added earlier today.

### DIFF
--- a/databroker/databroker.py
+++ b/databroker/databroker.py
@@ -445,6 +445,15 @@ def stream(headers, fields=None, fill=True):
         string name of the Document type and the Document itself.
         Example: ('start', {'time': ..., ...})
 
+    Example
+    -------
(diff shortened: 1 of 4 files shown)


databroker/databroker.py
```diff
@@ -465,3 +474,22 @@ def stream(headers, fields=None, fill=True):
...
+
+    Parameters
+    ----------
+    header : Header
+
+    Returns
+    -------
+    fields : set
+    """
+    fields = set()
```
tacaswell: `return set(chain(d['data_keys'].keys() for d in header['descritpors']))` :imp:

danielballan: Well :hankey:

PR #30, opened 2016-02-10
MNT: Use @tacaswell's set comprehension

ericdill:
Less lines are better!

### DIFF
--- a/databroker/databroker.py
+++ b/databroker/databroker.py
@@ -500,8 +500,4 @@ def get_fields(header):
     -------
     fields : set
     """
-    fields = set()
-    for desc in header['descriptors']:
-        for field in desc['data_keys'].keys():
-            fields.add(field)
-    return fields
+    return set(chain(d['data_keys'].keys() for d in header['descriptors']))
--- a/databroker/databroker.py
+++ b/databroker/databroker.py
@@ -500,4 +500,4 @@ def get_fields(header):
     -------
     fields : set
     """
-    return set([key for desc in h['descriptors'] for key in desc['data_keys']])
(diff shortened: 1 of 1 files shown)


tacaswell:
you need to import chain from itertools

ericdill:
Ah right. Thanks @tacaswell

databroker/databroker.py
```diff
@@ -500,8 +501,4 @@ def get_fields(header):
     -------
     fields : set
     """
-    fields = set()
-    for desc in header['descriptors']:
-        for field in desc['data_keys'].keys():
-            fields.add(field)
-    return fields
+    return set(chain(list(d['data_keys'].keys()) for d in header['descriptors']))
```
danielballan: If we're bikeshedding implementations, I believe `list` isn't needed here.

tacaswell: or actually `keys()`.

danielballan: Ha. Awesome.

ericdill: I'm not sure `chain` is the right approach here.

```python
>>> from itertools import chain
>>> list(chain([[1, 2, 3], [2, 3, 4]]))
[[1, 2, 3], [2, 3, 4]]
>>> set(list(chain([[1, 2, 3], [2, 3, 4]])))
  File "<stdin>", line 1, in <module>
TypeError: unhashable type: 'list'
```

But since bikeshedding is fun, lets just do a double list comprehension

danielballan: @ericdill I think you're just missing a `*` there. Also, `list` is not necessary.

```python
In [13]: set(chain(*[[1, 2, 3], [2, 3, 4]]))
Out[13]: {1, 2, 3, 4}
```

PR #31, opened 2016-02-11
ENH: Add process; rename stream -> restream (with back-compat)

danielballan:
A primary use case for the recently-introduced `restream` (formerly called `stream`) is a list comprehension:

```python
docs = restream(header)  # a generator
[some_callback(name, doc) for name, doc in docs]
```

This PR introduces a convenience function to make that easy.

```python
process(header, some_callback)
```

Perhaps we should accept a list of callbacks or, as in bluesky, the dict `{'all': [...], 'start': [...], ...}` routing documents by type. But I think it's best to keep it simple here. If users need to use multiple callbacks, they can always roll their own dispatching function or use `CallbackBase`.

### DIFF
--- a/databroker/databroker.py
+++ b/databroker/databroker.py
@@ -437,15 +437,14 @@ def get_table(headers, fields=None, fill=True, convert_times=True):
         return pd.DataFrame()
 
 
-def stream(headers, fields=None, fill=True):
+def restream(headers, fields=None, fill=True):
     """
     Get all Documents from given run(s).
-
 
     Parameters
     ----------
     headers : Header or iterable of Headers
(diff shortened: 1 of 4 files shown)


databroker/databroker.py
```diff
@@ -488,6 +491,46 @@ def stream(headers, fields=None, fill=True):
         yield 'stop', header['stop']
 
 
+stream = restream  # compat
+
+
+def process(headers, func, fields=None, fill=True):
```
tacaswell: The kwargs are not used in the function body.

danielballan: I meant to pass them through to `restream`. Fixed.

PR #32, opened 2016-02-12
ENH: Search for fields in event, descriptor, start, stop.

danielballan:
- Any fields not found in an event will be looked for in
  its descriptor's configuration, then in start, then in stop.
- This passes the existing tests, which use documents that have
  no 'configuration' field at all. More tests coming.
- A dependency on cytoolz or toolz (same API) is introduced.

### DIFF
--- a/databroker/databroker.py
+++ b/databroker/databroker.py
@@ -13,6 +13,12 @@ import filestore.api as fs
 import logging
 import numbers
 
+# Toolz and CyToolz have identical APIs -- same test suite, docstrings.
+try:
+    from cytoolz.dicttoolz import merge
+except ImportError:
+    from toolz.dicttoolz import merge
+
 
 try:
     from functools import singledispatch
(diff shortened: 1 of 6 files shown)


danielballan:
Tests pass. This fixes the bugs uncovered in #34 and also with the enhancements covered above.

databroker/pims_readers.py
```diff
@@ -44,7 +44,8 @@ def __init__(self, headers, name):
                 # do something
         """
         events = get_events(headers, [name], fill=False)
-        self._datum_uids = [event.data[name] for event in events]
+        self._datum_uids = [event.data[name] for event in events
+                            if name in event.data]
         self._len = len(self._datum_uids)
```
ericdill: if `_datum_uids` is an empty list here would it make sense to raise and let the user know that they gave a bad key?

databroker/databroker.py
```diff
@@ -258,7 +265,7 @@ def fill_event(event):
     """
     is_external = _inspect_descriptor(event.descriptor)
     for data_key, value in six.iteritems(event.data):
-        if is_external[data_key]:
+        if is_external.get(data_key, False):
```
ericdill: You could drop the `False` since `get()` returns `None` by default

databroker/databroker.py
```diff
@@ -336,29 +343,43 @@ def get_events(headers, fields=None, fill=True):
...
+                    # Look in the descriptor, then start, then stop.
+                    if field in config_data:
+                        event_data[field] = config_data[field]
+                        event_timestamps[field] = config_ts[field]
+                    elif field in start:
+                        event_data[field] = start[field]
+                        event_timestamps[field] = start['time']
+                    elif field in stop:
+                        event_data[field] = stop[field]
+                        event_timestamps[field] = stop['time']
```
ericdill: This is going to raise if the code path hits line 353

ericdill: Wait no its not, since stop will be empty and field won't be in stop. Deleting this and the previous comment

databroker/pims_readers.py
ericdill: Ah, looks like you are doing this inside of databroker._check_fields_exist.

ericdill:
ugh i hate it when you look at a word for too long and it starts to look weird.  that has happened to me with the word 'fields' in this PR.

danielballan:
https://www.youtube.com/watch?v=0vv0ru44NQA

databroker/databroker.py
```diff
@@ -505,3 +542,19 @@ def get_fields(header):
         for field in desc['data_keys'].keys():
             fields.add(field)
     return fields
+
+def _check_fields_exist(fields, headers):
+    all_fields = set()
+    for header in headers:
+        all_fields.update(header['start'])
```
ericdill: today I learned that you can update a set with a dictionary and it just grabs the keys out of the dict. neat!

ericdill:
lmao, yes, exactly @danielballan youtube link

ericdill:
Alright lgtm.  Nice work on this.

PR #33, opened 2016-02-12
typo in get_events()

chiahaoliu:
looks like there is a typo. Should `run_star` be `run_start' ?

Bug_report is #78

### DIFF
--- a/databroker/databroker.py
+++ b/databroker/databroker.py
@@ -344,7 +344,7 @@ def get_events(headers, fields=None, fill=True):
             else:
                 raise ValueError(('{!r} field is not in any descriptor '
                                   'of run start {rs.uid}').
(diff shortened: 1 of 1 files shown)


danielballan:
Thanks, @chiahaoliu. I caught this myself just now in the process of writing #32, which incidentally replaces this entire block of code.

ericdill:
@chiahaoliu It should be 'start'.  @danielballan The typo that is being fixed here is *right* before the block of code that you just wrote

danielballan:
I have unpushed code.

ericdill:
Ah, well nevermind then

danielballan:
Thanks, anyway @chiahaoliu. Sorry I stole your thunder on this one. :- )

PR #34, opened 2016-02-12
TST: Actually process the generators.

ericdill:
Actually prcessing the generator uncovered a few bugs! Now I need to 
squash them.

This PR will close #78 when done.

### DIFF
--- a/databroker/tests/test_broker.py
+++ b/databroker/tests/test_broker.py
@@ -78,6 +78,7 @@ def test_basic_usage():
     assert_true(table.notnull().all().all())
 
 
+@raises(ValueError)
 def test_get_events_bad_key():
     hdr = db[-1]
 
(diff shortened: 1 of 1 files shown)


databroker/tests/test_broker.py
```diff
@@ -63,20 +63,27 @@ def test_basic_usage():
...
 
     # test time shift issue GH9
     table = get_table(db[105])
     assert_true(table.notnull().all().all())
 
+
+def test_get_events_bad_key():
+    hdr = db[-1]
+
+    list(get_events(hdr, fields=['abcd123']))
```
danielballan: This should assert that it raises, right?

ericdill: yes, good catch.  I was having trouble making it raise and wanted to make sure it was actually failing!

ericdill: I just wrapped the function in an raises decorator

danielballan:
Master is currently broken. Merging this (which adds tests that expose the problems) merely makes that official. Merging.

danielballan:
Fixes for both of these are already in #32

PR #37, opened 2016-02-25
ENH: Allow hanlder override by data key.

danielballan:
- All get* functions also accept an optional `handler_overrides` argument
  that maps data_keys to handlers, overriding the lookup based on spec name
- The override only affects a single call.
- get_table and get_images now accept optional `handler_registry` argument
  (get_events already supported this)

Use Case:
- Debugging by say using a handler that returns raw resource/datum dicts
- Swapping lazy/non-lazy handlers as needed
- Suppose we collect data and then realize we want information that is not
  in the Events but *is* in the files (e.g., timestamps in HDF5 image files).
  This gives us a recourse: registing a custom handler that uses the same
  resource and datum documents but returns totally different data (the
  timestamps instead of the images). Obviously it's better to capture everything
  in advance, but this is liable to come up a lot.

Examples:
get_events(hdr, handler_overrides={'image': SpecialHandler})
get_table(hdr, handler_overrides{'light': Special1, 'dark': Special2})
get_images(hdr, handler_override=SpecialHandler)

### DIFF
--- a/databroker/databroker.py
+++ b/databroker/databroker.py
@@ -1,7 +1,7 @@
 from __future__ import print_function
 import warnings
 import six  # noqa
-from collections import deque
+from collections import deque, defaultdict
 from itertools import chain
 import pandas as pd
 import tzlocal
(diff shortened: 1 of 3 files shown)


klauer:
Looks handy.

tacaswell:
This is being held up by some work I need to do to mds to make the context manager behave better.

danielballan:
Test coverage is solid now.

tacaswell:
For reference, this was waiting on https://github.com/NSLS-II/filestore/pull/84 & https://github.com/NSLS-II/filestore/pull/86 which have been resolved.

PR #41, opened 2016-03-08
WIP: Add configurable Broker object; support old API.

danielballan:


danielballan:
This just follows the pattern established by @tacaswell in the recent refactoring of filestore and metadatastore. Nothing clever. If I do this right we won't have touch any existing tests (though we should add a few). Still WIP.

ericdill:
It looks like this is moving functions and `Header` from databroker.py to core.py. :+1: on that for sure.

Why move and rename the rest from databroker/DataBroker to broker/Broker?

danielballan:
Back-compatibility. DataBroker is an instance, Broker is a class.

On Tue, Mar 8, 2016 at 4:19 PM -0800, "Eric Dill" <notifications@github.com<mailto:notifications@github.com>> wrote:


It looks like this is moving functions and Header from databroker.py to core.py. [:+1:]  on that for sure.

Why move and rename the rest from databroker/DataBroker to broker/Broker?

ericdill:
I understand that you left `databroker.databroker.py` in as a shim for back compatibility.  I am trying to understand what motivated moving `databroker.databroker._DataBrokerClass` to `databroker.broker.Broker`

tacaswell:
Because we want to expose `Broker` as a (possibly) user facing class.

ericdill:
> Because we want to expose Broker as a (possibly) user facing class.

Ok, thanks.

cowanml:
any plans for a brokerbroker? or something to simplify user interaction with all the brokers?
re: https://github.com/NSLS-II/wishlist/issues/115#issuecomment-199424647

danielballan:
None yet.

databroker/broker.py
```diff
@@ -0,0 +1,448 @@
...
+# so we need to just grab the only element out of this
+@search.register(six.string_types,)
+def _(key, mds):
+    logger.info('Interpreting key = %s as a str' % key)
+    results = None
+    if len(key) == 36:
+        # Interpret key as a complete uid.
+        # (Try this first, for performance.)
+        logger.debug('Treating %s as a full uuid' % key)
+        results = list(mds.find_run_starts(uid=key))
```
tacaswell: This could just be `run_start_given_uid`

databroker/broker.py
```diff
@@ -0,0 +1,448 @@
...
+            Header objects
+
+        Examples
+        --------
+        >>> DataBroker(start_time='2015-03-05', stop_time='2015-03-10')
+        >>> DataBroker(data_key='motor1')
+        >>> DataBroker(data_key='motor1', start_time='2015-03-05')
+        """
+        data_key = kwargs.pop('data_key', None)
+        run_start = self.mds.find_run_starts(**kwargs)
```
tacaswell: This result might get over-written in the next block?

tacaswell: nm

databroker/broker.py
```diff
@@ -0,0 +1,448 @@
...
+    from cytoolz.dicttoolz import merge
+except ImportError:
+    from toolz.dicttoolz import merge
+
+
+try:
+    from functools import singledispatch
+except ImportError:
+    try:
+        # We are running on Python 2.6, 2.7, or 3.3
```
klauer: :bicyclist: :house_with_garden: Why bother with supporting these versions when all of our other stuff is >= 3.4?

databroker/broker.py
```diff
@@ -0,0 +1,448 @@
...
+
+        Parameters
+        ----------
+        event : document
+        handler_registry : dict, optional
+            mapping spec names (strings) to handlers (callable classes)
+        handler_overrides : dict, optional
+            mapping data keys (strings) to handlers (callable classes)
+        """
+        if handler_overrides is None:
```
tacaswell: This can call out to `core.fill_event`.

databroker/broker.py
```diff
@@ -0,0 +1,448 @@
...
+@search.register(str)
+@search.register(six.text_type)
+# py2: six.string_types = (basestring,)
+# py3: six.string_types = (str,)
+# so we need to just grab the only element out of this
+@search.register(six.string_types,)
+def _(key, mds):
+    logger.info('Interpreting key = %s as a str' % key)
+    results = None
+    if len(key) == 36:
```
klauer: :smiley:

databroker/core.py
```diff
@@ -0,0 +1,543 @@
...
+        -------
+        header : databroker.broker.Header
+        """
+        if isinstance(run_start, six.string_types):
+            run_start = mds.run_start_given_uid(run_start)
+        run_start_uid = run_start['uid']
+
+        try:
+            run_stop = doc.ref_doc_to_uid(mds.stop_by_start(run_start_uid),
+                                          'run_start')
```
tacaswell: I think this should be 'run_stop'

databroker/broker.py
```diff
@@ -0,0 +1,448 @@
...
+                    result.append(rs)
+            run_start = result
+        result = []
+        for rs in run_start:
+            result.append(Header.from_run_start(self.mds, rs))
+        return result
+
+    def find_headers(self, **kwargs):
+        "This function is deprecated."
+        warnings.warn("Use .__call__() instead of " ".find_headers()")
```
klauer: "a " "b" -> "a b"

databroker/broker.py
danielballan: We'll support 2.x on the *analysis* side for longer, I think. The collection side is pretty much all internal stuff, so it was easier to be aggressive there.

databroker/core.py
```diff
@@ -0,0 +1,543 @@
...
+        mapping data keys (strings) to handlers (callable classes)
+    """
+    if handler_overrides is None:
+        handler_overrides = {}
+    is_external = _inspect_descriptor(event.descriptor)
+    mock_registries = {data_key: defaultdict(lambda: handler)
+                       for data_key, handler in handler_overrides.items()}
+    for data_key, value in six.iteritems(event.data):
+        if is_external.get(data_key, False):
+            if data_key not in handler_overrides:
```
klauer: Probably should combine these two paths into one since only the registry differs

databroker/broker.py
tacaswell: because only collection is pinned to 3.4+, analysis (for now) is allowed to be in lower.

klauer: This makes things more difficult for hxntools (and such) which can include collection and analysis-related things. I will not support 2.x there. 

What's the timeline for pushing everything to 3.4+? (feel free to answer this elsewhere as this is not the place for it)

ericdill: The only blocking issue is Li's GUI is py2 only because of Enaml/Atom.  There are people working on making Enaml/Atom work on py3. I need to get to work on a conda recipe to package those stacks and see if works with Li's GUI, but I have not had the time to do so

databroker/core.py
tacaswell: There are deeper issues with this code in that we are building contexts per event, but FS lets you build contexts outside of the loop that is calling this on each event.

databroker/broker.py
danielballan: I am more worried about the issue of all the py2-only analysis projects that we do not yet know about. IMO, *some* version of databroker with 2.x support should be maintained for at least a couple years.

databroker/core.py
```diff
@@ -0,0 +1,543 @@
...
+            # some non-numpy-like type
+            return img
+
+
+def _inspect_descriptor(descriptor):
+    """
+    Return a dict with the data keys mapped to boolean answering whether
+    data is external.
+    """
+    # TODO memoize to cache these results
```
klauer: Not necessary to cache this as it's not being called a lot, plus this could be a one-liner dictionary comprehension

databroker/databroker.py
```diff
@@ -1,643 +1,22 @@
...
+get_table = DataBroker.get_table
+get_images = DataBroker.get_images
+restream = DataBroker.restream
+stream = DataBroker.stream
+process = DataBroker.process
 
 
-logger = logging.getLogger(__name__)
-TZ = str(tzlocal.get_localzone())
+# TODO Add docstrings.
```
klauer: Should just get the docstring from `DataBroker.fill_event` (programmatically even)

klauer: Scratch that, these are just methods from the singleton. This wrapper method seems unnecessary?

databroker/broker.py
klauer: How can we test 2.x with the proposed (test) dependency of bluesky on databroker/mds?

tacaswell: @klauer That is a good point :cry: 

The selling point for using bluesky/ophyd in testing mds / fs is that they are most effective method we have for generating documents :stuck_out_tongue_winking_eye:

arkilic: Bluesky is a huge dependency that changes quite often under the hood. Clearly there is value in generating these documents via bs, however I don't think it's a good idea to have co-dependent projects. One can use mds without bs but not vice versa.

databroker/core.py
danielballan: I think it should be 'run_start'. This line is dereferencing the 'run_start' field of the 'run_stop' document.

databroker/broker.py
danielballan: > Bluesky is a huge dependency that changes quite often under the hood. 

I would push back on that. The core functionality of bluesky (run_engine.py) depends on Python, jsonschema, and super_state_machine, all of which are light dependencies. And that list has been stable since the summer. The dependency creep is all due to the callbacks, and I think we agree that future callback development will be moving outside of the bluesky package. They are packed together currently only for our convenience.

databroker/core.py
danielballan: @tacaswell already wrote code to do this (using the boltons cache) which I have since cherry-picked. No harm since the work is already done!

databroker/broker.py
danielballan: I think we should leave this as is until https://github.com/NSLS-II/metadatastore/issues/205 is improved and MDS is subsequently tagged.

databroker/databroker.py
danielballan: Mysteriously, without a wrapper, importing `fill_event` fails:

```python
    from databroker.databroker import fill_event
E   ImportError: cannot import name 'fill_event'
```

But, once `fill_event` is patched up, all the other methods masquerading as top-level functions import just fine. Now that the tests are passing I'll revisit this.

databroker/broker.py
arkilic: I was referring to things like callback development moving outside bluesky package when I suggested the changes under the hood. I have no future remarks as w/e I say is not going to change the outcome

danielballan: I hope my remark above didn't come off as dismissive. I think you have a valid point. Adopting bluesky as a dev dependency of mds may indeed be a bad idea that we would come to regret. Healthy debate is needed.

databroker/databroker.py
danielballan: Ah, it was a typo all along (flll -> fill). I jumped to the conclusion that some dirty weakref stuff was going on.

databroker/core.py
```diff
@@ -0,0 +1,558 @@
...
+       This mutates the event's ``data`` field in-place
+    """
+    if handler_overrides is None:
+        handler_overrides = {}
+    is_external = _external_keys(event.descriptor)
+    mock_registries = {data_key: defaultdict(lambda: handler)
+                       for data_key, handler in handler_overrides.items()}
+    for data_key, value in six.iteritems(event.data):
+        if is_external.get(data_key) is not None:
+            if data_key not in handler_overrides:
```
danielballan: @klauer I think you comment on this codeblock (or similar) but you comment is not showing up on github anymore. The code here is inefficient and problematic, but it's not worse than current master. I'd like to revisit in the next day or two in a separate PR.

danielballan:
This is ready for another review.

databroker/broker.py
```diff
@@ -0,0 +1,448 @@
...
+from .core import (Header, _inspect_descriptor,
+                   get_events as _get_events,
+                   get_table as _get_table,
+                   restream as _restream,
+                   process as _process, Images)
+
+
+# Toolz and CyToolz have identical APIs -- same test suite, docstrings.
+try:
+    from cytoolz.dicttoolz import merge
```
klauer: This import isn't far off from 'import left_pad'. One single function to do something pretty simple in Python for something that is being prematurely optimized - dictionary merges are for each header, each item in each descriptor. I rather doubt they are a performance bottleneck.

I vote to nuke this dependency.

danielballan: Ha.

Background: I introduced this dependency fairly recently, viewing it as the first of *many* places in our databroker/datamuxer codebase that we could be using toolz/cytoolz for simplified and faster code. If we don't find more than one place to use it, sure, but I think more will come up, especially as we start dealing more often with multiple event streams.

Watching Matthew Rocklin's talk sold me on the value of the library. It implements functional patterns common in many languages -- nothing clever here -- and, like boltons, it could have been part of the standard lib.

danielballan: P.S. If we *do* decide to nuke this (which, again, is a reasonable POV) please let's not do it in the PR.

databroker/core.py
tacaswell: well, re-referencing.

PR #43, opened 2016-03-17
API: Skip empty events.

danielballan:
At beamlines, I find myself using this pattern a lot:

```python3
events = [ev for ev in get_events(hdr, ['x']) if 'x' in ev.data]  # filter out events that don't have 'x'
```

I think `get_events` should be filtering out events that are completely empty.

I've been sitting on this awhile, because this is not a decision we should thrash on, but I've become pretty convinced that the upside outweighs in the downside in the use cases I see. Thoughts?

### DIFF
--- a/databroker/core.py
+++ b/databroker/core.py
@@ -173,6 +173,10 @@ def get_events(mds, fs, headers, fields=None, fill=True, handler_registry=None,
                         event_data[field] = stop[field]
                         event_timestamps[field] = stop['time']
                     # (else omit it from the events of this descriptor)
+                if not event_data:
+                    # Skip events that are now empty because they had no
+                    # applicable fields.
+                    continue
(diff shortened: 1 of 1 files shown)


tacaswell:
How do we end up with event streams with out data in the events?

danielballan:
Data is stripped out by get_events when the user passes a whitelist of fields. Any events from streams that have none of these fields end up completely empty of data.

On Thu, Mar 17, 2016 at 8:16 AM -0700, "Thomas A Caswell" <notifications@github.com<mailto:notifications@github.com>> wrote:


How do we end up with event streams with out data in the events?

tacaswell:
Ah :sheep: .  Thought we were special casing failed data collection.

tacaswell:
This needs a rebase

PR #48, opened 2016-03-30
FIX: remove unicode

CJ-Wright:
Remove unicode statements which seem to produce a problem with python 2.


danielballan:
Thanks for your prompt efforts here, @CJ-Wright.

PR #51, opened 2016-04-21
At long last: archiver appliance integration

danielballan:
The databroker can query the EPICS Archiver Appliance and return the results as documents that are compatible with those from metadatastore. (A simulated descriptor and event stream are written.)

This is implemented using a new "plugin" capability that could also be used to integrate other external sources without changing the databroker code. So, individual beamlines can customized their databrokers. For example, we might programmatically access the Olog through databroker as well.

Tested successfully using LIX archiver and CSX metadataservice:

```python3
import metadatastore.commands, filestore.api
import databroker

# define new ArchiverPlugin (this would be done in an IPython profile...)
plugin = databroker.broker.ArchiverPlugin('http://xf16idc-ca.cs.nsls2.local:17668/', 'US/Eastern')

# define new Broker instance, with plugin mapped on the custom keyword argument 'pvs'
db = databroker.Broker(metadatastore.commands, filestore.api, {'pvs': plugin})

# Now `db.get_events` accepts a 'pvs' argument and returns more data
h = db['6988d9bd']
db.get_events(h, pvs=['XF:16IDA-OP{Mir:WBM-Ax:P}Mtr.RBV'], fill=False)
```

### DIFF
--- a/databroker/broker.py
+++ b/databroker/broker.py
@@ -1,11 +1,16 @@
+
 from __future__ import print_function
 import warnings
 import six  # noqa
 from collections import deque, defaultdict
-
+import uuid
+import datetime
 import tzlocal
+import pytz
 import logging
 import numbers
+import requests
+import doct
 from .core import (Header, _external_keys,
                    get_events as _get_events,
                    get_table as _get_table,
@@ -144,9 +149,23 @@ def _(key, mds):
 
 
 class Broker(object):
-    def __init__(self, mds, fs):
+    def __init__(self, mds, fs, plugins=None):
+        """
+        Unified interface to data sources
+
+        Parameters
+        ----------
+        mds : metadatastore or metadataclient
+        fs : filestore
+        plugins : dict or None, optional
+            mapping keyword argument name (string) to Plugin, an object
+            that should implement ``get_events``
+        """
         self.mds = mds
         self.fs = fs
+        if plugins is None:
+            plugins = {}
+        self.plugins = plugins
 
     def __getitem__(self, key):
         """DWIM slicing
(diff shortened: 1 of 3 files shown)


klauer:
👍

danielballan:
attn @hhslepicka Did you mention that LIX was looking forward to this integration, too?

licode:
good work. I will play with this. I think other beam lines also require API like this. For example, get all the ring current or stage temperature within given time window from databroker.

arkilic:
@licode In order to do that just set the archiver appliance URL to the URL of the machine archiver:
```
plugin = databroker.broker.ArchiverPlugin(**'http://<machine_archiver_address>.cs.nsls2.local:17668/'**, 'US/Eastern')
```

arkilic:
@danielballan Are we deprecating archiverlib for good now?

danielballan:
I think so. All the important bits are in this PR. Thank you for your service! ;- )

arkilic:
R.I.P. archiverlib. You lived a short but useful life.

stuwilkins:
This is great, for a long time this has been a really important feature for us. Thanks so much. Am I to understand this only works with the archiver appliance and not the traditional archiver? If so then we should probably migrate over to the new archiver appliance. This was never completed at CSX for example. 

Does this mean that we are now standardizing on this archiver for the beamlines?

arkilic:
@stuwilkins Yes, this only works with the appliance. We started standardizing with the accelerator. We had no choice after the legacy archive 'a indices were about 5TB. Archiver appliance is neater in terms of resource management, both our machine and Slac are pleased so far based on my convos with the folks. i inquired about pushing the appliance out to the beamlines, Leon is in charge of this. You might want to get with Robert and schedule.

stuwilkins:
Thanks! That all sounds great. The shutdown now should be a great time to do this. Let me touch base with Petkus’s group.S

S

PR #52, opened 2016-04-22
API: Set fill=False by default; add 'name' filtering arg.

danielballan:
Motivation:

* new `name` arg: Bluesky will be producing more asynchronous event streams. We want `get_table` to continue to "just give me my data".
* fill: Users commonly experience slow performance unnecessarily because they don't know to set `fill=False`. Also, this makes `restream` and `process` compatible with bluesky callbacks, which expect the unfilled documents emitted by the RunEngine.

See the added docstrings for details. Tests pass but coverage on new features is low. Careful review would be appreciated. Tests will improve when someone has the time to generate test data from bluesky itself, as has been proposed elsewhere.

### DIFF
--- a/databroker/core.py
+++ b/databroker/core.py
@@ -266,7 +266,7 @@ def get_table(mds, fs, headers, fields=None, name='primary', fill=False,
             stop = {}
 
         # shim for back-compat with old data that has no 'primary' descriptor
-        if not [d for d in descriptors if d.get('name') == 'primary']:
+        if not any(d for d in descriptors if d.get('name') == 'primary'):
             name = None
 
         for descriptor in descriptors:
--- a/databroker/core.py
+++ b/databroker/core.py
@@ -93,8 +93,9 @@ class Header(doc.Document):
         return cls('header', d)
 
 
-def get_events(mds, fs, headers, fields=None, fill=True, handler_registry=None,
-               handler_overrides=None, plugins=None, **kwargs):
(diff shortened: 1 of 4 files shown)


databroker/core.py
```diff
@@ -249,9 +261,17 @@ def get_table(mds, fs, headers, fields=None, fill=True, convert_times=True,
         # cache these attribute look-ups for performance
         start = header['start']
         stop = header['stop']
+        descriptors = header['descriptors']
         if stop is None:
             stop = {}
-        for descriptor in header['descriptors']:
+
+        # shim for back-compat with old data that has no 'primary' descriptor
+        if not [d for d in descriptors if d.get('name') == 'primary']:
```
arkilic: Should we start indexing descriptors on "name" in the backend?

danielballan: Maybe in the future, but I don't see this as any kind of bottleneck yet. The number of descriptors per run is not large and will probably never be more than 10s, so we just fetch them all and filter in Python.

klauer: Minor bikesheddy temporary list-removing improvement: 
`if not any(d for d in descriptors if d.get('name') == 'primary'):`

arkilic: 🚲  🏠 indeed

klauer:
I have some minor concerns with the handling of `name=None` differing in collection (bluesky) vs retrieval (broker), but otherwise am 👍 here.

PR #53, opened 2016-04-28
ENH: Add support for full-text queries using syntax `db('keyword')`

danielballan:
Exposing the full-text search added in NSLS-II/metadatastore#215 via the broker. Demo:

```python
In [5]: len(db('linear_motor'))
Out[5]: 4
```

### DIFF
--- a/databroker/broker.py
+++ b/databroker/broker.py
@@ -173,7 +173,7 @@ class Broker(object):
         """
         return search(key, self.mds)
 
-    def __call__(self, **kwargs):
+    def __call__(self, text_search=None, **kwargs):
         """Given search criteria, find Headers describing runs.
 
         This function returns a list of dictionary-like objects encapsulating
@@ -183,6 +183,8 @@ class Broker(object):
 
         Parameters
         ----------
+        text_search : str, optional
+            search full text of RunStart documents
         start_time : time-like, optional
             Include Headers for runs started after this time. Valid
             "time-like" representations are:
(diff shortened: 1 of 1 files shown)


danielballan:
^ So, to summarize, this adds a positional argument, `text_search` to `Broker.__call__`, which until now only accepted keyword arguments.

licode:
Good. 
So If I run ```DataBroker('linear_motor')  ``` and ```DataBroker(data_key='linear_motor') ```
this first one will be much faster?

databroker/broker.py
```diff
@@ -183,6 +183,8 @@ def __call__(self, **kwargs):
 
         Parameters
         ----------
+        text_serach : str, optional
```
klauer: `text_se ar ch`

databroker/broker.py
```diff
@@ -241,10 +243,16 @@ def __call__(self, **kwargs):
...
                     result.append(rs)
             run_start = result
-        result = []
+        elif text_search is not None:
+            query = {'$text': {'$search': text_search}}
+            run_start = self.mds.find_run_starts(**query)
+        else:
+            query = kwargs
+            run_start = self.mds.find_run_starts(**query)
+        headers = []
```
klauer: (bikeshed) while you're here, this is ripe for a list comprehension

klauer:
I missed the mds PR, so I'll ask here: how does case (in)sensitivity work with these searches?

danielballan:
It's highly configurable -- you can even match special characters like umlaut ~ u -- but it is case insensitive by default I believe.

On Thu, Apr 28, 2016 at 3:05 PM -0700, "K Lauer" <notifications@github.com<mailto:notifications@github.com>> wrote:


I missed the mds PR, so I'll ask here: how does case (in)sensitivity work with these searches?

danielballan:
I am abandoning this PR. Perhaps some courageous soul will figure out how to configure Travis's mongo to allow text indexes. In the meantime, it is possible to use the text-search feature with this PR by passing a raw mongo query to the broker: `db({'$text': {$'search': keyword}})`

PR #55, opened 2016-05-02
CI: Try to enable text search

ericdill:


ericdill:
Ok that didn't really help. Apparently text search is enabled by default.

```
2016-05-02T16:59:16.532+0000 I COMMAND  [main] Text search is enabled by default and cannot be disabled.
```

tacaswell:
I think the issue is that on master we are now installing mongo 3.2 (which also solves text search not available problem), but then caused this problem on merge.

PR #57, opened 2016-05-09
FIX: bug with default value of handler_registry

tacaswell:


klauer:
Per @tacaswell, tests will be fixed later

PR #62, opened 2016-06-06
API: change arg name from name -> stream_name

tacaswell:
Change the name of the 5th positional arg in `get_table` from `name` ->
'stream_name`.  The new argname is clearer and aligns with the naming
convention used in LiveTable.

### DIFF
--- a/databroker/broker.py
+++ b/databroker/broker.py
@@ -318,8 +318,8 @@ class Broker(object):
         for event in res:
             yield event
 
-
-    def get_table(self, headers, fields=None, name='primary', fill=False,
+    def get_table(self, headers, fields=None, stream_name='primary',
+                  fill=False,
                   convert_times=True, timezone=None, handler_registry=None,
                   handler_overrides=None):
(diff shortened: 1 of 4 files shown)


tacaswell:
attn @licode You should follow this naming convention in suitcase as well.

klauer:
Breaks backward-compatibility and will cause confusion 👎

tacaswell:
as far as I know no one know how to use this yet.  The point of this change is to reduce confusion with naming conventions.

licode:
Sure, I didn't start the other PR to suitcase yet. I will do it today following the name convention.

danielballan:
Yeah, AFAIK no one knows that this kwarg exists. Change now -- avoid pain later.

klauer:
Fine, I suppose I'm in the minority then. I'll update all the places HXN uses it.

tacaswell:
We can add it back as kwarg at the end and have it warn for a bit.

PR #64, opened 2016-06-10
TST/ENH: Switch tests to pytest

ericdill:


ericdill:
@tacaswell @danielballan @klauer This should be good to go.  Give it a once-over?  Things to note:

- travis.yml is simplified by using install/init scripts from github.com/nsls-ii/nsls2-ci
- make sure I didn't screw up any of the tests... There is no test suite for modifying the test suite!

klauer:
Much, much cleaner! Since I'm not an expert on these tests to verify the functionality, I'll have to leave that to @tacaswell and @danielballan.

.travis.yml
```diff
@@ -3,69 +3,46 @@ sudo: false
...
+  - . ~/scripts/init-filestore.sh
+  - . ~/scripts/init-metadatastore.sh
 
 install:
   - export GIT_FULL_HASH=`git rev-parse HEAD`
-  - conda config --set always_yes true
-  - conda install "conda<4"
-  # pin to numpy < 1.10 because pandas 0.16 and numpy 1.10 do not play nice
-  - conda create -n testenv pip nose python=$TRAVIS_PYTHON_VERSION six "numpy<1.10" pandas h5py coverage jsonschema jinja2 cytoolz
+  - conda create -n testenv pip pytest python=$TRAVIS_PYTHON_VERSION six \
```
tacaswell: something is going wrong with this line.  I just just give up on line wrapping here.

ericdill:
Good catch on the travis stuff @tacaswell thanks.

ericdill:
Wow and +14% coverage!

tacaswell:
Did you add tests?  Was something just not being run correctly before?

I am happy for the coverage increase, but it is a bit worrying.

ericdill:
I did not add tests.  I would imagine that this is a result of the difference in the old way that we were executing tests (requesting nose to start the coverage plugin) versus the new way where we are executing tests through the coverage plugin (`coverage run run_tests.py`).  Ned Batchelder was talking about this difference on the [python testing podcast] (http://pythontesting.net/podcast/coverage-ned-batchelder/).  This is the first time that I've noticed an actual difference though.

danielballan:
Wow. I think you just achieved a new level of Python cred. I am not up on Python testing podcasts.

On Fri, Jun 10, 2016 at 5:15 PM -0400, "Eric Dill" <notifications@github.com<mailto:notifications@github.com>> wrote:


I did not add tests. I would imagine that this is a result of the difference in the old way that we were executing tests (requesting nose to start the coverage plugin) versus the new way where we are executing tests through the coverage plugin (coverage run run_tests.py). Ned Batchelder was talking about this difference on the python testing podcast<http://pythontesting.net/podcast/coverage-ned-batchelder/>. This is the first time that I've noticed an actual difference though.

PR #66, opened 2016-06-16
API: if config is not found, don't import DataBroker

chiahaoliu:
Author: @danielballan

### DIFF
--- a/databroker/__init__.py
+++ b/databroker/__init__.py
@@ -1,22 +1,26 @@
+import warnings
 import logging
 
 
 logger = logging.getLogger(__name__)
-__all__ = ['DataBroker', 'get_images', 'get_events', 'get_table', 'stream',
-           'get_fields', 'restream', 'process', 'Broker', 'ArchiverPlugin']
 
 
-# generally useful imports
-from .databroker import (DataBroker, DataBroker as db,
-                         get_events, get_table, stream, get_fields,
-                         restream, process)
+try:
+    from .databroker import (DataBroker, DataBroker as db,
+                             get_events, get_table, stream, get_fields,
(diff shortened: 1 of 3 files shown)


danielballan:
As we move toward explicitly defining MDS and FS objects in the beamline configs, do the following on `import databroker`:

If MDS or FS raise because they can't find global config, warn, and do not import the top-level funcs (get_events, get_table, DataBroker). It's *very* unusual to have contingent imports but I think alternatives (require global config forever; replace `DataBroker` with `None`...) are worse.

The whole issue is that `DataBroker`, `get_events`, and friends are importable from the top level `__init__` and that import is *very* ingrained with staff at this point. For this reason, `from databroker import db` should continue to work forever for users that have a global config in place.

danielballan:
bump @tacaswell @arkilic

databroker/__init__.py
```diff
@@ -1,22 +1,26 @@
...
+    from .pims_readers import get_images
+    from .handler_registration import register_builtin_handlers
+
+    # register all built-in filestore handlers
+    register_builtin_handlers()
+    del register_builtin_handlers
+except ImportError:
+    warnings.warn("The top-level functions (get_table, get_events, etc.)"
+                  "cannot be created because "
+                  "the necessary configuration was not found.")
```
klauer: Might want to include the exception message here as this is a big try block

danielballan:
Superceded by #66

danielballan:
Err, #75

databroker/__init__.py
danielballan: Addressed in #75. Thanks.

PR #71, opened 2016-07-06
Enh fs resource

tacaswell:


tacaswell:
I can't tell if that is a joke or I can not type someplace :disappointed_relieved:

databroker/broker.py
```diff
@@ -575,6 +545,68 @@ def get_events(self, header, pvs):
...
+            for k, v in d['data_keys'].items():
+                if 'external' in v:
+                    external_keys.add(k)
+
+        ev_gen = self.get_events(header, stream_name=None,
+                                 fields=external_keys, fill=False)
+        resources = set()
+        for ev in ev_gen:
+            for v in ev['data'].values():
+                res = self.fs.resource_given_eid(v)
```
licode: what's the different between ...given_eid and ...given_uid? eid is for individual datum?

databroker/broker.py
```diff
@@ -575,6 +545,68 @@ def get_events(self, header, pvs):
                 yield Document('Event', doc)
 
 
+class DangerBroker(Broker):
```
licode: Any reason to have a new inherited class? Can we just put this function inside Broker class?

licode: Or just too dangerous to be performed by general users.

danielballan: See the warning in the docstring to the method `move_files_for_run`, below. But, @tacaswell, maybe your snarky name should come with a one-sentence explanation in the top-level docstring.

tacaswell:
attn @danielballan

tacaswell:
although, now that we need export, moving that method should probbaly be reverted

databroker/broker.py
```diff
@@ -575,6 +575,39 @@ def get_events(self, header, pvs):
...
+
+        Parameters
+        ----------
+        header : Header
+            The header to move the files of.
+
+        new_roow : str
+            The new root to move the files to
+        '''
+        resource = self.get_resource_uids(header)
```
klauer: (Reviewing with @tacaswell and @danielballan)
Method has no coverage - wait for tests for merging.

PR #63, opened 2016-06-09
ENH: Working on pretty-table summary for a list of headers

ericdill:
Leaving this here so I don't forget about it.  Also so people can give
it a look and let me know right track or wrong track

Closes #61

PR #72, opened 2016-07-22
Enh summerrize

tacaswell:
Takes over #63

### DIFF
--- a/databroker/broker.py
+++ b/databroker/broker.py
@@ -3,11 +3,15 @@ import warnings
 import six  # noqa
 import uuid
 from datetime import datetime
+import datetime as dt
+import tzlocal
 import pytz
 import logging
 import numbers
 import requests
 from doct import Document
+import prettytable
+
 from .core import (Header,
                    get_events as _get_events,
                    get_table as _get_table,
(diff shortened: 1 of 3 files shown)


databroker/broker.py
```diff
@@ -562,3 +566,146 @@ def _munge_time(t, timezone):
...
+    'duration': lambda header: (datetime.fromtimestamp(header.stop.time) -
+                                datetime.fromtimestamp(header.start.time)),
+    # The number of events per event descriptor
+    'num_events': lambda header: [
+        len(list(mdsc.get_events_generator(descriptor, False)))
+        for descriptor in header.descriptors],
+}
+
+def _get_uid(key, document):
+    try:
```
klauer: This function is very confusing... in bad need of a docstring

databroker/broker.py
```diff
@@ -562,3 +566,146 @@ def _munge_time(t, timezone):
...
+        # see if we have a hyphen
+        key, num = key.split('-')
+        num = int(num)
+    except ValueError:
+        # Nope! Must be asking for the whole uid
+        num = None
```
klauer: Slice below will fail in this case?

danielballan:
Superceded by the Qt browser.

PR #75, opened 2016-08-02
API: if config is not found, don't import DataBroker

danielballan:
Rebased copy of #66


danielballan:
Travis is not kicking off.

PR #76, opened 2016-08-08
Db export

chiahaoliu:
In progress of ``db.export`` functionality

### DIFF
--- a/databroker/broker.py
+++ b/databroker/broker.py
@@ -590,6 +590,7 @@ class Broker(object):
                 target instance of filestore to export info
         """
         for header in headers:
+            # insert mds
             mds.insert_run_start(**header['start'])
             for descripto in header['descriptor']:
                 mds.insert_descriptor(**descriptor)
@@ -597,7 +598,11 @@ class Broker(object):
                 for event in events:
                     mds.insert_event(event, descriptor)
             mds.insert_run_stop(**header['stop'])
-        pass
+            # insert fs
+            res_uids = self.get_resource_uids(header)
+            for res_uid in res_uids:
+
+        return
 
 class ArchiverPlugin(object):
     def __init__(self, url, timezone):
--- a/databroker/broker.py
+++ b/databroker/broker.py
@@ -577,6 +577,19 @@ class Broker(object):
 
     get_fields = staticmethod(get_fields)  # for convenience
 
+    def export(headers, mds, fs):
+        """ export a list of headers
+
+            Parameters:
+            -----------
+            headers : databroker.header
+                list of headers that are going to be exported
+            mds : metadatasote
+                metadata store object #FIXME
+            fs : file store
+                file store object #FIXME
+        """
+        pass
 
 class ArchiverPlugin(object):
     def __init__(self, url, timezone):
--- a/databroker/broker.py
+++ b/databroker/broker.py
@@ -600,7 +600,16 @@ class Broker(object):
             mds.insert_run_stop(**header['stop'])
             # insert fs
             res_uids = self.get_resource_uids(header)
-            for res_uid in res_uids:
+            for uid in res_uids:
+                res = self.fs.resource_given_uid(uid)
+                fs.insert_resource(res['spec'],
+                                   res['resource_path'],
+                                   None) #FIXME: resource_kwargs = None?
+                datums = self.fs.datum_gen_given_resource(uid)
+                for datum in datums:
+                    fs.insert_datum(res, datum_id, datum_kwargs)
+                    # FIXME: just the signature now, need to figure out
+                    # waht datum_uid and datum_kwargs is
 
         return
 
--- a/databroker/broker.py
+++ b/databroker/broker.py
@@ -592,9 +592,9 @@ class Broker(object):
         for header in headers:
             # insert mds
             mds.insert_run_start(**header['start'])
+            events = self.get_events(header)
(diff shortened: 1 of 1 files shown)


databroker/broker.py
```diff
@@ -509,10 +509,18 @@ def export(headers, mds, fs):
...
             mds : metadatasote
-                metadata store object #FIXME
+                metadata store object
             fs : file store
-                file store object #FIXME
+                file store object
         """
+        for header in headers:
+            mds.insert_run_start(**header['start'])
+            for descripto in header['descriptor']:
```
arkilic: typo `descripto`

databroker/broker.py
```diff
@@ -501,6 +501,28 @@ def process(self, headers, func, fields=None, fill=False):
...
+            fs : file store
+                file store object
+        """
+        for header in headers:
+            mds.insert_run_start(**header['start'])
+            for descripto in header['descriptor']:
+                mds.insert_descriptor(**descriptor)
+                events = self.get_events(header)
+                for event in events:
+                    mds.insert_event(event, descriptor)
```
danielballan: The `insert_event` API works the same as the other insert methods: it takes `**event`. The `descriptor` arg in its signature refers to `event['descriptor']`, which is the uid for `descriptor`.

Also, this for loop should be un-indented. `get_events` returns events for *all* descriptors, in bulk . It does not need to be repeated for each descriptor.

databroker/broker.py
chiahaoliu: caught me

chiahaoliu:
update: code was modified based on instruction and typo was fixed

danielballan:
Looks good. Next up is FileStore. Steps will be:

1. Use the databroker method `get_resource_uids` to identify which resource documents we need to copy.
2. Get these resource documents from `self.fs` and insert them into `fs`.
3. Use a (yet to be written) method on filestore to identify which datum documents we need to copy, and insert them into `fs` as well.

At this point, the `mds` and `fs` will work correctly, but `fs` still refers to files (like TIFFs) that are on the local filesystem. We need to copy the underlying files to a new location and then update `fs` to point there instead.

4. Add an additional required parameter to the signature of `db.export` named `new_root`.
5. Copy the files into the `new_root` directory using a (yet to be written) method on filestore handlers.
6. Update the copied resource documents in the new `fs` using `change_root`.

databroker/broker.py
```diff
@@ -565,6 +565,26 @@ def process(self, headers, func, fields=None, fill=False):
...
+    def export(headers, mds, fs):
+        """ export a list of headers
+
+            Parameters:
+            -----------
+            headers : databroker.header
+                list of headers that are going to be exported
+            mds : metadatasote
+                metadata store object
+            fs : file store
```
danielballan: file store -> FileStore

databroker/broker.py
```diff
@@ -565,6 +565,26 @@ def process(self, headers, func, fields=None, fill=False):
...
                  fields=fields, fill=fill)
 
+    def export(headers, mds, fs):
+        """ export a list of headers
+
+            Parameters:
+            -----------
+            headers : databroker.header
+                list of headers that are going to be exported
+            mds : metadatasote
```
danielballan: metadatasote -> MDS
metadata store object -> target instance of metadatastore to export into

databroker/broker.py
```diff
@@ -565,6 +565,26 @@ def process(self, headers, func, fields=None, fill=False):
...
+                metadata store object
+            fs : file store
+                file store object
```
danielballan: file store object -> target instance of filestore to export into

danielballan:
Other people are expressing interest in this feature. Do you have time to work with me to push this across the finish line next week?

chiahaoliu:
Sure, I would love to finish it by next week.

databroker/broker.py
```diff
@@ -577,6 +577,46 @@ def process(self, headers, func, fields=None, fill=False):
...
+                target instance of filestore to export info
+            new_root : str
+                optional. root directory of files that are going to
+                be exported
+        """
+        for header in headers:
+            # insert mds
+            mds.insert_run_start(**header['start'])
+            events = self.get_events(header)
+            for descripto in header['descriptor']:
```
danielballan: typo

databroker/broker.py
```diff
@@ -577,6 +577,46 @@ def process(self, headers, func, fields=None, fill=False):
 
     get_fields = staticmethod(get_fields)  # for convenience
 
+    def export(headers, mds, fs, new_root=None):
```
danielballan: We can simplify this signature by just passing in a `Broker` instance that has `mds` and `fs` as attributes. I would suggest the signature: `export(headers, db, new_root=None`)

databroker/broker.py
```diff
@@ -577,6 +577,46 @@ def process(self, headers, func, fields=None, fill=False):
...
+        """ export a list of headers
+
+            Parameters:
+            -----------
+            headers : databroker.header
+                list of headers that are going to be exported
+            mds : MDS
```
danielballan: (update docstring accordingly)

databroker/broker.py
danielballan: Also, this should have `self` as the first arg.

danielballan:
These commits were incorporated in #88.

PR #77, opened 2016-09-01
Add alias feature

danielballan:


databroker/broker.py
```diff
@@ -230,12 +232,28 @@ def clear_filters(self, **kwargs):
...
+        if callable(query):
+            query = query()
+        return self(**query)
+
+    def alias(self, key, **query):
+        if hasattr(self, key) and key not in self.aliases:
+            raise ValueError("'%s' is not a legal alias." % key)
+        self.aliases[key] = query
+
+    def dynamic_alias(self, key, func):
```
tacaswell: clever

databroker/broker.py
```diff
@@ -230,12 +232,28 @@ def clear_filters(self, **kwargs):
...
+        if hasattr(self, key) and key not in self.aliases:
+            raise ValueError("'%s' is not a legal alias." % key)
+        self.aliases[key] = func
```
tacaswell: This seems like it will not play well with historydict

danielballan: Heh, duh. Yes. I was thinking, historydict might be overkill here anyway. People can set these up in their IPython profiles. I think the dynamic-ness is more useful than the automatic persistence. Especially since the "history" aspect of historydict is unlikely to be of much use here.

danielballan: Yeah, I think historydict was too much -- at least for a first pass. Removed.

danielballan:
Thanks, @arkilic.

arkilic:
This is nifty

hhslepicka:
👏  sweet!!!

databroker/broker.py
```diff
@@ -233,6 +234,25 @@ def __getitem__(self, key):
...
+    def __getattr__(self, key):
+        try:
+            query = self.aliases[key]
+        except KeyError:
+            raise AttributeError(key)
+        if callable(query):
+            query = query()
+        return self(**query)
+
+    def alias(self, key, **query):
```
klauer: Too late to bikeshed, darn. `alias` being ambiguously as a verb/noun and there being no docstring, I was confused to where this was going...

danielballan: Oops I'll add a docstring.

On Thu, Sep 1, 2016 at 7:27 PM -0400, "K Lauer" <notifications@github.com<mailto:notifications@github.com>> wrote:


In databroker/broker.py<https://github.com/NSLS-II/databroker/pull/77#discussion_r77271961>:

> @@ -233,6 +234,25 @@ def __getitem__(self, key):
>          """Do-What-I-Mean slicing"""
>          return search(key, self.mds)
>
> +    def __getattr__(self, key):
> +        try:
> +            query = self.aliases[key]
> +        except KeyError:
> +            raise AttributeError(key)
> +        if callable(query):
> +            query = query()
> +        return self(**query)
> +
> +    def alias(self, key, **query):


Too late to bikeshed, darn. alias being ambiguously as a verb/noun and there being no docstring, I was confused to where this was going...

tacaswell:
It is not tagged yet and definitly not deployed anywhere.

:+1: to bikeshedding the name.

PR #78, opened 2016-09-01
WIP: do not use singletons from mds/fs

tacaswell:
this breaks all the tests

attn @danielballan

### DIFF
--- a/databroker/databroker.py
+++ b/databroker/databroker.py
@@ -1,17 +1,25 @@
 "This module exists for back-compatability only."
 import warnings
 
-from .broker  import Broker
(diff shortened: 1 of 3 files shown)


arkilic:
@tacaswell this breaks EVERYTHING

klauer:
@arkilic @tacaswell Does this break everything?

danielballan:
Sounds like a great opportunity for me to rename everything.

tacaswell:
We should only use the letters `{'n', 'd','b', 'l', 'a', }` in the names of things.

danielballan:
Superceded.

PR #80, opened 2016-09-02
Refactor tests

danielballan:
With :heart: for @arkilic, this is a complete rewrite of the databroker tests, clearing the way for the commands module to be removed from MDS.

Improvements:
* The databroker tests use a pytest fixture that tests against both standard (pymongo) metadatastore/filestore and the sqlite varaiant of portable-mds/portable-fs (cc @chiahaoliu). This means that a Broker backed by a portable-* on some user's laptop is a first-class citizen, passing all the same tests as a "production" Broker running at NSLS-II.
* Remove all the messy old code for generating sample data (mostly my own work), which pre-dated bluesky. Now sample data is generated in each test using an actual RunEngine, so we know that the sample data is realistic. This adds bluesky as a test dependency but not a run dep.


danielballan:
P.S. This means we will have to write completely separate tests for Python 2, since bluesky is 3 only. That's fine, though; a couple quick tests will be enough to check for any major problems, which is all we really need (since Python 3 tests will cover the subtle bugs).

danielballan:
py3 tests pass on Travis :tada:

chiahaoliu:
"❤️ for @arkilic"  👍

arkilic:
Too much love in this PR. I wish travis showed some ❤️ as well

PR #82, opened 2016-09-06
WIP: Fix stream name

danielballan:
One more instance of "maybe Ken was right," for the road.

This is WIP, needs more tests.

### DIFF
--- a/databroker/core.py
+++ b/databroker/core.py
@@ -22,6 +22,11 @@ except ImportError:
 logger = logging.getLogger(__name__)
 
 
+class ALL:
+    "Sentinel used as the default value for stream_name"
+    pass
+
+
 def fill_event(fs, event, handler_registry=None, handler_overrides=None):
     """
     Populate events with externally stored data.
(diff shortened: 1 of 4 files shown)


databroker/core.py
```diff
@@ -22,6 +22,11 @@
 logger = logging.getLogger(__name__)
 
 
+class ALL:
```
tacaswell: Is this pattern better than `ALL = object()`?

klauer: Isn't it just so he can shove the docstring in there?

tacaswell: never mind, this is the pattern numpy follows.

danielballan:
This passes all pymongo-based tests but fails sqlite because monitors emit documents from a separate thread. That is a problem with portable-mds, not databroker. Effectively, we are requiring `insert_*` methods to be threadsafe, and portable-mds needs to be fixed to satisfy that.

tacaswell:
That is a fun bug.  The thread-non-safeness of sqlite has caused us a fair amount of pain.

PR #86, opened 2016-09-12
Revert "ENH: typos"

arkilic:
Reverts NSLS-II/databroker#84 bc travis isn't happy


danielballan:
Changes to `code-block` content don't break Travis. Travis is permanently unhappy until we figure out threading at portable-mds. I can mark the tests in question as known-fail if we don't get to that soon.

PR #87, opened 2016-09-14
Remove hard dependency on metadatastore

danielballan:
See commits msgs for details.

### DIFF
--- a/databroker/broker.py
+++ b/databroker/broker.py
@@ -1,16 +1,14 @@
 from __future__ import print_function
 import warnings
 import six  # noqa
-from collections import deque, defaultdict
 import uuid
 from datetime import datetime
-import tzlocal
 import pytz
 import logging
 import numbers
 import requests
 from doct import Document
-from .core import (Header, _external_keys,
+from .core import (Header,
                    get_events as _get_events,
                    get_table as _get_table,
                    restream as _restream,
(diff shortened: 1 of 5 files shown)


databroker/core.py
```diff
@@ -442,6 +441,7 @@ def process(mds, fs, headers, func, fields=None, fill=False):
 
 def register_builtin_handlers(fs):
     "Register all the handlers built in to filestore."
+    from filestore import handlers, HandlerBase
```
tacaswell: This is a compelling case that handlers should live in the databroker

databroker/broker.py
```diff
@@ -19,7 +19,103 @@
                    get_fields,  # for conveniece
                    ALL
                   )
-from metadatastore.core import _format_time
+
+
+def _format_time(search_dict, tz):
```
tacaswell: Is this functionality going to move up to databroker as well?

databroker/core.py
arkilic: 👍

databroker/broker.py
danielballan: It already has.

databroker/core.py
danielballan: Yeah, I think that's right. Let's keep that in mind but wait until we are sure how the packaging in going to settle out overall.

danielballan:
The only test still failing is the portable-mds threading-related one. Ready to merge.

tacaswell:
Can you put an `pytset.xfail` on that test?

danielballan:
Done. Looks like we have :+1: all around, so I'm merging this to fix Travis builds for other PRs.

PR #88, opened 2016-09-15
Db export, again

danielballan:
This is a rebase of #76 with some extra commits to add a test and make it pass.


danielballan:
This may be ready. It could use a careful read from more than one reviewer. As you can see, the outline for copying external files is here, but not finished. (The docs include warning about that.) I'd like to roll this out as is and handle the external files piece in a second pass.

databroker/broker.py
```diff
@@ -708,6 +708,62 @@ def process(self, headers, func, fields=None, fill=False):
...
+            db.mds.insert_run_stop(**_clean_doct(header['stop']))
+            # insert fs
+            res_uids = self.get_resource_uids(header)
+            for uid in res_uids:
+                res = self.fs.resource_given_uid(uid)
+                db.fs.insert_resource(res['spec'],
+                                      res['resource_path'],
+                                      res['resource_kwargs'],
+                                      root=new_root)
+                                      # FIXME: revisit root when dealing
```
licode: The new_root is the root place where the external data file is going to transferred? So basically we can transfer data between different servers.

danielballan:
@licode Yes, that is the plan. As written so far, this only copies database documents between servers. A future iteration will make use of the new_root argument to copy files as well.

licode:
Got it. thanks. 
Can we detect the root path automatically? So we do not need to manually change root, or keep this as an optional argument.

danielballan:
How would that work?

On Thu, Sep 15, 2016 at 3:52 PM -0400, "Li Li" <notifications@github.com<mailto:notifications@github.com>> wrote:


Got it. thanks.
Can we detect the root path automatically? So we do not need to manually change root, or keep this as an optional argument.

licode:
I think I oversimplify the issue. Please ignore that.

licode:
👍  This looks good to me

databroker/broker.py
```diff
@@ -708,6 +708,62 @@ def process(self, headers, func, fields=None, fill=False):
...
+                    db.mds.insert_event(**event)
+            db.mds.insert_run_stop(**_clean_doct(header['stop']))
+            # insert fs
+            res_uids = self.get_resource_uids(header)
+            for uid in res_uids:
+                res = self.fs.resource_given_uid(uid)
+                db.fs.insert_resource(res['spec'],
+                                      res['resource_path'],
+                                      res['resource_kwargs'],
+                                      root=new_root)
```
tacaswell: This should pass through `res['root']` until this is actually implemented.

databroker/broker.py
```diff
@@ -708,6 +708,62 @@ def process(self, headers, func, fields=None, fill=False):
...
+        """
+        try:
+            headers.items()
+        except AttributeError:
+            pass
+        else:
+            headers = [headers]
+        for header in headers:
+            # insert mds
+            db.mds.insert_run_start(**_clean_doct(header['start']))
```
tacaswell: Why do you need this?

tacaswell: and by 'this' I mean `_clean_doct`

databroker/broker.py
```diff
@@ -799,3 +855,10 @@ def _munge_time(t, timezone):
     """
     t = datetime.fromtimestamp(t)
     return timezone.localize(t).replace(microsecond=0).isoformat()
+
+
+def _clean_doct(d):
```
tacaswell: There is a function in doct for doing this splitting.

databroker/broker.py
danielballan: Because a doct has a _name attribute hanging off of it that gets unpacked in insert_run_start.

danielballan: I am pretty sure we should get rid of '_name' -- an artifact of mongoengine documents -- but that's a question for another day.

danielballan:
This is close, but it requires a `FileStore` that implements resource_given_eid. @tacaswell only implemented this method on `FileStoreMoving`, which means that users would need to set up `FileStoreMoving` on the production databroker before exporting. It seem this is actually encouraging the opposite of careful behavior, but I think @tacaswell has some opinions about these methods, so I'll leave the rest to him.

danielballan:
If https://github.com/NSLS-II/filestore/pull/121 is merged, this PR should just work. But I don't want to run roughshod over any concerns @tacaswell may have here, so please do not merge unless you are @tacaswell. :- )

databroker/core.py
```diff
@@ -190,7 +191,7 @@ def get_events(mds, fs, headers, fields=None, stream_name=ALL, fill=False,
                 # Look in the descriptor, then start, then stop.
                 config_data_fields = set(filter(comp_re.match, config_data)) - selected_fields
                 for field in config_data_fields:
-                    selected_fields.append(field)
+                    selected_fields.add(field)
```
danielballan: This part is an unrelated bug fix.

danielballan:
Note to whoever has the patience to finish this someday: `FileStore.datum_gen_given_resource` calls `core.get_datumkw_by_resuid_gen` which returns the kwargs from the datums but successfully obscures the datum_ids. FileStore is designed to make it hard to get this stuff, and that aspect of the design is proving to be a wild success.

databroker/broker.py
danielballan: Note to anyone following along: this is now actually implemented.

databroker/broker.py
```diff
@@ -722,6 +724,73 @@ def process(self, headers, func, fields=None, fill=False):
...
+            res_uids = self.get_resource_uids(header)
+            for uid in res_uids:
+                fps = self.fs.copy_files(uid, new_root=new_root, **copy_kwargs)
+                file_pairs.extend(fps)
+                res = self.fs.resource_given_uid(uid)
+                db.fs.insert_resource(res['spec'],
+                                      res['resource_path'],
+                                      res['resource_kwargs'],
+                                      root=new_root)
+                datums = self.fs.datum_gen_given_resource(uid)
```
danielballan: https://github.com/NSLS-II/filestore/pull/122 will make this method,`datum_gen_given_resource`, return the full datum document (including the ID which we need) instead of just the datum kwargs.

danielballan:
Test passes locally and should pass here once https://github.com/NSLS-II/filestore/pull/122 and https://github.com/NSLS-II/bluesky/pull/612 are in. Phew.

tacaswell:
Not wanting to feel left out I added a commit to this PR too!

danielballan:
Test failure is unrelated to this PR but probably real. I'll open a separate issue and expand on that.

Seeing @tacaswell's marked approval above, and knowing that three of us have code in here, I guess I'll be the one to pull the trigger....

databroker/broker.py
```diff
@@ -722,6 +724,75 @@ def process(self, headers, func, fields=None, fill=False):
 
     get_fields = staticmethod(get_fields)  # for convenience
 
+    def export(self, headers, db, new_root=None, copy_kwargs=None):
+        """
+        Export a list of headers.
+
+        .. warning::
```
CJ-Wright: Does this warning still apply?

tacaswell: It does not

PR #89, opened 2016-09-28
FIX: use default stream name in get_resource_uids

tacaswell:
This was missed when the stream_name API was change.


tacaswell:
@danielballan @licode this probably should be merged, re-tagged and built before we start deploying.

danielballan:
I just tagged v0.6.2. @licode would you please update the recipe?

PR #90, opened 2016-09-28
MNT: add a cli tool for renaming files

tacaswell:
Given a runstart, rename all of the files belonging to all of the
resources.

If the root is not set on the resource you _must_ use preshift to move
some of the directory level into 'root' to be replaced.

Use of this tool is strongly discouraged.

attn @heroux


danielballan:
Might be useful to add an entry point to the setup.py.

PR #91, opened 2016-10-28
WIP: get_events fields now accepts regular expressions as filters

hhslepicka:


databroker/core.py
```diff
@@ -175,23 +180,29 @@ def get_events(mds, fs, headers, fields=None, stream_name=ALL, fill=False,
...
                 event_data = event.data  # cache for perf
                 event_timestamps = event.timestamps
                 for field in discard_fields:
                     del event_data[field]
                     del event_timestamps[field]
-                for field in extra_fields:
+                if not no_fields_filter:
                     # Look in the descriptor, then start, then stop.
-                    if field in config_data:
+                    config_data = set(filter(comp_re.fullmatch, config_data)) - selected_fields
```
tacaswell: Can't this be done outside of the loop over events? (above line 185)?

It looks like we should be able do to a 

```python
all_the_extra_data = {...}
all_the_extra_ts = {...}
for ev in mds.get_events_generator(...):
    ev['data'].update(all_the_extra_data)
    ev['timestamps'].update(all_the_extra_ts)
    yield ev
```

tacaswell:
I have expressed some concern about this to @danielballan in person, but I am starting to worry a lot about mutating the events as they go by without also updating the descriptors.  We are currently not doing much with the descriptors (@licode's use in suitcase is all I can think of off the top of my head), but suspect it will become important going forward.

databroker/core.py
hhslepicka: Yes, it can be done outside! =) I will fix it.

hhslepicka:
@tacaswell @danielballan Would you guys want to get together and talk about the concerns in person to make it easier?

hhslepicka:
@tacaswell @danielballan Now all the tests are passing and this is good for another round of review.

licode:
looks good to me.

licode:
👍

hhslepicka:
Thanks @licode!

PR #96, opened 2016-11-02
Fix timezones

tacaswell:
```python
import databroker
import bluesky
import bluesky.plans as bp
import datetime
from bluesky.examples import det

# or a port
# import metadatastore.mds
# mds = metadatastore.mds.MDS({'host': 'localhost', 'database': 'time_test', 'port': 27017,
#                              'timezone': 'US/Eastern'}, auth=False)

import portable_mds.sqlite.mds as pmsm

mds = pmsm.MDS({'directory': '/tmp/pmds', 'timezone': 'US/Eastern'})
db = databroker.Broker(mds, None)



class TimeFaker:
    def __init__(self, mds, start_time, step):
        self.start_time = start_time.timestamp()
        self.step = step
        self.counter = 0
        self.mds = mds
        self.last_uid = None
        
    def insert(self, name, doc):
        if name == 'start':
            self.counter = 0
            self.last_uid = doc['uid']
            
        if 'time' in doc:
            doc['time'] = self.start_time + self.counter * self.step
        self.counter += 1
        self.mds.insert(name, doc)
    

start_time = datetime.datetime(2016, 3, 12, 23, 5)
# start_time = datetime.datetime(2016, 11, 5, 23, 5)

tf = TimeFaker(db.mds, start_time, 5*60)

RE = bluesky.RunEngine()
RE.subscribe('all', tf.insert)

RE(bp.count([det], 50))

print(db.get_table(db[tf.last_uid]))
```

which gives the following tables
<details>
```
In [20]:                   time  det
1  2016-11-05 23:15:00  1.0
2  2016-11-05 23:20:00  1.0
3  2016-11-05 23:25:00  1.0
4  2016-11-05 23:30:00  1.0
5  2016-11-05 23:35:00  1.0
6  2016-11-05 23:40:00  1.0
7  2016-11-05 23:45:00  1.0
8  2016-11-05 23:50:00  1.0
9  2016-11-05 23:55:00  1.0
10 2016-11-06 00:00:00  1.0
11 2016-11-06 00:05:00  1.0
12 2016-11-06 00:10:00  1.0
13 2016-11-06 00:15:00  1.0
14 2016-11-06 00:20:00  1.0
15 2016-11-06 00:25:00  1.0
16 2016-11-06 00:30:00  1.0
17 2016-11-06 00:35:00  1.0
18 2016-11-06 00:40:00  1.0
19 2016-11-06 00:45:00  1.0
20 2016-11-06 00:50:00  1.0
[...]

PR #98, opened 2016-11-02
ENH: add flag to get_table to control tz-awareness

tacaswell:
Get tz-aware datetime objects back

This still needs tests + doc + consider if this is really the best way to do this.


tacaswell:
See #96 for testing code to play with this.

tacaswell:
This rebase is terrible due to renaming `core.py` to `_core.py`, as it still needs tests, punting.

PR #100, opened 2016-11-10
MNT: Add flexibility that would allow a separate EventStore.

danielballan:
As we experiment with different MDS backends for portability and performance, it might make sense to be able to choose the backend for headers and the backend for events separately. For example, it seems likely that all beamlines will use the standard MongoDB backend for headers forever, but HXN might experiment with a more performant backend for events.

This adds the possibility of having separate MDS-like objects for the header documents and the event documents. If a separate "EventStore" object is specified, that object is used for all calls to `get_events_generator` and `get_events_table`. If not, the MDS object is used as now.

There are not implementations of `EventStore` per se, and I don't plan to start making more alternative backends any time soon. But whenever someone gets around to it it would be good to have this flexibility in Broker ready and widely deployed.

### DIFF
--- a/databroker/broker.py
+++ b/databroker/broker.py
@@ -246,7 +246,7 @@ def _(key, mds):
 
 
 class Broker(object):
-    def __init__(self, mds, fs, plugins=None, filters=None):
+    def __init__(self, mds, fs, plugins=None, filters=None, es=None):
         """
         Unified interface to data sources
 
@@ -260,9 +260,12 @@ class Broker(object):
         filters : list
             list of mongo queries to be combined with query using '$and',
             acting as a filter to restrict the results
+        es : EventStoreRO, optional
+            If not provides, mds is expected to provide event storage.
         """
         self.mds = mds
         self.fs = fs
+        self.es = es
         if plugins is None:
             plugins = {}
         self.plugins = plugins
@@ -270,6 +273,17 @@ class Broker(object):
             filters = []
         self.filters = filters
         self.aliases = {}
+
+    @property
+    def es(self):
+        return self._es
+
+    @es.setter
+    def es(self, es):
+        if es is None:
+            self._es = self.mds
+        else:
+            self._es = es
 
     ALL = ALL  # sentinel used as default value for `stream_name`
 
@@ -514,7 +528,7 @@ class Broker(object):
         ------
         ValueError if any key in `fields` is not in at least one descriptor pre header.
         """
-        res = _get_events(mds=self.mds, fs=self.fs, headers=headers,
+        res = _get_events(mds=self.mds, fs=self.fs, es=self.es, headers=headers,
                           fields=fields, stream_name=stream_name, fill=fill,
                           handler_registry=handler_registry,
                           handler_overrides=handler_overrides,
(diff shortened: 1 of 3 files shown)


klauer:
Interesting! 👍

danielballan:
Technologies we have tried for MDS:

* pymongo
* mongoquery
* sqlite
* hdf5 (but only partly)

Technologies we could try for EventStore:

* asyncpg
* hdf5

Technologies we could try for EventStore as an emphemeral caching layer (with some more robust solution providing long-term storage underneath):

* feather
* parquet
* numpy binary

databroker/broker.py
```diff
@@ -260,9 +260,12 @@ def __init__(self, mds, fs, plugins=None, filters=None):
         filters : list
             list of mongo queries to be combined with query using '$and',
             acting as a filter to restrict the results
+        es : EventStoreRO, optional
+            If not provides, mds is expected to provide event storage.
         """
         self.mds = mds
         self.fs = fs
+        self.es = es
```
tacaswell: I suggest doing this by assigning es as mds if es is None and making it required in `core`

danielballan:
That would be easier, but currently mds and fs are mutable attributes, didn't want to break that.

On Thu, Nov 10, 2016 at 9:05 PM -0500, "Thomas A Caswell" <notifications@github.com<mailto:notifications@github.com>> wrote:


@tacaswell commented on this pull request.

________________________________

In databroker/broker.py<https://github.com/NSLS-II/databroker/pull/100#pullrequestreview-8136461>:

>          """
         self.mds = mds
         self.fs = fs
+        self.es = es


I suggest doing this by assigning es as mds if es is None and making it required in core

danielballan:
Ah, nevermind. Disregard last comment. Yes, your way is better.

On Thu, Nov 10, 2016 at 9:15 PM -0500, "Allan, Daniel" <dallan@bnl.gov<mailto:dallan@bnl.gov>> wrote:

That would be easier, but currently mds and fs are mutable attributes, didn't want to break that.

On Thu, Nov 10, 2016 at 9:05 PM -0500, "Thomas A Caswell" <notifications@github.com<mailto:notifications@github.com>> wrote:


@tacaswell commented on this pull request.

________________________________

In databroker/broker.py<https://github.com/NSLS-II/databroker/pull/100#pullrequestreview-8136461>:

>          """
         self.mds = mds
         self.fs = fs
+        self.es = es


I suggest doing this by assigning es as mds if es is None and making it required in core

danielballan:
Refactored as suggested above.

tacaswell:
This needs API changes docs for the core methods.

tacaswell:
The other composition point might make sense here is to require the `MDS` instance to have a `ES` instance and make _that_ set of things more pluggable.

I _think_ doing it at the databroker layer makes more sense and is more flexible.

danielballan:
> I *think* doing it at the databroker layer makes more sense and is more flexible.

Agree that the implications of this are not *completely* clear but that broker seems like the best composition point. If nothing else it is already *a* composition point so we are keeping things simpler by avoided nested composition.

tacaswell:
The thing that feel weird as that we now have 3 very sightly coupled 'plugins' (mds, fs, es) and that a system for adding other plugins which are much less coupled (the archiver plugin).  Those do not seem like they should be on the same footing.

A way of slicing this that might make some sense in that the start/stop collections are special and then the databroker has 'plugins' that provide `descs = obj.get_descs(start, stop=None)` and `data = obj.get_data(desc)`.  That puts the events / fs / descriptors on the same footing as the archive code/abstraction wise which might make this easier to think about.

tacaswell:
After letting it sit for two weeks, I still think that notionally bundling the start/stop and the descriptor/event into separate blended implementations makes sense.

The `get_descriptor` calls should be run against the `es` instance.

This also would get mds back closer to the initial design ;)

danielballan:
Superceded by #107 :tada:

Issue #104, opened 2016-11-21
ENH: track databroker or (MDS/FS) upon header return


CJ-Wright:
It might be extremely useful to include some sort of database identifying uid(s) (or connection information?) in the returned headers, that way we could access the database that the header came from only knowing the header. An example of this is getting all the events from an otherwise unknown databroker given the header.

Moved from NSLS-II/metadatastore#260

danielballan:
As currently factored there is no nice way to customize what is in the header. The code could and probably should be refactored so that subclasses can do this.

CJ-Wright:
Actually I have found a temp fix. The current plan is to load all the raw data into the analysis databroker. We'll keep the exp data in a backup db just in case we blow something up. This way we only have one databroker to talk to at a time. Long term it would be good to have though

CJ-Wright:
I think this has been fixed by #107

PR #107, opened 2016-11-30
make Headers aware of the DataBroker from whence they came

tacaswell:
There is more to come on this PR

### DIFF
--- a/databroker/broker.py
+++ b/databroker/broker.py
@@ -246,7 +246,7 @@ def _(key, db):
 
 
 class Broker(object):
-    def __init__(self, mds, fs, plugins=None, filters=None):
+    def __init__(self, mds, fs, plugins=None, filters=None, es=None):
         """
         Unified interface to data sources
 
@@ -260,9 +260,12 @@ class Broker(object):
         filters : list
             list of mongo queries to be combined with query using '$and',
             acting as a filter to restrict the results
+        es : EventStoreRO, optional
+            If not provides, mds is expected to provide event storage.
         """
         self.mds = mds
         self.fs = fs
+        self.es = es
         if plugins is None:
             plugins = {}
         self.plugins = plugins
@@ -270,6 +273,17 @@ class Broker(object):
             filters = []
         self.filters = filters
         self.aliases = {}
+
+    @property
+    def es(self):
+        return self._es
+
+    @es.setter
+    def es(self, es):
+        if es is None:
+            self._es = self.mds
+        else:
+            self._es = es
 
     ALL = ALL  # sentinel used as default value for `stream_name`
 
@@ -578,7 +592,7 @@ class Broker(object):
         """
         if timezone is None:
             timezone = self.mds.config['timezone']
-        res = _get_table(mds=self.mds, fs=self.fs, headers=headers,
+        res = _get_table(mds=self.mds, fs=self.fs, es=self.es, headers=headers,
                          fields=fields, stream_name=stream_name, fill=fill,
                          convert_times=convert_times,
                          timezone=timezone, handler_registry=handler_registry,
@@ -610,8 +624,9 @@ class Broker(object):
         >>> for image in images:
                 # do something
         """
-        return Images(self.mds, self.fs, headers, name, handler_registry,
-                      handler_override)
+        return Images(mds=self.mds, fs=self.fs, es=self.es, headers=headers,
+                      name=name, handler_registry=handler_registry,
+                      handler_override=handler_override)
 
     def get_resource_uids(self, header):
         '''Given a Header, give back a list of resource uids
@@ -681,7 +696,8 @@ class Broker(object):
         --------
         process
         """
-        res = _restream(self.mds, self.fs, headers, fields=fields, fill=fill)
+        res = _restream(mds=self.mds, fs=self.fs, es=self.es, headers=headers,
+                        fields=fields, fill=fill)
         for name_doc_pair in res:
             yield name_doc_pair
 
@@ -721,8 +737,8 @@ class Broker(object):
         --------
         restream
         """
-        _process(mds=self.mds, fs=self.fs, headers=headers, func=func,
-                 fields=fields, fill=fill)
+        _process(mds=self.mds, fs=self.fs, es=self.es, headers=headers,
+                 func=func, fields=fields, fill=fill)
 
     get_fields = staticmethod(get_fields)  # for convenience
 
--- a/databroker/broker.py
+++ b/databroker/broker.py
@@ -486,23 +486,24 @@ class BrokerES(object):
         warnings.warn("Use .get_events() instead.")
         return self.get_events(headers, fill=fill)
 
-    def fill_event(self, event, handler_registry=None, handler_overrides=None):
+    def fill_event(self, event, inplace=False,
+                   handler_registry=None, handler_overrides=None):
         """
         Populate events with externally stored data.
-
-        .. warning
-
-            This mutates the event's ``data`` field in-place
 
         Parameters
         ----------
         event : document
+        inplace : bool, optional
+            If the event should be filled 'in-place' by mutating the data
+            dictionary.  Defaults to `False`.
         handler_registry : dict, optional
             mapping spec names (strings) to handlers (callable classes)
         handler_overrides : dict, optional
             mapping data keys (strings) to handlers (callable classes)
         """
-        return self.es.fill_event(event, handler_registry=handler_registry,
+        return self.es.fill_event(event, inplace=inplace,
+                                  handler_registry=handler_registry,
                                   handler_overrides=handler_overrides)
 
     def get_events(self, headers, fields=None, stream_name=ALL, fill=False,
--- a/databroker/broker.py
+++ b/databroker/broker.py
@@ -518,13 +518,11 @@ class Broker(object):
         ------
         ValueError if any key in `fields` is not in at least one descriptor pre header.
         """
-        res = _get_events(mds=self.mds, fs=self.fs, headers=headers,
-                          fields=fields, stream_name=stream_name, fill=fill,
-                          handler_registry=handler_registry,
-                          handler_overrides=handler_overrides,
-                          plugins=self.plugins, **kwargs)
-        for event in res:
-            yield event
+        return _get_events(headers=headers,
+                           fields=fields, stream_name=stream_name, fill=fill,
+                           handler_registry=handler_registry,
+                           handler_overrides=handler_overrides,
+                           plugins=self.plugins, **kwargs)
 
     def get_table(self, headers, fields=None, stream_name='primary',
                   fill=False,
--- a/databroker/broker.py
+++ b/databroker/broker.py
@@ -243,9 +243,11 @@ def _(key, db):
 
 
 class Broker(object):
-    def __init__(self, mds, fs, plugins=None, filters=None, es=None):
+    def __init__(self, mds, fs=None, plugins=None, filters=None):
         """
         Unified interface to data sources
+
+        Eventually this API will change to ``__init__(self, hs, es, **kwargs)``
 
         Parameters
         ----------
@@ -257,8 +259,6 @@ class Broker(object):
         filters : list
             list of mongo queries to be combined with query using '$and',
             acting as a filter to restrict the results
-        es : EventStoreRO, optional
-            If not provides, mds is expected to provide event storage.
         """
         self.hs = HeaderSourceShim(mds)
         self.es = EventSourceShim(mds, fs)
--- a/databroker/broker.py
+++ b/databroker/broker.py
@@ -12,7 +12,6 @@ from .core import (Header,
                    get_events as _get_events,
                    get_table as _get_table,
                    restream as _restream,
-                   fill_event as _fill_event,
                    process as _process, Images,
                    get_fields,  # for conveniece
                    ALL,
@@ -496,8 +495,8 @@ class Broker(object):
         handler_overrides : dict, optional
             mapping data keys (strings) to handlers (callable classes)
         """
-        _fill_event(self.fs, event, handler_registry=handler_registry,
-                    handler_overrides=handler_overrides)
+        return self.es(event, handler_registry=handler_registry,
+                       handler_overrides=handler_overrides)
 
     def get_events(self, headers, fields=None, stream_name=ALL, fill=False,
                    handler_registry=None, handler_overrides=None, **kwargs):
--- a/databroker/broker.py
+++ b/databroker/broker.py
@@ -486,7 +486,7 @@ class BrokerES(object):
         warnings.warn("Use .get_events() instead.")
         return self.get_events(headers, fill=fill)
 
-    def fill_event(self, event, inplace=False,
+    def fill_event(self, event, inplace=True,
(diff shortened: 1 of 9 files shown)


CJ-Wright:
I think this addresses the concerns from #104

danielballan:
I think this makes sense. I'm not sure I know what you mean by "mirror out" (in the Design Manifesto).

tacaswell:
This is what I ran back to my office to push right before we left yesterday.  I was interrupted mid-sentence previously.

tacaswell:
This has the commits from #100 in here now.

databroker/core.py
```diff
@@ -673,3 +735,97 @@ def _external_keys(descriptor, _cache=boltons.cacheutils.LRU(max_size=500)):
...
+
+        return self._start
+
+    @property
+    def stop(self):
+        while self._stop is None:
+            try:
+                self._get_next()
+            except StopIteration:
+                raise Exception("stream does not contain a start")
```
danielballan: start -> stop

databroker/core.py
```diff
@@ -384,14 +443,15 @@ def get_table(mds, fs, headers, fields=None, stream_name='primary', fill=False,
         return pd.DataFrame()
 
 
-def restream(mds, fs, headers, fields=None, fill=False):
+def restream(mds, fs, es, headers, fields=None, fill=False):
```
danielballan: The name of this function/method is now even more confusing/misleading than before. It *might* do more long-term good than harm to rethink and rename.

databroker/core.py
```diff
@@ -60,35 +62,46 @@ def fill_event(fs, event, handler_registry=None, handler_overrides=None):
                     event.data[data_key] = _fs.get_datum(value)
 
 
-class Header(doc.Document):
+@attr.s(frozen=True)
+class Header(object):
     """A dictionary-like object summarizing metadata for a run."""
 
+    _name = 'header'
```
danielballan: I gather that this is here for back-compat with the other doct variant. Can you confirm that its name was lowercase 'header'? I remember it being 'Header'.

doc/source/design_manifesto.rst
```diff
@@ -0,0 +1,115 @@
...
+ - a few accumulator/buffer objects to aid working with sequences of (name, document) pairs
+
+
+
+Random concerns
+===============
+
+ - should implement a global registry of known ``DataBroker`` /
+   components so that un-pickling a header does not recreate all of
+   the db connections.  We clearly do not have enough meta-classes.
```
danielballan: :imp:

doc/source/design_manifesto.rst
```diff
@@ -0,0 +1,115 @@
...
+   The need for this goes away when
```
danielballan: ?

doc/source/design_manifesto.rst
```diff
@@ -0,0 +1,115 @@
...
+   events via filtering / broadcasting
+  - if we do this, should probably give new uid to descriptor.  This
+    will require doubling down on the idea that for streams of
+    documents are always mixed types and of the form ``(name, doc)``
+  - we may also want to back off on the aggressive de-normalization of
+    the descriptors at every level.  Working always in one process the
+    cost of de-normalizing is low because we can share an object
+    (which is the reason that `doct.Document` is immutable), however
+    if we move to a model where these documents are streamed between
+    process (local or not) this can result in massive overheads.  This dumps
```
danielballan: This thought trailed off, so up until that point I agree.

doc/source/design_manifesto.rst
```diff
@@ -0,0 +1,115 @@
...
+ - not clear we are not going to end up with two worlds, a document
+   streaming one and a DataFrame based one.
```
danielballan: We should keep an eye on this, but I think maintaining an API for column-optimized access is desirable and to some degree unavoidable. We can't be unpacking a million dictionaries at HXN if we want to give reasonable perf.

databroker/core.py
klauer: Exception -> AnythingElseException...

databroker/core.py
tacaswell: https://github.com/NSLS-II/databroker/pull/107/files#diff-369696e385332d7139756bd5f292bd03L100

databroker/core.py
tacaswell: I agree, this would be replaced by `Header.events` (which I note is a bad name, maybe `Header.documents`?

databroker/core.py
danielballan: Ha. You said it so I didn't have to. I'm completely sold on your argument to move toward inheriting from package-specific base exceptions, btw. Not sure if this PR is the right place to start that crusade, but we should do it soon.

databroker/core.py
danielballan: :sheep:

tacaswell:
attn @NSLS-II/dama

databroker/core.py
```diff
@@ -833,3 +833,90 @@ def __iter__(self):
...
+
+    def streams_given_header(self, header):
+        return set(d['name'] for d in
+                   self.descriptors_given_header(header))
+
+    def descriptors_given_header(self, header):
+        return list(desc for desc in
+                    self.mds.descriptors_by_start(header.start['uid']))
+
+    def descriptors_for_stream(self, header, stream_name):
```
danielballan: for -> given

databroker/core.py
```diff
@@ -833,3 +833,90 @@ def __iter__(self):
                 yield doc
             else:
                 self.__stash_values(name, doc)
+
+
+class HeaderSourceShim(object):
```
danielballan: This is not how the mds API works. Instead you need `find_run_start`, `find_run_stop`, etc. The slicing and __call__ logic are implemented once in Broker, not per variant of MDS.

tacaswell: but I think this is how it should work, the object that wraps the start/stop collections should know these semantics.  As it is now we are insisting that `find_run_start` implement it exactly the way that makes sense for the mongo derived collections.

databroker/broker.py
```diff
@@ -260,9 +257,11 @@ def __init__(self, mds, fs, plugins=None, filters=None):
         filters : list
             list of mongo queries to be combined with query using '$and',
             acting as a filter to restrict the results
+        es : EventStoreRO, optional
+            If not provides, mds is expected to provide event storage.
         """
-        self.mds = mds
-        self.fs = fs
+        self.hs = HeaderSourceShim(mds)
+        self.es = EventSourceShim(mds, fs)
```
licode: Since es is passed as a parameter, maybe add
``` 
if es is None:
    self.es = ...
else:
...
```

databroker/broker.py
```diff
@@ -884,3 +902,38 @@ def _munge_time(t, timezone):
     """
     t = datetime.fromtimestamp(t)
     return timezone.localize(t).replace(microsecond=0).isoformat()
+
+
+class HeaderSourceShim(object):
```
licode: move this to core, the same as EventSourceShim?

databroker/broker.py
```diff
@@ -884,3 +902,38 @@ def _munge_time(t, timezone):
...
+        starts = tuple(self.mds.find_run_starts(**query))
+
+        stops = tuple(_safe_get_stop(self.mds, s) for s in starts)
+        return zip(starts, stops)
+
+    def __getitem__(self, k):
+        return search(k, self)
+
+
+def _safe_get_stop(mds, s):
```
licode: I thought stop was always saved.

licode:
LGTM, add minor comments.

databroker/broker.py
tacaswell: If you hit this _before_ stop comes out or things went sideways so badly it did not come out.

databroker/broker.py
tacaswell: This is here for circular import reasons.  Once the refatoring/rotating is done, the re-arrangement will be easier.

tacaswell:
@arkilic I think this is ready to try with your hdf5 version of `EventStore`.

arkilic:
@tacaswell sounds good. i have a working version, let's do it tmrw.

doc/source/api_changes.rst
```diff
@@ -1,11 +1,59 @@
...
-API changes
-***********
+=============
+ API changes
+=============
 
 Non-backward compatible API changes
 
+
+v0.7.0
```
danielballan: 0.8, right?

doc/source/api_changes.rst
```diff
@@ -1,11 +1,59 @@
...
+Explicitly passed mds/fs have been removed, instead relying on the
+DataBroker instance included in the header.
+
+Break up internal structure of databroker
+-----------------------------------------
+
+* The core functions that touch events have a new required argument, ``es``.
+  This does not affect the API of the ``Broker`` object, only the functions in
+  the ``core`` module.
+
```
danielballan: Also, note in the API changes that we will have to update all the BL configs to say `db.insert` instead of `db.mds.insert`.

doc/source/design_manifesto.rst
```diff
@@ -0,0 +1,131 @@
...
+--------------
+ - Attributes:
+   - Exactly 1 ``HeaderSource``
+   - ``EventSource`` list
+ - Methods:
+  - mirror out all search from ``HeaderSource``, converts to ``Header`` objects
+  - manage filters + search state
+  - drop everything else!
+  - insert?
+    - how to handle ES distribution.
```
danielballan: I'm comfortable filing this under, "we'll figure it out later."

doc/source/design_manifesto.rst
```diff
@@ -0,0 +1,131 @@
...
+   - ``Filestore`` collection(s) / object
+
+ - Methods:
+   - insert ``es.insert(name, doc)``
+   - get streams given a ``Header``
+     - ``es.streams_given_header(header) -> {stream_names}``
+
+   - get descriptors
+     - ``es.descriptors_given_header(header, **kwargs) -> [Descriptor]``
+     - ``es.descriptor_given_stream(header, stream, **kwargs) -> Descriptor``
```
danielballan: A stream can have more than one descriptor.

danielballan: I think this can probably be one method  (`descriptors()`) with an optional `stream` argument.

databroker/broker.py
```diff
@@ -232,24 +230,25 @@ def _(key, mds):
...
 
 
-class Broker(object):
-    def __init__(self, mds, fs, plugins=None, filters=None):
+class BrokerES(object):
+    def __init__(self, hs, es, plugins=None, filters=None):
         """
         Unified interface to data sources
 
+        Eventually this API will change to ``__init__(self, hs, es, **kwargs)``
```
danielballan: The only place plugins ever was used was CSX, and I don't think it is used actively there. Safe to remove `plugins` now if you want, especially since archiver integration is broken.

databroker/core.py
```diff
@@ -669,7 +423,478 @@ def _external_keys(descriptor, _cache=boltons.cacheutils.LRU(max_size=500)):
...
         ek = _cache[descriptor['uid']]
     except KeyError:
-        data_keys = descriptor.data_keys
+        data_keys = descriptor['data_keys']
         ek = {k: v.get('external', None) for k, v in data_keys.items()}
         _cache[descriptor['uid']] = ek
     return ek
+
+
+class DocBuffer:
```
danielballan: I reviewed this previously. Looks good.

databroker/core.py
```diff
@@ -669,7 +423,478 @@ def _external_keys(descriptor, _cache=boltons.cacheutils.LRU(max_size=500)):
...
+    into the implementations
+    '''
+    def __init__(self, mds, fs):
+        self.mds = mds
+        self.fs = fs
+
+    def insert(self, name, doc):
+        return self.mds.insert(name, doc)
+
+    def streams_given_header(self, header):
```
danielballan: rename to `stream_names` (*all* of these are `_given_header` -- no need to spell that out)

databroker/core.py
```diff
@@ -669,7 +423,478 @@ def _external_keys(descriptor, _cache=boltons.cacheutils.LRU(max_size=500)):
...
+        return set(d['name'] for d in
+                   self.descriptors_given_header(header))
+
+    def descriptors_given_header(self, header):
```
danielballan: rename to `descriptors`

databroker/core.py
```diff
@@ -669,7 +423,478 @@ def _external_keys(descriptor, _cache=boltons.cacheutils.LRU(max_size=500)):
...
+        return list(desc for desc in
+                    self.mds.descriptors_by_start(header.start['uid']))
+
+    def descriptors_given_stream(self, header, stream_name):
```
danielballan: merge with method above, set `stream_name=ALL` by default

databroker/core.py
```diff
@@ -669,7 +423,478 @@ def _external_keys(descriptor, _cache=boltons.cacheutils.LRU(max_size=500)):
...
+        return [d for d in self.descriptors_given_header(header)
+                if stream_name is ALL or d['name'] == stream_name]
+
+    def docs_given_header(self, header, stream_name=ALL,
```
danielballan: rename to `events`

databroker/core.py
```diff
@@ -669,7 +423,478 @@ def _external_keys(descriptor, _cache=boltons.cacheutils.LRU(max_size=500)):
...
+                if not event_data:
+                    # Skip events that are now empty because they had no
+                    # applicable fields.
+                    continue
+
+                yield 'event', ev
+
+        yield 'stop', header.stop
+
+    def table_given_header(self, header, stream_name,
```
danielballan: rename to `table`

databroker/broker.py
```diff
@@ -470,6 +490,10 @@ def fill_event(self, event, handler_registry=None, handler_overrides=None):
         """
         Populate events with externally stored data.
 
+        .. warning
+
+            This mutates the event's ``data`` field in-place
```
danielballan: This warning does not match with what happens below. Did you mean to pass `inplace=True` below?

danielballan:
Also, consider `s/in_place/inplace/g` for consistency with pandas.

databroker/core.py
tacaswell: all documents come out, recently renamed this from events -> doc ;)

databroker/broker.py
tacaswell: Would rather do that in a second pass, this still passes the plugin tests

databroker/core.py
danielballan: Ah, I see. Then the docstring needs to be updated.

databroker/broker.py
danielballan: :+1:

tacaswell:
I like the `_given_header` in the names.  In is very clear what is going on and that these take in headers.

These are methods on the inner details of the system,  I think clarity for developers is worth the verbosity here.

doc/source/api_changes.rst
tacaswell: It will still work, just warn at you.

tacaswell:
Fold in changes from #121

CJ-Wright:
Actually it seems that #121 causes the events to think that they are filled when they might not be?

PR #117, opened 2017-01-24
ENH: Add store_decorator based on @CJ-Wright's work.

danielballan:
This is based on @CJ-Wright's work [here](github.com/xpdAcq/redsky/pull/9).

Changes:
* Instead of instantiating the Writer classes anew each time an event is processed, instantiate them once per *run* (upon receiving a RunStart). It can be an internal detail of the Writer to decide whether to create a new Resource once per event (i.e., every time `.write()` is called), once per run (when `__init__` is called) or something in between.
* performance improvement for the case where nothing needs to be mutated
* naming changes and more docs/comments


databroker/broker.py
```diff
@@ -880,3 +880,76 @@ def _munge_time(t, timezone):
...
+                    # The writer writes data to an external file, creates a
+                    # datum record in the filestore database, and return that
+                    # datum_id. We modify fs_doc in place, replacing the data
+                    # values with that datum_id.
+                    for data_key, writer in writers.items()
+                        fs_uid = writer.write(fs_doc['data'][data_key])
+                        fs_doc['data'][data_key] = fs_uid
+
+                    doc.update(
+                        filled={k: True for k in fs_data_name_save_map.keys()})
```
CJ-Wright: This needs to be converted to `{k: True for k in writers.keys()}`

databroker/broker.py
```diff
@@ -880,3 +880,76 @@ def _munge_time(t, timezone):
...
+                    fs_doc = copy.deepcopy(doc)
+                else:
+                    fs_doc = doc  # for perf
+
+                if name == 'start':
+                    # Make a fresh instance of any WriterClass classes.
+                    writers = {data_key: cl(db.fs)
+                               for data_key, cl in in external_writers.items()}
+
+                if name == 'descriptor':
```
CJ-Wright: why `elif` below but not here?

databroker/broker.py
```diff
@@ -880,3 +881,152 @@ def _munge_time(t, timezone):
...
+
+                # The pristine doc is yielded.
+                yield name, doc
+
+        return wrapped_f
+
+    return wrap
+
+
+def event_map(stream_name, data_keys, provenance):
```
CJ-Wright: Is it possible to auto capture some of this provenance inside the `event_map`?

danielballan: Yes, likely some of it will be automatically captured: at least we can know the name of the function and that it was applied to events by `event_map`. Even though I think we have a good of some things that will be in provenance, a long discussion is needed to come up with a complete plan for its contents. I'm completely punting for now.

CJ-Wright: Also should this go into its own module? This is essentially a header function interface and we might have many of these for each kind of interface.

CJ-Wright: Punting is fair (I didn't know how much of my provenience stuff was worth duplicating)

danielballan: I agree we will want many. I think it's fine to dump them in `broker.py` for simplicity, at least for now.

databroker/broker.py
```diff
@@ -913,7 +912,7 @@ def wrapped_f(*args, **kwargs):
                 # doc will pass through unchanged; fs_doc may be modified to
                 # replace some values with references to filestore.
                 if external_writers:
-                    fs_doc = copy.deepcopy(doc)
+                    fs_doc = dict(doc)
```
CJ-Wright: This might be a problem?
I went with a deepcopy because when `fs_doc` got mutated it mutated the the yielded doc.
Does this fix this?

CJ-Wright: Never mind I think this works.

databroker/tests/test_pipelines.py
```diff
@@ -0,0 +1,89 @@
+import time
+import tempfile
+import os
+import uuid
+from functools import partial
+import numpy as np
+from databroker.broker import store_dec
```
CJ-Wright: I think you need to import the `event_map` decorator too

databroker/broker.py
```diff
@@ -880,3 +902,161 @@ def _munge_time(t, timezone):
...
+                                          data_keys=new_data_keys,
+                                          name=stream_name)
+                    yield 'descriptor', new_descriptor
+
+                elif (name == 'event' and
+                      doc_or_uid_to_uid(doc['descriptor']) == descriptor_uid):
+                    if run_start_uid is None:
+                        raise RuntimeError("Received Event before RunStart.")
+                    try:
+                        new_event = dict(doc)
```
CJ-Wright: We should rewrite the event from scratch. Just mutating it could cause all sorts of problems
eg I ran a scan with images and motor positions the images will be in the writers but the motor won't therefore the motor will be passed downstream as if we had just analyzed it, which we didn't. Also note that the time/timestamps will be incorrect.

databroker/broker.py
```diff
@@ -880,3 +902,161 @@ def _munge_time(t, timezone):
...
+                                         parents=[doc['uid']],
+                                         provenance=provenance)
+                    yield 'start', new_start_doc
+
+                elif name == 'descriptor' and doc.get('name') == stream_name:
+                    if run_start_uid is None:
+                        raise RuntimeError("Received EventDescriptor before "
+                                           "RunStart.")
+                    descriptor_uid = doc_or_uid_to_uid(doc)
+                    new_data_keys = dict(doc['data_keys'])
```
CJ-Wright: Should we be eating all the keys from the prior run, even the ones we aren't going to transform?

databroker/broker.py
```diff
@@ -1060,3 +1060,16 @@ def inner(stream):
                     yield 'stop', new_stop
         return inner
     return outer
+
+
+def header_io(db_in, db_out):
+    def outer(f):
+        def inner(header):
+            output_uids = []
+            stream = db_in.restream(header, fill=True)
```
CJ-Wright: Would it be possible to do this in a many header/stream case 
Streams = [restream for hdr in hdrs]

databroker/broker.py
CJ-Wright: Another issue is inheritance, if you have a multiple stream `event_map` what do you do with the events? Do you take on both event's data? What if they have the same keys?

CJ-Wright: To expand more on this a little. One of the main issues is this dereferenceing flattens the task graph. Take for example a temperature and image based event stream. At the end of the day you need 1/T and corrected images. These are separate processes and can be handled in their own separate nodes. However, propagating the data encourages users to depend on having the extra data in a single stream. This in turn causes reprocessing issues, if we need to reprocess the data for some reason (faulty image processing) we will also need to reperform the 1/T transformation somewhere in the stack, that where is now a sticky question because depending on where in the chain that transformation occurred either we need to propagate it a lot or a little (even more painful if instead of temperature the second data was another set of images with their own corrections)

ordirules: @CJ-Wright  I also agree with you and vote on that, should be `new_event = dict()`, safer that way. The run points to its parent anyway so the extra meta data is not lost.

Not sure how that's resolved, but as you mention [here] (https://github.com/CJ-Wright/databroker/pull/2) it may eventually make more sense by resolving it outside of db for now (so worry about headers, not the details of streams). Please let me know if I misunderstood anything.

CJ-Wright: The first part is spot on. I think we should work on the streams and headers (aka live and post collection) using the same code (if possible), developing two different code bases could be awful to maintain. I think the current plan is to use the `header_io` decorator to turn headers into streams as needed which I am on board with. But you are mostly correct that we might not have enough active use cases to properly abstract the more general (multi stream) problem into a decorator.

databroker/broker.py
```diff
@@ -880,3 +902,178 @@ def _munge_time(t, timezone):
...
+                if name == 'start':
+                    # Make a fresh instance of any WriterClass classes.
+                    writers = {data_key: cl(db.fs)
+                               for data_key, cl in external_writers.items()}
+
+                if name == 'descriptor':
+                    # Mutate fs_doc here to mark data as external.
+                    for data_name in external_writers.keys():
+                        # data doesn't have to exist
+                        if data_name in fs_doc['data_keys']:
```
tacaswell: Think you need to also copy the data_keys dict before you update it.

databroker/broker.py
```diff
@@ -880,3 +902,178 @@ def _munge_time(t, timezone):
...
+                        # data doesn't have to exist
+                        if data_key in fs_doc['data']:
+                            fs_uid = writer.write(fs_doc['data'][data_key])
+                            fs_doc['data'][data_key] = fs_uid
+
+                    doc.update(
+                        filled={k: False for k in external_writers.keys()})
+
+                elif name == 'stop':
+                    for data_key, writer in list(writers.items()):
```
tacaswell: Why the popping? This hits all of them so just clear it at the end?

databroker/broker.py
```diff
@@ -880,3 +902,178 @@ def _munge_time(t, timezone):
...
+        instantiated (possible multiple times) with the argument ``db.fs``.
+        If it requires additional arguments, use ``functools.partial`` to
+        produce a callable that requires only ``db.fs`` to instantiate the
+        ``WriterClass``.
+        """
+    if external_writers is None:
+        external_writers = {}  # {'name': WriterClass}
+
+    def wrap(f):
+        def wrapped_f(*args, **kwargs):
```
tacaswell: This looks like it assumes you will never get mixed event streams, that will not be true for 'live' data.

databroker/broker.py
licode: I think this is fine because data in descriptors doesn't take much space.

databroker/broker.py
```diff
@@ -880,3 +902,178 @@ def _munge_time(t, timezone):
...
+                    yield 'start', new_start_doc
+
+                elif name == 'descriptor' and doc.get('name') == stream_name:
+                    if run_start_uid is None:
+                        raise RuntimeError("Received EventDescriptor before "
+                                           "RunStart.")
+                    descriptor_uid = doc_or_uid_to_uid(doc)
+                    new_data_keys = dict(doc['data_keys'])
+                    for k, v in new_data_keys.items():
+                        new_data_keys[k].update(v)
```
licode: Can we simply do deepcopy here instead of one by one update?

databroker/broker.py
CJ-Wright: I meant this more in the context of the above manifesto advocating for not copying all the event data.

databroker/broker.py
ordirules: I agree, that's a good idea. I don't think our analysis pipelines so far are well defined yet on our end (CMS), and they can be quite complicated (lots of inputs, lots of outputs, sometimes input is db data, sometimes something else) so I think we may take the more argument based approach and leave this aside. For ex:
```python 
def process(*args, **kwargs):
    for arg in args:
        if isinstance(arg, Header):
            # do some reading replace arg with what header has
    # now run function stuff
```
etc. (of course this would be done with decorators acting on functions) The storing I imagine would just be one document (since an event is always just one function in my mind, even a 'stream' doesn't make sense since outside functions in another stream may eventually depend on an intermediate step in another 'stream'. It's more a huge interconnected graph.).
Later on, we can use our implementation to decide whether `event_map` (and permutations) are useful and build them in. That's our take for now, just to give context. We can hold off discussion on this for later, @danielballan is helping moderate our approaches.

I definitely see the use for this routine, it's very quick for incoming data. I think it is suitable. My only concern is how to save the documents. But perhaps if it's a batch processed stream, it should remain as a stream (and not something other, like an 'analysis' document). Sorry for the long comment. Might be better suited in an eventual discussion in person.

databroker/broker.py
CJ-Wright: I think that `new_data_keys` should be `data_keys` from the decorator kwargs.

databroker/broker.py
tacaswell: I would not worry too much about exactly how to orchestrate these things, think about building this from the inside out, not the outside in.

I would also start thinking about _very_ coarse processes to put in these 'boxes' and then save the results of.   To the point of, don't worry about connecting multiple steps together yet as that opens a whole other can of worms and brings us to the problem space that vistrails, orange, airflow, lugi, storm, .... are all working in and we do not want to re-develop that wheel if we do not have too.

databroker/broker.py
ordirules: For the record, that's my fault. in adding some changes I robotically made the same change here.
Should be `data_keys`, not `new_data_keys`

databroker/broker.py
CJ-Wright: I am not so certain that this will be so easy to do. Many of the multi-stream functions require data that takes a while to make, requires humans to make it, or needs to be reused many times or a combination of all 3 (eg detector calibration). I understand the concern about redoing the workflow problem and I have no intention to solve it. Many of the problems addressed by the workflow managers we don't need yet (parallelization, job management, caching, etc.). I think a good middle ground would be to hard code the pipelines. This could be accomplished by just having a massive function which takes in the data (most likely as a generator) and properly dispatches it to the sub-functions (all decorated to the hilt) most likely with liberal use of `itertools.tee` where copies of the streams are needed.

CJ-Wright: As a note on the original "don't mutate the event" discussion. Event mutation could cause issues when we go and try to bifurcate the streams using `itertools.tee` since this returns a shallow copy of the generator's contents any mutations that occur will be in the tee'ed stream thus causing issues when you were expecting one kind of data and get the analyzed result.

danielballan:
@ordirules

databroker/tests/test_pipelines.py
```diff
@@ -0,0 +1,112 @@
+import time
+import tempfile
+import os
+import uuid
+from functools import partial
+import numpy as np
+from databroker.broker import store_dec, event_map, header_io
+
+
+class NpyWriter:
```
ordirules: I think it might be nice to eventually have default writers in databroker, just as you do handlers. `NPYwriter`looks fine to me as it is now. Also, this used to be an 'up to the user' part of data acquisition (see README.md [here](https://github.com/NSLS-II/fileStore)).
`store_dec` is now breaking that assumption. I think it's nice in that it's completing a missing void. If you add these writers, I think you officially declare support for the storage (which as a user I like :) ). Just a thought....

databroker/broker.py
```diff
@@ -880,3 +902,178 @@ def _munge_time(t, timezone):
...
+                                    run_start=run_start_uid,
+                                    exit_status='success')
+                    descriptor_uid = None
+                    run_start_uid = None
+                    yield 'stop', new_stop
+        return inner
+    return outer
+
+
+def header_io(db_in, db_out):
```
ordirules: I'd be tempted to remove this. As a user, I am fine with being forced to think in terms of streams for more complex tasks. I can convert whatever object it is I use to store data to a compatible stream no problem. I think if you add this, you might force too many standards on yourselves.
I think @danielballan and @tacaswell  you were mentioning this today (correct me if I'm wrong), that you want to make sure we understand documents and see headers as a convenient cul-de-sac (sorry I like alliteration).
What do you think? Might require longer discussion.

danielballan:
This was a useful exercise, but it's not going to be merged. Since it's my PR I'm going to close it unilaterally. Thanks for the conversation, everyone!

CJ-Wright:
Many of the ideas from here were taken up by https://github.com/xpdAcq/SHED please feel free to contribute and/or get code from there!

jrmlhermitte:
I agree (decorators get messy very quickly).
Thanks for the interesting discussions about it!

PR #121, opened 2017-02-07
Fill

### LINKED ISSUE #120: ENH: fill filled when filling events
It seems that the `filled` event key does not get switched when pulling from the databroker with `restream(hdr, fill=True)`. Would it be possible to fix this?

CJ-Wright:
Let me fill in this gap in the code base.
Closes #120 (One might even say that this _fulfills_ the issue)
(Its too bad we don't have a reviewer named Phil. Then I think I would have filled this PR with all the potential uses of fill)


CJ-Wright:
@tacaswell @danielballan This fails because of the timezone stuff but otherwise this passes with flying (full?) colors.

PR #123, opened 2017-02-08
Fill

CJ-Wright:
add tests to make certain that the events are different


CJ-Wright:
@arkilic @danielballan We found this error earlier today it seems to be due to metadatastore, which I have put in a PR to fix (although it does not fix it on my machine). Do you have any insight into the cause?

PR #124, opened 2017-02-17
Tst rootless export

tacaswell:


tacaswell:
Restarted CI to run against updated portable mds

Issue #126, opened 2017-02-22
list of headers should be reversed


CJ-Wright:
The list of headers should be reversed, currently they are in chronological order.
This means that `db[-1]` and `db()[-1]` give back different answers, this is confusing.

tacaswell:
I think this is a bug in portable mds, they should be coming out in ascending time order.

CJ-Wright:
I think I was getting this from the mongo backed system.

CJ-Wright:
Closing as a feature not bug

PR #127, opened 2017-02-22
TST: test reverse

CJ-Wright:
does `db[-1] ==db()[-1]`?


CJ-Wright:
@tacaswell per issue #126

tacaswell:
attn @danielballan @arkilic I recall long discussions about this in mds, but apparently remember the result wrong.

What is the sorting of headers in a return from `db()`?

danielballan:
It is never safe to assume that because `db.filters` do not apply to slicing syntax, intentionally. I'm not sure about the sorting: I'd have to check the source.

CJ-Wright:
@danielballan I guess my point was more along the lines of
a) without filters that should be true
b) The headers returned from db() are in reverse chronological order?

CJ-Wright:
trigger ci

tacaswell:
The test coverage changes are clearly spurious.

I am now very confused as these tests are verifying the opposite of the bug report in #126

danielballan:
@arkilic can you confirm that sorting?

arkilic:
Queried headers are returned in [descending order](https://github.com/NSLS-II/metadatastore/blob/f107e0527070548bd6cea3d9b90f91d205478ee9/metadatastore/core.py#L892-L893)

Last headers (mds.find_last) are returned in [descending order](https://github.com/NSLS-II/metadatastore/blob/f107e0527070548bd6cea3d9b90f91d205478ee9/metadatastore/core.py#L892-L893)

Whatever we get from mongo seems in descending order with regards to headers. In other words, mongo driver will provide an ascending list: ``db[-1] == db()[0]`` will return `True`

To me, this is fairly intuitive since most database drivers return descendingly ordered data in this fashion. The most recent data is the first element of the list (since you want to access to the latest data first, ordering it in this fashion makes sense).

CJ-Wright:
My issue with this is that if you make a list of headers then how you access the "latest" header depends on how you created the list (query vs. get_item).

tacaswell:
@arkilic But the tests are passing and seem to contradict your comment....

arkilic:
I don't know what's going on with tests in this PR but...

screen shot 2017-03-01 at 3 02 47 pm

danielballan:
It's possible that not all flavors of MDS are consistent?

CJ-Wright:
Sorry for the confusion, there were some issues with the tests, they are failing as expected now.

tacaswell:
https://github.com/NSLS-II/metadatastore/pull/204 and https://github.com/NSLS-II/metadatastore/pull/196/commits/dd49c85e3882ca78db87077292f510ecad744d42 are where the sorting came in (with an sign error on the first try).

Further as Arman points out, we apply the same sorting (newest first) to the `db[-5:]` so `db[-5:][-1] != db[-1]`.

The question is 'newest first' vs 'chronological order' for consuming the search result.  I think we settled on 'newest first' because sitting at the BL the more recent headers are more likely to be the relevant ones.

I am not concerned about confusion around `db[-1]` vs `db()[-1]`.  I am actually more concerned that `__call__` returns a list not a generator.

tacaswell:
I am in favor of closing both this and #126 as wont-fix.

attn @stuartcampbell

CJ-Wright:
If `__call__` gave back a generator I'd be happy with that

PR #138, opened 2017-03-02
Finish new API

danielballan:
This adds some convenience methods and properties previously discussed and refactors the Archiver code as an EventSource. The databroker API is getting a lot nicer!

All the old tests pass. The only API break here is that using `plugins` fails noisily instead of failing silently.

Don't merge yet -- I want to add new tests for the new methods.


databroker/core.py
```diff
@@ -111,19 +111,88 @@ def __str__(self):
         return doct.vstr(self)
 
     def __len__(self):
-        return 3
+        if 'stop' in self:
```
tacaswell: This should always be 3, or `len(h) != len(list(h))`

danielballan: I believe that if there is no `stop` that `len(list(h))` is 2. Am I confused?

tacaswell: Yes

```
In [12]: h = bc.Header(db=None, start={})

In [13]: len(hg
  File "<ipython-input-13-a35bbbf60b6a>", line 1
    len(hg
          ^
SyntaxError: unexpected EOF while parsing


In [14]: len(h)
Out[14]: 3

In [15]: list(h)
Out[15]: ['start', 'descriptors', 'stop']

In [16]: h.stop
Out[16]: {}

In [17]: 
```

tacaswell:
I am going to take over this branch and do things like force push to it, that cool with you?

PR #142, opened 2017-03-09
FIX: iterating over key instead of items

tacaswell:


tacaswell:
odd, coverage did not change....

Issue #154, opened 2017-04-01
AttributeError: module 'attr' has no attribute 's'


ordirules:
So there's a potential subtle error from a package conflict that might happen to outside users (I say outside because it won't happen with the right conda recipe).

Basically, if you do this:
```pip install attr```
you'll install the wrong `attr` library and see this:

```python
In [1]: import databroker

/home/lhermitte/research/projects/databroker/databroker/core.py in <module>()
     30 
     31 
---> 32 @attr.s(frozen=True)
     33 class Header(object):
     34     """A dictionary-like object summarizing metadata for a run."""

AttributeError: module 'attr' has no attribute 's'
```

The correct way to resolve this is basically:
```
pip uninstall attr
pip install attrs
```
`attr` installs `attr-0.2.0` and `attrs` installs `installed attrs-16.3.0`.

Maybe add some check or something? Figured I'd share since it's something to look out for and had me scratching my head for the past 15 min... :-|

CJ-Wright:
Yea there is a problem with the current conda-forge standard for attrs.
https://github.com/conda-forge/attrs-feedstock/pull/1
I am working on it, although help would be appreciated.
(Something tells me that there are going to be some very chilly yaks at the end of this)

danielballan:
As of #143, attrs should be installed when databroker is installed. Nonetheless it might be worth doing a specific check for this package at import time and providing an customized ImportError, given the weird attrs vs attr confusion.

danielballan:
See #155

ordirules:
`pip install attr` will still have `import attr` be successful. I don't know if that resolves it
Maybe do something like:
```python
try:
    attr.s
except AttributeError("Error wrong attr library imported. Please uninstall attr and install attrs")
```

I think could be wrong... I think this naming conflict should maybe be resolved higher up.

tacaswell:
Closing as fixed by the requirements updates, leaving associated PR open.

PR #155, opened 2017-04-03
ENH: Provide a specific error if attr is missing warning about package name.

danielballan:


danielballan:
Closes #154

ordirules:
`pip install attr` will still have `import attr` be successful. I don't know if that resolves it
Maybe do something like:
```python
try:
    attr.s
except AttributeError("Error wrong attr library imported. Please uninstall attr and install attrs")
```

I think could be wrong... I think this naming conflict should maybe be resolved higher up.

danielballan:
Good point. Fixed. I think this is the best we can do without asking attr or attrs to break their naming, which seems unlikely.

ordirules:
great, looks good to me.
yeah, it's a shame that nomenclature conflict somehow went through...

thanks!

databroker/__init__.py
```diff
@@ -6,6 +6,13 @@
 
 
 try:
+    import attr.s
```
tacaswell: `s` is not an importable thing, needs to be `import attr; attr.s`

tacaswell:
I am not sold on this.  We now have `attrs` in setup.py so the correct one should be installed with data broker.

If the user then goes and `pip install attr` later, that is on the user and I am not sure we should be responsible for saving them.

stuartcampbell:
I think it's not a necessity, but it does seem a nicety for those folks who are not intimate with the software and the various `attr[s]` packages.  For such a small change, it could help someone figure out this potential issue in the future if it arises again.

danielballan:
Point take that fixing the setup.py should be enough that we never see this error again. But I do find it exceptionally confusing. I think it's worth this special check.

It seems that's the consensus above.

PR #156, opened 2017-04-20
TST: test double access of table

CJ-Wright:
Ran across an issue where the databroker returns empty tables if you ask it for the same table twice. I am trying to test it.
Note that this is using a mongoquery backend although I think this may be a problem with `AD_TIFF`.

### DIFF
--- a/databroker/tests/conftest.py
+++ b/databroker/tests/conftest.py
@@ -2,24 +2,27 @@ import sys
 import pytest
 
 from databroker.tests.utils import (build_sqlite_backed_broker,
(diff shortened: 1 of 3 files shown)


CJ-Wright:
@danielballan @tacaswell @arkilic Thoughts?
I tried using JSON after the sqlite issues.

CJ-Wright:
@chiahaoliu if you can, please take a look at this.

CJ-Wright:
@justcalamari

danielballan:
Fixed by #323 (we think!) -- please reopen if this is actually another issue.

PR #158, opened 2017-05-02
API: Broker.__call__ returns iterable Results object. Listify ret val in tests.

danielballan:


databroker/tests/test_broker.py
```diff
@@ -52,21 +52,21 @@ def test_get_events(db, RE):
     RE.subscribe('all', db.insert)
     uid, = RE(count([det]))
     h = db[uid]
-    assert len(list(db.get_events(h))) == 1
-    assert len(list(h.stream())) == 1 + 3
+    assert len(list(list(db.get_events(h)))) == 1
```
tacaswell: Do not need to double list

danielballan: force-pushed a revision

tacaswell:
Can you add a note to API changes?

PR #160, opened 2017-05-19
Enh stream name in image

tacaswell:
only the last 2 commits are different than #159 

putting this here so we do not lose it.


danielballan:
The functions that this is extending are now deprecated.

PR #162, opened 2017-05-23
API: Make HeaderSource lazy so that Results can be lazy as well.

danielballan:
#158 did not actually finish the job of making `Broker.__call__` lazy! As @tacaswell mused aloud in the car this morning, there is a call to `tuple()` in the new `HeaderSource` code that materializes the mongo query before it gets to `Broker.__call__`. This PR finishes the job. It is technically API changes on `HeaderSource`. I advocate tagging (yet another) micro release and pushing this to the floor so that databroker-browser can be responsive.

I also threw in a commit to refactor the dependencies into a requirements.txt and add a missing one. This is a little bit unnecessary but it does no harm.

### DIFF
--- a/databroker/broker.py
+++ b/databroker/broker.py
@@ -1,4 +1,5 @@
 from __future__ import print_function
+import itertools
 import warnings
 import six  # noqa
 import uuid
@@ -253,7 +254,8 @@ class Results(object):
         self._data_key = data_key
 
     def __iter__(self):
-        for start, stop in self._res:
+        self._res, res = itertools.tee(self._res)
+        for start, stop in res:
             header = Header(start=start, stop=stop, db=self._db)
             if self._data_key is None:
                 yield header
--- a/databroker/broker.py
+++ b/databroker/broker.py
@@ -1099,10 +1099,8 @@ class HeaderSourceShim(object):
             _format_time(kwargs, self.mds.config['timezone'])
             query = {'$and': [{}] + [kwargs] + filters}
 
-        starts = tuple(self.mds.find_run_starts(**query))
-
-        stops = tuple(_safe_get_stop(self, s) for s in starts)
-        return zip(starts, stops)
+        starts = self.mds.find_run_starts(**query)
+        return ((s, _safe_get_stop(self, s)) for s in starts)
 
     def __getitem__(self, k):
         return search(k, self)
--- a/databroker/tests/test_broker.py
+++ b/databroker/tests/test_broker.py
@@ -664,3 +664,15 @@ def test_export_size_smoke(broker_factory, RE):
     db1.fs.register_handler('RWFS_NPY', ReaderWithFSHandler)
     size = db1.export_size(db1[uid])
     assert size > 0.
+
+
+@py3
+def test_results_multiple_iters(db, RE):
+    RE.subscribe('all', db.insert)
+    RE(count([det]))
+    RE(count([det]))
+    res = db()
+    first = list(res)  # First pass through Result is lazy.
+    second = list(res)  # The Result object's tee should cache results.
+    third = list(res)  # The Result object's tee should cache results.
+    assert first == second == third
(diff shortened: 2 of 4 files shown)


tacaswell:
The results object should probably cache the headers so it can be iterated more than once?

danielballan:
Makes sense. See new commit (includes test).

danielballan:
@ordirules As it turns out (and as you suspect, given the slow response time), `Results` was not lazy _ at all_. This should make it much faster. It's still true that there is some ahead-of-time results-gathering done on the mongo side (for example, mongo knows the length of the results set) but we should save a lot of time by not moving all the results into Python greedily.

ordirules:
ah thanks!
Actually, it makes me think of one possible slightly related feature request. Would be be possible to have an asynchronous version in the `Future`? (pun intended) I'm thinking mainly, is it possible to have mongo return some Queue that is populated as it obtains search results for potentially long searches?
Something similar to `buffer` [here] (https://github.com/mrocklin/streams/blob/master/streams/core.py#L500).
Just curious.

danielballan:
Seems doable and reasonable to integrate with databroker. I would start by checking with pymongo provides anything like this itself. Care to work on a PR?

ordirules:
Sure, sounds fun, I'll give it a shot. Maybe in a day or two, have a few things to finish currently. First thing will be to see if this speeds things up. It will also likely be pymongo dependent for now. But yes, definitely interested in this!

danielballan:
P.S. I verified in production that `db()` is indeed lazy now: it takes 30 microseconds to run at CSX, where there are tens of thousands of headers.

ordirules:
thanks! great! :-)

One small issue with this (not sure if it's related, but it's related to new db version). The processing of `Header` is now more reliant on it being a `doct`.
Anyway, databroker (in `core.py`, see function `docs_given_header` as just one example) now accesses start through `header.start` instead of `header['start']`.

That's fine, except that when I pass headers to `get_events`, I pass `dict(header)` so that it's serializable (removes the mongo db connection).
I think to resolve this either:
1. Need to treat this use case where user transforms `dict(header)`
2. Allow some other way to serialize headers.

I need this, because I am running in a distributed regime, and I have cases where I would like to pass an header to a worker node and have the worker node do the events retrieval when necessary. I pass `dict(header)` and a string that helps the worker identify the database to access and it connects to the database through its local dictionary that contains the db connections.

I can open PR to highlight the line differences maybe? However, my suggested approach may not be the best, it would mainly be to highlight. let me know thanks!

ordirules:
@arkilic may have some ideas

danielballan:
We try to use `header['start']` internally everywhere, so I would say that should be fixed.

ordirules:
thanks, could you please ping me when it's fixed (or if i can help)
Currently, my quick workaround is:
```python
    from doct import Document
    header = Document('header', **dict(db[uid]))

```

danielballan:
See https://github.com/NSLS-II/databroker/pull/168 for a fix. I think we should still make Headers serializable, but that requires making Broker, FileStore, MDS, and all the "portable" variants of these serializable, so it will take some coordinated effort.

ordirules:
sounds good :-) thanks!

ordirules:
Looks like motor resolves this: https://motor.readthedocs.io. So it is possible, good news.

I need to understand this and get a working implementation in some test mongodb before thinking about databroker. This may not be so straightforward. I don't have much experience with this. Perhaps I try it out as I mentioned and make a case for it if useful first?

arkilic:
Motor usually lags behind pymongo in terms of development and generally slower than pymongo unless there are a few thousand clients but it'll definitely do what you'd like it to do. If you can think of a way that's not directly mongo oriented and applicable to the other backends it'd be neater. Let me know if you need a hand :)

stuartcampbell:
My only concern would be would be this would introduce a hard dependency on mongo and might be problematic for other back ends (as @arkilic mentioned).

PR #166, opened 2017-05-23
TST: add next test

CJ-Wright:
I assume I should be able to use next on the databroker `Results`.


tacaswell:
It is an iterable, (meaning you can do `iter(res)`) not an iterator (meaning you can call `next(res)`).

PR #168, opened 2017-05-24
FIX: Consistently access header fields using __getitem__ not __getattr__.

danielballan:
Verified that I got them all with grep:

```
$ git grep header.start
$ git grep header.stop
$ git grep header.descriptors
tests/test_broker.py:    assert [] == header.descriptors
```

and added a test for defense against future changes.


ordirules:
+1 for the test :-)

PR #171, opened 2017-05-26
FIX: If resource_path is absolute, patch it up on the way out

tacaswell:
- do not insert an invalid resource with both a non-trivial root and
   an absolute resource_path

attn @CJ-Wright 

No test because I am racing the train into jamacia!

### DIFF
--- a/databroker/_core.py
+++ b/databroker/_core.py
@@ -1730,10 +1730,26 @@ class BrokerES(object):
                                           **copy_kwargs)
                 file_pairs.extend(fps)
                 res = self.reg.resource_given_uid(uid)
+                r_path = res['resource_path']
+                if os.path.isabs(r_path):
+                    if not res['root']:
+                        r_path = os.path.relpath(r_path, os.path.sep)
+                        if new_root is None:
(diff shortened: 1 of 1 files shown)


danielballan:
Needs rebase.

databroker/broker.py
```diff
@@ -849,7 +849,7 @@ def export(self, headers, db, new_root=None, copy_kwargs=None):
             an instance of databroker.Broker class, which has
             filestore (fs) and metadatastore (mds) attributes
             that will be the target to export info
-        new_root : str
+.        new_root : str
```
mrakitin: Is it a random dot, or some undocumented trick?

tacaswell: that is a random dot

tacaswell:
rebased, git failed to track moving everything to `_core.py`.

danielballan:
Looks like a real test failure related to this PR:

https://travis-ci.org/bluesky/databroker/jobs/569966262#L1208

danielballan:
Bump

Issue #175, opened 2017-07-11
yield readonly or copies of docs


CJ-Wright:
I've found that when using stream that we get the live dictionaries, which if edited cause problems. Can we either yield docts or copies?

danielballan:
This was fixed in #219. See comment in #176.

PR #176, opened 2017-07-11
TST: test breaks on fill if yielded docs are modified

CJ-Wright:
See #175

### DIFF
--- a/databroker/tests/test_broker.py
+++ b/databroker/tests/test_broker.py
@@ -396,6 +396,33 @@ def test_stream(db, RE):
 def _stream(method_name, db, RE):
     RE.subscribe('all', db.insert)
     uid = RE(count([det]), owner='Dan')
+    s = getattr(db, method_name)(db[uid])
+    name, doc = next(s)
+    assert name == 'start'
+    assert 'owner' in doc
+    name, doc = next(s)
+    assert name == 'descriptor'
+    assert 'data_keys' in doc
+    last_item  = 'event', {'data'}  # fake Event to prime loop
+    for item in s:
+        name, doc = last_item
+        assert name == 'event'
+        assert 'data' in doc  # Event
+        last_item = item
+    name, doc = last_item
+    assert name == 'stop'
+    assert 'exit_status' in doc # Stop
+
+
+@py3
+def test_stream_copy(db, RE):
+    _stream_copy('restream', db, RE)
+    _stream_copy('stream', db, RE)  # old name
+
+
+def _stream_copy(method_name, db, RE):
+    RE.subscribe('all', db.insert)
+    uid = RE(count([det]), owner='Dan')
     s = getattr(db, method_name)(db[uid], fill=True)
     name, doc = next(s)
     assert name == 'start'
@@ -413,6 +440,7 @@ def _stream(method_name, db, RE):
     name, doc = last_item
     assert name == 'stop'
     assert 'exit_status' in doc # Stop
+
 
 
 @py3
--- a/databroker/tests/test_broker.py
+++ b/databroker/tests/test_broker.py
@@ -396,13 +396,14 @@ def test_stream(db, RE):
 def _stream(method_name, db, RE):
     RE.subscribe('all', db.insert)
     uid = RE(count([det]), owner='Dan')
-    s = getattr(db, method_name)(db[uid])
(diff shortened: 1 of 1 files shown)


danielballan:
Sorry, I've looked at this for a couple minutes and I don't understand what this change is for. I'm sure I just need more coffee, but in the meantime could you give me a leg up?

CJ-Wright:
This test shows that if you modify a dict which comes out of databroker that downstream documents can suffer. This is a rather extreme example, we entirely get rid of the `data_keys` by replacing them with `None` which causes the events to not properly load their data.

danielballan:
Can this be demonstrated in a separate test instead of modifying our basic test of the `stream` method?

CJ-Wright:
Sure, I've put it into the new commit.

danielballan:
OK, so this PR is not intended to be merged, but rather intended to illustrate the point that you shouldn't mutate descriptors and then expect the Events that depend on them to load properly? I agree, but I'm not sure what the action item is here.

CJ-Wright:
I guess I was hoping that we could return a copy or a doct, but I don't know which.

CJ-Wright:
Or more to the point, we shouldn't be yielding things that if modified break all the things.

danielballan:
I see. I think it should be up to the caller to make the copy if they need to mutate something, rather than make a copy and force _all_ callers to pay that performance penalty.

CJ-Wright:
Don't we do that with turning events into `Doct`?

danielballan:
Coming back to your original point,

> This test shows that if you modify a dict which comes out of databroker that downstream documents can suffer. 

I would phrase this as, if the user for some reason alters a descriptor to make it inconsistent with the contents of its events, then the descriptor and events become inconsistent. The `Doct` provides *some* protection because its top-level keys are immutable, but it is not tamper-proof. Ultimately, users who modify the documents must do so self-consistently.

Does stream actually return a `dict` as in `type(doc) is dict`?

CJ-Wright:
```python
for name, doc in stream:
    print(name, type(doc)

=======================
start <class 'doct.Document'>
descriptor <class 'dict'>
event <class 'doct.Document'>
stop <class 'doct.Document'>
```

CJ-Wright:
Just as some context for this, I finally got tired of mongo vs. python tests not working properly so I wrote this
```python
def clean_databroker(doc):
    doc = dict(doc)
    for k2 in ['run_start', 'descriptor']:
        if k2 in doc.keys():
            doc[k2] = doc[k2]['uid']
    if '_name' in doc.keys():
        doc.pop('_name')
    if 'data_keys' in doc.keys():
        doc['data_keys'] = dict(doc['data_keys'])
        for k, v in doc['data_keys'].items():
            doc['data_keys'][k] = dict(doc['data_keys'][k])
            if 'external' in v.keys():
                doc['data_keys'][k].pop('external')
    return doc
```
Which should clear up most of the python vs. mongo inconsistencies.

tacaswell:
What do you mean 'mongo vs python'?

CJ-Wright:
python gives back tuples, if you put the same dict into mongo and get it back those tuples will be lists. This causes any kind of "did what I put in come back out as I expected" kind of test to fail.

danielballan:
That can't be fixed. The information of whether it was a tuple or list (or numpy array) is lost. Databroker intentionally does not store that language-specific information.

CJ-Wright:
The purpose of the above code was to normalize both the python data and the mongo data so that they can be compared successfully.

tacaswell:
@CJ-Wright How did you get that stream which had dictionaries in it?

CJ-Wright:
`data_keys` is a dictionary isn't it?

CJ-Wright:
Wait sorry, I think I misinterpreted your question. `hdr.stream()` I think gives out a descriptor document as a dict. [Relevant line in pySHED, aka redsky](https://github.com/xpdAcq/redsky/pull/14/files#diff-7671fba4fb53f10c943e2cca2851c540R38)

danielballan:
Got it. The bug you illustrated in [your comment above](https://github.com/NSLS-II/databroker/pull/176#issuecomment-315090058) was fixed in #219, and a test was added. Additionally, notice that the return types are now configurable: https://nsls-ii.github.io/databroker/api.html#advanced-controlling-the-return-type

PR #203, opened 2017-07-28
Refactor of HDF5 with addition of stats

stuwilkins:
This is an attempt to refactor the HDF5 handlers. The effort is to:

* Incorporate SWMR into the same handler, negating the need for different SPECs to support SWMR
* Add the timestamp and stats handlers. These should be added to Bluesky also. 

Currently this is WIP, but a test notebook can be found here:

http://nbviewer.jupyter.org/gist/stuwilkins/1466f3ed5b67d146799dde8b8953fe3d


stuwilkins:
Note:

This PR brings some questions:

* We should define the HDF5 layout via an AreaDetector HDF5 XML file to sync the __hardcoded__ keys with the files generated by areaDetector. 
* We should define naming for the stats as these are defines in the NDAttributes XML file for the stats plugin

stuwilkins:
Updated notebook example:

http://nbviewer.jupyter.org/gist/stuwilkins/c52d4ba096dcf38eeed34c55e0ebe737

@tacaswell would you mind having a look and giving some comments. Thanks!

stuwilkins:
Closed as superseded by #254

PR #204, opened 2017-07-31
Add pipeline to Image (FramesSequence) to process on per image basis

stuwilkins:
@danielballan, @tacaswell 
cc @stuartcampbell 

I wanted to add this as a discussion point. I tracked the problem with pipelines, and it comes from handlers that return on a per datum (point) basis a `FramesSequence`. I think this adds the possibility of adding a pipeline at the correct point. 

Thoughts? 

I also though about cleaning up the `__repr__` also.

### DIFF
--- a/databroker/broker.py
+++ b/databroker/broker.py
@@ -598,7 +598,7 @@ class BrokerES(object):
             return pd.DataFrame()
 
     def get_images(self, headers, name, handler_registry=None,
-                   handler_override=None):
+                   handler_override=None, pipeline=None):
         """
         Load images from a detector for given Header(s).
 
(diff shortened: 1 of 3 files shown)


stuwilkins:
Attn @ambarb and @cmazzoli This would fix the issue with number of images at CSX

stuwilkins:
Notebook showing usage

http://nbviewer.jupyter.org/gist/stuwilkins/f790c0d3982a7163e7170164b1d3263f

stuwilkins:
Just to add, this is the other way to do this:

https://github.com/NSLS-II-CSX/csxtools/pull/54

One could easily deconstruct the `FramesSequence` object returned and reconstruct with the pipeline. 

Any thoughts @danielballan @tacaswell @ambarb @cmazzoli

PR #195, opened 2017-07-18
TST/FIX: Ensure that numpy arrays are properly ingested.

danielballan:
- Test 1D and 3D arrays can go into RunStart and into Event['data'].
- These are the tests that must pass before
  https://github.com/NSLS-II/bluesky/pull/674 can be safely merged.

PR #211, opened 2017-08-03
API: do not de-reference foreign Keys 

tacaswell:


danielballan:
THS PR HS A RLY CRYP. TLT.

databroker/headersource/base.py
```diff
@@ -7,10 +7,7 @@
 
 
 class MDSROTemplate(object):
-    def __init__(self, config, version=1, auth=False):
-        if auth:
-            raise NotImplementedError("This implementation of MDS does not "
-                                      "support authentication.")
+    def __init__(self, config, version=1):
```
danielballan: While you're here, do you want to [move version into the config dict](https://github.com/NSLS-II/databroker/issues/185)?

databroker/headersource/base.py
```diff
@@ -1,8 +1,16 @@
 from __future__ import (absolute_import, division, print_function,
                         unicode_literals)
 from . import core
+import numpy as np
 
 
+def _sanitize_np(val):
```
danielballan: In #195 (awaiting fixes before merge) I add `sanitize_np` a public function in utils. Leaving this here is fine for this PR, but removing the leading underscore from the name would make the rebase easier. Optional!

tacaswell: There is probably a longer discussion to be had about where this should be done (would be better to push lower for ex the hdf5 backend?

danielballan: It's possible that this is actually safe to remove once #195 is in, which does handle it lower at least in places. Let's revisit there. If you don't want to delete the underscore, that's fine, just trying to save 5 seconds of hassle while I have just burned it here. :- )

tacaswell:
Both done.

danielballan:
:clap:

PR #214, opened 2017-08-04
ENH: add method for only changing the root of a resource without

CJ-Wright:
moving the files

This is useful when you export the data, copy it to a new machine, and
try to load it up (which causes an error since the roots aren't the
same).

### DIFF
--- a/databroker/resource_registry/handlers.py
+++ b/databroker/resource_registry/handlers.py
@@ -594,3 +594,33 @@ class Xspress3HDF5Handler(HandlerBase):
 
     def __repr__(self):
         return '{0.__class__.__name__}(filename={0._filename!r})'.format(self)
+
+
+class DebugHandler(HandlerBase):
+    print_attrs = ['_path', '_template', '_filename']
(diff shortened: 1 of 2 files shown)


CJ-Wright:
@danielballan @tacaswell @arkilic ready for review!

tacaswell:
Can you implement this as a flag on `change_root` instead?

CJ-Wright:
Sure, what is the best way to get the root of a resource during retrieval? I'd like to be able to print it for debugging.

tacaswell:
During normal retrieval or weird 'I want to move things around' retrieval?

CJ-Wright:
Normal retrieval, although if you need to debug the entries in filestore usually things end in editing the resource.

tacaswell:
This is addressed by https://github.com/NSLS-II/databroker/pull/216/commits/b9e438a746c499c9d51abce514ac4834ac84db5d, closing.

danielballan:
:+1:

PR #215, opened 2017-08-07
Api rr cleanup

tacaswell:
there is more to come, but might be worth looking at and merging this as-is before I make it worse.


danielballan:
Judging from the Travis tests it looks like some py3-isms may have sneaked in.

databroker/resource_registry/base.py
```diff
@@ -251,9 +251,6 @@ def get_spec_handler(self, resource):
         root = resource.get('root', '')
         root = self.root_map.get(root, root)
         if root:
-            if os.path.isabs(rpath):
```
danielballan, line 232: :+1:

tacaswell:
This is probably ready for a commit-by-commit review, but I am concerned by the coverage drop on `db.hs.core`

danielballan:
When the dust settles, the end of this week might be a good time to sprint on vertical integration tests.

Issue #194, opened 2017-07-15
New configuration system spec draft


danielballan:
Goal: Support multiple databrokers in the same process so it is easy to switch between beamlines or combine data between beamlines or juggle different databrokers for raw and analyzed data.

Plan, previously outlined with @tacaswell and @stuartcampbell on a whiteboard:

Deprecate `databroker.DataBroker` and `databroker.db` with noisy warnings. (We probably have to maintain them for a long time or we will break ~1000 user notebooks.)

The new encouraged usage will be:

```python
from databroker import Broker
db = Broker.named('chx')
```

relying on a config file something like:

```yaml
# /etc/databroker/chx.yml
headersource:  # mapped to one dict
  module: databroker.headersource.mongo
  class: MDSRO
  config:
    database: metdatastore-production-v1
    host: xf23id-broker
    port: 27017
    timezone: US/Eastern
    
event_sources:  # mapped to a list of dicts (in this case, a list with one element)
  - module: databroker.headersource.mongo   # in the future, this will a proper EventSource
    class: MDSRO
    config:
      database: metdatastore-production-v1
      host: xf23id-broker
      port: 27017
      timezone: US/Eastern
    filestore:
      module: databroker.filestore.mongo
      class: FileStoreRO
      config:
        database: filestore-production-v1
        host: xf23id-broker
        port: 27017
```

New broker code would look something like:

```python
SEARCH_PATH = ('/etc/databroker', os.path.expanduser('~/.config/databroker/'))

def lookup_config(name):
    # Look for a file called {name}.yml in the SEARCH_PATH.
    # Open and parse it into a dict
    return config

from importlib import import_module

class Broker:
    ...
    @classmethod
    def from_config(cls, config):
        modname = config['headersource']['module']  # a string
        clsname = config['headersource']['class']  # a string
        config = config['headersource']['config']  # a dict
        mod = import_module(modname)
        cls = getattr(mod, clsname)
        headersource = cls(config)
        # similar for event_source(s) and their filestores
        return cls(headersource, *eventsources)

    @classmethod
    def named(cls, name):
        return cls.from_config(lookup_config(name))
```

CJ-Wright:
[...]

PR #218, opened 2017-08-09
ENH: New config system

danielballan:
As outlined in #194. For now we're targeting the current Broker structure (1 metadatstore + 1 assets). In the future we'll reorganize into (1 header source + N event sources + N assets).


databroker/broker.py
```diff
@@ -170,6 +172,56 @@ def __iter__(self):
                         break
 
 
+CONFIG_SEARCH_PATH = (os.path.expanduser('~/.config/databroker/'),
```
tacaswell: can we also include the conda env etc?

databroker/tests/test_config.py
```diff
@@ -0,0 +1,58 @@
...
+        'class': 'MDSRO',
+        'config': {
+            'host': 'localhost',
+            'port': 27017,
+            'database': 'mds_database_placholder',
+            'timezone': 'US/Eastern'}
+    },
+    'assets': {
+        'module': 'databroker.assets.mongo',
+        'class': 'FileStoreRO',
```
tacaswell: This needs to be `RegistryRO` I think.

databroker/broker.py
danielballan: Do we want to special-case conda here? We did that previously, but in retrospect it seems like kind of a hacked thing to do. Is there any reason we need it, given our plans for the config deployment?

tacaswell: I think a clean way to do this is :

```python
import sys
from pathlib import Path
local_etc = Path(sys.executable).parents[1] / 'etc' / 'databroker'
```

which will work without activation and with venvs.

danielballan: Is there a py2-friendly equivalent of `Path(sys.executable).parents[1]`?

danielballan: Oh, I see what that does. Easy enough.

tacaswell:
:+1: Can you add a test of the legacy stuff?

Issue #191, opened 2017-07-14
Revisit our choice to use Doct and document our reasoning


danielballan:
@tacaswell and I last revisited this, by phone, about six months ago. One can make a case for returning plain dicts, as RunEngine does, but ultimately we decided it wasn't a strong enough case. The original motivations for creating doct were immutability (of which it does a less-than-perfect job) and adding getattr-based access as a synonym for getitem-based access, which is convenient but has downsides.

If we decide to stick with doct (which I suspect we will) then we should document our reasoning somewhere.

danielballan:
We are strongly leaning toward dropping doct. One original reason for needing it becomes moot when documents have uids instead of references to other documents. This change would also break dot access (but tab complete would still work on getitem access). Users handle documents interactively much less than we expected.

Concerns, anyone?

CJ-Wright:
I'm just happy that this means that all the documents will be consistent :smile:

danielballan:
Documenting an in-person design discussion with @tacaswell:

Goals:
* Create a path to fully removing `doct` and just representing documents as plain dictionary objects.
* Do not break dot access in this release. That is, continue to support `event.data` as a synonym for `event['data']`... at least for now.

Proposed path:

1. Remove the special `_name` key from doct, so that the `Doct` object becomes nothing more than a dict subclass that supports `__getattr__` (dot access) as a synonym for `__getitem__` (square bracket lookup). Removing `_name` is an API change, but in practice it is not expected to break any user code.
2. Make all internal Broker components (mds and fs variants) represent documents as plain dict objects internally, thus saving on performance.
3. Wherever Broker returns document to the user, as in `header.stream()`, pass the document through a new `Broker.prepare_hook(name, doc)` method on the way out. By default, `prepare_hook` will convert the dict into a `Doct`, thus maintaining API continuity for the user. But the user can change `prepare_hook` to, for example, make a deep copy instead, and return a plain dict.
4. Make `Doct.__getattr__` warn that dot access may be removed in a future release. See how much push-back we get about removing dot access.
5. If removing doct entirely (and breaking dot access) still feel like a good idea, change the default behavior of `prepare_hook` in a future release.

danielballan:
I followed this path in #219. Instead of changing `doct` itself, I wrapped it in a subclass `DeprecatedDoct`.  When you try to use dot access to access items (as in `event.data`), it warns we _might_ remove that in the future and that that should use `event['data']` instead.

[...]

PR #219, opened 2017-08-09
ENH: Make the return type configurable with Broker.prepare_hook.

danielballan:
See discussion in #191

### DIFF
--- a/databroker/broker.py
+++ b/databroker/broker.py
@@ -258,7 +258,7 @@ class DeprecatedDoct(doct.Document):
             # This is not a standard dict attribute.
             # Warn that dot access is deprecated.
             warnings.warn("Dot access may be removed in a future version."
-                 "instead of .{0}".format(key))
+                          "Use [{0}] instead of .{0}".format(key))
         return res
 
 
--- a/databroker/broker.py
+++ b/databroker/broker.py
@@ -1,5 +1,6 @@
 from __future__ import print_function
 from importlib import import_module
+from functools import partial
 import itertools
 import warnings
 import six  # noqa
@@ -9,7 +10,7 @@ import pytz
 import logging
 import numbers
 import requests
-from doct import Document
+import doct
 import pandas as pd
 import sys
 import os
@@ -161,7 +162,9 @@ class Results(object):
     def __iter__(self):
         self._res, res = itertools.tee(self._res)
         for start, stop in res:
-            header = Header(start=start, stop=stop, db=self._db)
+            header = Header(start=self._db.prepare_hook('start', start),
+                            stop=self._db.prepare_hook('stop', stop),
+                            db=self._db)
             if self._data_key is None:
                 yield header
             else:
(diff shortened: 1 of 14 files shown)


databroker/broker.py
```diff
@@ -231,6 +232,32 @@ def load_component(config):
...
+              'stop': 'Run Stop'}
+
+
+def wrap_in_doct(name, doc):
+    return doct.Document(DOCT_NAMES[name], doc)
+
+
+class DeprecatedDoct(doct.Document):
+    def __getattr__(self, key):
+        if key not in dir(dict):
```
tacaswell: call this once and stash it as class attr to avoid doing it everytime?

databroker/tests/test_core.py
```diff
@@ -3,13 +3,13 @@
 from databroker.core import Header
 
 
-def test_header_dict_conformance():
+def test_header_dict_conformance(db):
```
tacaswell, line 7: ah, we need this so that have have access to the hook?

danielballan: Correct.

danielballan: The test is busted because of `_name`. It looks like you have future plans for this test but I can't figure out what they are. What's the right "fix" here?

danielballan: OK, I figured this out.

databroker/core.py
```diff
@@ -61,14 +61,13 @@ def from_run_start(cls, db, run_start):
...
-                                           'run_start')
+            run_stop = mds.stop_by_start(run_start_uid)
         except mds.NoRunStop:
             run_stop = None
 
-        d = {'start': run_start}
+        d = {'start': db.prepare_hook('start', run_start)}
         if run_stop is not None:
-            d['stop'] = run_stop
+            d['stop'] = db.prepare_hook('stop', run_stop)
```
tacaswell: So that we have consistent (but empty) objects should this be `d['stop'] = db.prepare_hook('stop', run_stop or {})`?

databroker/tests/test_broker.py
```diff
@@ -761,3 +765,95 @@ def test_events(db, RE):
...
+
+    RE.subscribe(db.insert)
+    uid, = RE(count([det]))
+
+    for h in (db[uid], list(db())[0]):
+        for doc in _get_docs(h):
+            assert not isinstance(doc, DeprecatedDoct)
+            assert not isinstance(doc, doct.Document)
+            assert isinstance(doc, dict)
+            assert isinstance(doc, dict)
```
tacaswell: making _very_ sure it is a `dict`

danielballan: Can't be too careful these days!

databroker/core.py
danielballan: Reasonable. We used to say that if the RunStop was missing we just wouldn't put the key in. Now it looks like we make it an empty dict by default? Fair enough....

PR #223, opened 2017-08-10
API/DOC: Some API cleanup previously discussed and new API docs

### LINKED ISSUE #207: Rethink Header.events and Header.stream
Believe or not, @tacaswell is advocating for a name change. I'm just the note-taker here. Adjust my reputation among my peers accordingly.

* Header.events -> Header.stream
* Header.stream -> Header.documents

Also, the new Header.stream should *require* its stream_name arg and disallow `ALL`. That is, it should return rows from a table. That analogy implies that the rows have homogeneous columns.

danielballan:
The databroker documentation still needs a lot of love, but this
is a good start.

Also, closes #207

### DIFF
--- a/databroker/broker.py
+++ b/databroker/broker.py
@@ -262,16 +262,16 @@ def wrap_in_deprecated_doct(name, doc):
 
 
 class BrokerES(object):
-    def __init__(self, hs, *event_sources):
-        """
-        Unified interface to data sources
+    """
+    Unified interface to data sources
 
-        Parameters
-        ----------
-        hs : HeaderSource
-        *event_sources :
-            zero, one or more EventSource objects
-        """
+    Parameters
+    ----------
+    hs : HeaderSource
+    *event_sources :
+        zero, one or more EventSource objects
+    """
+    def __init__(self, hs, *event_sources):
         self.hs = hs
         self.event_sources = event_sources
         # Once we drop Python 2, we can accept initial filter and aliases as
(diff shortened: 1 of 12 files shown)


databroker/broker.py
```diff
@@ -351,12 +405,42 @@ def clear_filters(self, **kwargs):
...
 
     def __getitem__(self, key):
-        """Do-What-I-Mean slicing"""
+        """
+        Search runs based on recently, unique id, or counting number scan_id.
+
+        This function returns a :class:`Header` object (or a list of them, if
+        the input is a list or slice). Each Header encapsulates the metadata
+        for a run -- start time, instruments used, and so on, and provides
+        methods for loading the data. In
```
tacaswell: Missing sentence.

doc/source/api.rst
```diff
@@ -0,0 +1,179 @@
...
+===============
+
+You can instantiate a :class:`Broker` by passing it a dictionary of
+configuration or by providing the name of a configuration file on disk.
+
+.. autosummary::
+   :toctree: _as_gen
+   :nosignatures:
+
+   Broker.from_config
```
licode, line 31: I think this is a much convenient interface for configuration setup at different beamlines.

databroker/broker.py
```diff
@@ -418,53 +502,58 @@ def dynamic_alias(self, key, func):
...
+        queries using the MongoDB query syntax.
 
         Parameters
         ----------
         text_search : str, optional
             search full text of RunStart documents
-        start_time : time-like, optional
-            Include Headers for runs started after this time. Valid
-            "time-like" representations are:
-                - float timestamps (seconds since 1970), such as time.time()
```
tacaswell, line 435: is'time-like' defined someplace else?

licode:
waiting for the doc. good work 👍

databroker/broker.py
licode: something like this:  
https://github.com/NSLS-II/databroker/blob/master/databroker/headersource/core.py#L767

databroker/broker.py
```diff
@@ -510,7 +599,7 @@ def fill_event(self, event, inplace=True,
     def get_events(self, headers, fields=None, stream_name=ALL, fill=False,
```
tacaswell, line 650: This function now very problematic.   It should probably take exactly 1 header and switch to a single stream.

PR #227, opened 2017-08-11
ENH:Enforce uniqueness of uid field in runstart

arkilic:
See https://github.com/NSLS-II/databroker/issues/190

### DIFF
--- a/databroker/headersource/core.py
+++ b/databroker/headersource/core.py
@@ -515,7 +515,12 @@ def insert_run_start(run_start_col, run_start_cache,
     col = run_start_col
     run_start = dict(time=time, uid=uid, **kwargs)
     apply_to_dict_recursively(run_start, sanitize_np)
-
+    try:
+        if run_start_given_uid(uid, run_start_col,
+                               run_start_cache):
+            raise ValueError('A RunStart with given uid exists %s', uid)
+    except NoRunStart:
+        pass
     col.insert_one(run_start)
 
     _cache_run_start(run_start, run_start_cache)
--- a/databroker/tests/test_mds.py
+++ b/databroker/tests/test_mds.py
@@ -174,6 +174,24 @@ def test_insert_run_start(mds_all):
     check_for_id(run_start_mds)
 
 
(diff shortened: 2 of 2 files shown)


tacaswell:
Can you also add a test for this?

databroker/headersource/core.py
```diff
@@ -515,7 +515,12 @@ def insert_run_start(run_start_col, run_start_cache,
     col = run_start_col
     run_start = dict(time=time, uid=uid, **kwargs)
     apply_to_dict_recursively(run_start, sanitize_np)
-
+    try:
+        if run_start_given_uid(uid, run_start_col,
```
tacaswell, line 519: Wouldn't be better to push this down into the `insert_one`?  Doing the search might be a cheap process.

ghost: I originally put it there then changed my mind. Do events ever get inserted via insert_one()? If not, it is definitely a cheap process.

ghost: The reasoning here is to address the issue in one place for all backends. The "brute-force" check is exactly the same operation the database will perform when it is enforcing uniqueness.

danielballan: True, but it will do so without the extra network call, right? And, in the case of MongoDB, this means it will be checking redundantly.

danielballan:
(P.S. This PR came in when master was broken. Rebase and the tests should pass.)

databroker/headersource/core.py
ghost: If the RunStart with this uid already exists in the cache, no network call would be necessary. This is a very likely scenario if uids are generated by some user script. In terms of Mongo, the network call we are worrying about here is literally a few milli seconds since uids are already indexed. Are we that sensitive to latency? In case of sqlite, hdf5 etc. backends this is a call to some database/file loaded in memory.

stuartcampbell: Would it be worth doing a test to see what the actual performance is for a real system rather than second guessing ourselves ?

PR #230, opened 2017-08-11
BUG: need to format time first on filters

licode:
This may fix the issue. Just need more tests. 

This is related to https://github.com/NSLS-II/databroker/issues/119

### DIFF
--- a/databroker/headersource/shim.py
+++ b/databroker/headersource/shim.py
@@ -19,14 +19,14 @@ class HeaderSourceShim(object):
     def __call__(self, text_search=None, filters=None, **kwargs):
         if filters is None:
             filters = []
+        else:
+            [_format_time(v, self.mds.config['timezone']) for v in filters]
         if text_search is not None:
             query = {'$and': [{'$text': {'$search': text_search}}] + filters}
         else:
             # Include empty {} here so that '$and' gets at least one query.
             _format_time(kwargs, self.mds.config['timezone'])
-            [_format_time(v, self.mds.config['timezone']) for v in filters]
             query = {'$and': [{}] + [kwargs] + filters}
-
         starts = self.mds.find_run_starts(**query)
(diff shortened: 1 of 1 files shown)


databroker/headersource/shim.py
```diff
@@ -24,7 +24,11 @@ def __call__(self, text_search=None, filters=None, **kwargs):
         else:
             # Include empty {} here so that '$and' gets at least one query.
             _format_time(kwargs, self.mds.config['timezone'])
-            query = {'$and': [{}] + [kwargs] + filters}
+            formatted_filters = []
+            for v in filters:
+                _formatted_time(v, self.mds.config['timezone'])
```
danielballan: I think you want `_format_time` here, not `_formatted_time`. Also, this might be a nice place to use a list comprehension.

danielballan: Also, notice that `filters` is also used in the text_search case above. The `filters` should be "formatted" above if/else statement that we're inside of here, so that both cases work.

licode: thanks for the comments.

databroker/headersource/shim.py
```diff
@@ -24,6 +24,7 @@ def __call__(self, text_search=None, filters=None, **kwargs):
         else:
             # Include empty {} here so that '$and' gets at least one query.
             _format_time(kwargs, self.mds.config['timezone'])
+            [_format_time(v, self.mds.config['timezone']) for v in filters]
```
danielballan: This is still inside the `else` block, meaning that the case under `if` above will be working with un-formatted filters. See my previous comment about moving this line up above the `if` so that it applies to both cases.

licode: good catch.

danielballan:
Thanks, @licode!

licode:
sure thing

PR #236, opened 2017-08-11
Api fill changes

tacaswell:


databroker/broker.py
```diff
@@ -1069,6 +1069,50 @@ def export_size(self, headers):
...
+                ev = self.prepare_hook('event', ev)
+            data = ev['data']
+
+            desc_id = ev['descriptor']
+            for k in fill_map:
+                data[k] = registry_map[(desc_id, k)].retrieve(data[k])
+
+            yield ev
+
+    def fill_table(self, evt_tbl, descriptor, fields=None, inplace=False):
```
danielballan: Note to self to revisit the argument name `evt_tbl` later. Carry on for now!

databroker/broker.py
```diff
@@ -690,15 +698,23 @@ def get_events(self, headers, fields=None, stream_name=ALL, fill=False,
...
-                        fields=fields,
-                        handler_registry=handler_registry,
-                        handler_overrides=handler_overrides)
-                for nm, ev in gen:
-                    if nm == 'event':
+                        fill=False,
+                        fields=fields)
+                gen = (ev for nm, ev in gen if nm == 'event')
+                # dirty hack!
+                es = self.assets['']
```
danielballan: Is there are reason to reuse `es` here? Isn't this a Registry, not an event source?

databroker/broker.py
```diff
@@ -1069,6 +1085,59 @@ def export_size(self, headers):
 
         return total_size * 1e-9
 
+    def fill_events(self, events_iterable, descriptors,
```
danielballan: events_iterable -> events

databroker/broker.py
```diff
@@ -1069,6 +1085,59 @@ def export_size(self, headers):
...
+                                .retrieve(d_id))
+                    filled[dk] = d_id
+
+            yield ev
+
+    def fill_table(self, evt_tbl, descriptor, fields=None, inplace=False):
+        """Fill a table
+
+        """
+
```
danielballan: raise NotImplementedError for this release?

databroker/core.py
```diff
@@ -311,11 +304,20 @@ def events(self, stream_name='primary', fill=False, **kwargs):
 
         >>> events = list(h.events())
         """
-        for name, doc in self.documents(stream_name=stream_name,
-                                        fill=fill,
-                                        **kwargs):
-            if name == 'event':
-                yield doc
+        def inner():
```
danielballan: You can simplify this by scrapping `inner` entirely and using `self.events`

databroker/broker.py
```diff
@@ -1069,6 +1085,59 @@ def export_size(self, headers):
...
+                yield ev
+            return
+        elif fields:
+            fields = set(fields)
+        registry_map = {}
+        fill_map = defaultdict(list)
+        for d in descriptors:
+            fill_keys = set()
+            desc_id = d['uid']
+            for k, v in d['data_keys'].items():
```
danielballan: six.iteritems

databroker/broker.py
```diff
@@ -1069,6 +1085,59 @@ def export_size(self, headers):
...
+
+        for ev in events_iterable:
+            if not inplace:
+                ev = _sanitize(ev)
+                ev = copy.deepcopy(ev)
+                ev = self.prepare_hook('event', ev)
+            data = ev['data']
+            filled = ev['filled']
+            desc_id = ev['descriptor']
+            for k, v in fill_map.items():
```
danielballan: six.iteritems

databroker/broker.py
```diff
@@ -1069,6 +1085,59 @@ def export_size(self, headers):
...
+                    fields=True, inplace=False):
+        """Fill a sequence of events
+
+        """
+        if fields is True:
+            fields = None
+        elif fields is False:
+            # if no fields, we got nothing to do!
+            for ev in events_iterable:
+                yield ev
```
danielballan: Explicitly comment here that we assume we don't need prepare_hook because this event has already had that applied in the first place.

databroker/core.py
tacaswell: This is in `self.events`

databroker/broker.py
```diff
@@ -718,7 +717,7 @@ def get_events(self, headers, fields=None, stream_name=ALL, fill=False,
                         yield self.prepare_hook('event', ev)
 
     def get_documents(self, headers, fields=None, stream_name=ALL, fill=False,
-                      handler_registry=None, handler_overrides=None):
+                      handler_registry=None):
```
danielballan: Did you mean to remove `handler_registry` also? It is not being passed through.

databroker/broker.py
```diff
@@ -1087,16 +1087,62 @@ def export_size(self, headers):
 
         return total_size * 1e-9
 
-    def fill_events(self, events_iterable, descriptors,
-                    fields=True, inplace=False):
+    def fill_events(self, events, descriptors, fields=True, inplace=False):
         """Fill a sequence of events
 
+        This method will be used both inside of other `DataBroker`
```
danielballan: All `DataBroker` should be `Broker`. `DataBroker` is the name of the old singleton instance.

databroker/broker.py
```diff
@@ -1130,13 +1130,27 @@ def fill_events(self, events, descriptors, fields=True, inplace=False):
...
         """
+        # create the processing generator
+        proc_gen = self._fill_events_coro(descriptors,
+                                          fields=fields, inplace=inplace)
+        # and prime it
+        proc_gen.send(None)
+
+        try:
+            for ev in events:
+                yield proc_gen.send(ev)
```
danielballan, line 1186: ooooo

databroker/broker.py
tacaswell: :frowning_face:

databroker/tests/test_broker.py
```diff
@@ -173,14 +173,6 @@ def test_get_events_filtering_field(db, RE):
 
 
 @py3
-def test_deprecated_api(db, RE):
-    RE.subscribe(db.insert)
-    uid, = RE(count([det]))
-    h, = db(uid=uid)
-    assert list(db.get_events(h))
```
danielballan: `db.get_events` is not fully removed.

tacaswell: Look at the previous 2 commits, I removed `fetch_events` and `find_headers` in 2 steps.  This test now covers existing functionality that is tested else where (without the misleading name).

databroker/broker.py
```diff
@@ -614,18 +622,7 @@ def __call__(self, text_search=None, **kwargs):
...
-        return self(**kwargs)
-
-    def fetch_events(self, headers, fill=True):
-        "This function is deprecated."
-        warnings.warn("Use .get_events() instead.", stacklevel=2)
-        return self.get_events(headers, fill=fill)
-
-    def fill_event(self, event, inplace=True,
-                   handler_registry=None, handler_overrides=None):
+    def fill_event(self, event, inplace=True, handler_registry=None):
```
CJ-Wright, line 625: Is there a partial `fill_event` that only fills one piece of data

tacaswell: see `fill_events`

CJ-Wright: maybe `fill_event(self, event, inplace=True, field=None, handler_registry=None)` 
if `field=None` fill everything else fill only that field?

CJ-Wright: Can we just expose the rest of the API? I only have a single event document. If this method is not long for this world can we put some sort of deprecation warning?

danielballan: I will refrain from closing https://github.com/NSLS-II/databroker/issues/213 until @tacaswell decide whether to deprecate `fill_event`, extend it with a fields parameter, or leave it as is.

PR #238, opened 2017-08-12
TST: munge time broken

### LINKED ISSUE #237: _munge_time broken
https://travis-ci.org/xpdAcq/xpdAn/jobs/263914210#L1159

It seems that datetime is no longer imported.

CJ-Wright:
Closes #237 
Or at least adds a test which breaks.
@arkilic

### DIFF
--- a/databroker/tests/test_broker.py
+++ b/databroker/tests/test_broker.py
@@ -11,13 +11,14 @@ import uuid
 from datetime import datetime, date, timedelta
 import itertools
 from databroker.broker import (wrap_in_doct, wrap_in_deprecated_doct,
-                               DeprecatedDoct)
+                               DeprecatedDoct, _munge_time)
 import doct
 import copy
 
 import pytest
 import six
 import numpy as np
+import pytz
 
 if sys.version_info >= (3, 0):
     from bluesky.examples import (det, det1, det2, Reader, ReaderWithRegistry,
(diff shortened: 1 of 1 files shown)


danielballan:
A PR that "Closes #237" would have added the missing import.

In this case, however, the right fix is just to delete `_munge_time`, which was moved to `databroker.eventsources.archiver` (along with that vanished `datetime` import) but never deleted here.

Given that `_munge_time` isn't even used internally, I'm very curious how you discovered this.

danielballan:
Closing in favor of #242. Sorry to discard this work, CJ. You'd have to have read our minds to know what the right fix was here. :- D

CJ-Wright:
Its fine, this PR was (as a few others were) designed to highlight an error. I find that PRs with breaking tests are the best way to provide a bug report since they are run against a neutral system (and PRs usually grab your attention :smiling_imp:) and give an example of expected behavior. I kinda wish that _munge_time was more accessible since we use it in some xpdAn/xpdAcq code to produce more human readable date/times. Is there a more accessible function for that?

CJ-Wright:
As to your question of how I found it; I'm me.
I most likely saw when it in to the code base originally. It is used in my implementation of [`beamtime_dates`](https://github.com/xpdAcq/xpdAn/blob/master/xpdan/fuzzybroker.py#L131) as a short cut way to make nice human readable dates.

danielballan:
> I kinda wish that _munge_time was more accessible since we use it in some xpdAn/xpdAcq code to produce more human readable date/times. Is there a more accessible function for that?

It's a two-liner; I suggest vendoring it in xpdAcq directly.

CJ-Wright:
Fair enough

PR #241, opened 2017-08-13
FIX: Use pkg_resources correctly.

### LINKED ISSUE #240: file resources broken?
It seems that a pip install from master does not capture the schemas?
https://travis-ci.org/CJ-Wright/xpdAn/jobs/263936602#L1531

`pip install https://github.com/NSLS-II/databroker/zipball/master#egg=databroker`
```python
from databroker.tests.conftest import db
```
yields error
`FileNotFoundError: [Errno 2] No such file or directory: '/home/travis/mc/envs/testenv/lib/python3.5/site-packages/databroker/assets/schemas/AD_HDF5_resource.json'`

danielballan:
Fixes #240 and adds a text. This bug cannot be reproduced with 'develop'
installation, which is likely why it wasn't immediately caught.

Thanks for the report, @CJ-Wright.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,7 +50,7 @@ install:
       pip install https://github.com/NSLS-II/bluesky/zipball/master#egg=bluesky
     fi;
   - 'pip install https://github.com/NSLS-II/event-model/zipball/master#egg=event_model'
-  - pip install -ve .
+  - pip install .
 
 script:
   - source activate testenv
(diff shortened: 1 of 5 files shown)


danielballan:
Travis failing on that intermittent timezone bug, but this is good to go otherwise.

tacaswell:
can we get pytest to play nice with a non-source install on travis?

danielballan:
Yes we can! New commit does this. The 3.5 build passed; 3.6 hit a random server failure... rerunning...

PR #242, opened 2017-08-14
CLN: Remove old _munge_time, which now lives in databroker.eventsource.archiver.

danielballan:
Closes #238


tacaswell:
failures are due to time zones, merging.

PR #246, opened 2017-08-14
ENH: Make it safe to insert events with a 'filled' key.

danielballan:
A prerequisite for https://github.com/NSLS-II/bluesky/issues/661

An event with ``filled={'some_key': False}`` will have its filled
item stripped off.

An event with ``filled={'some_key': 'some_datum_id'}`` will be
inserted with ``data={'some_key': 'some_datum_id'}`` instead of
the filled data being used.

### DIFF
--- a/databroker/headersource/base.py
+++ b/databroker/headersource/base.py
@@ -485,7 +485,7 @@ class MDSTemplate(MDSROTemplate):
                                            **kwargs)
 
     def insert_event(self, descriptor, time, seq_num, data, timestamps, uid,
-                     validate=False):
+                     validate=False, filled=None):
         """Create an event in metadatastore database backend
 
         .. warning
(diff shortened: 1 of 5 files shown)


CJ-Wright:
This is pretty awesome! I was going to ask for this with the pipelines. Does this assume that the uid is already a valid asset token?

CJ-Wright:
Will detectors start to give back both data and an asset uid?

danielballan:
> Does this assume that the uid is already a valid asset token?

I'm not sure I understand the question. Yes? If by 'uid' you mean 'datum_id'.

> Will detectors start to give back both data and an asset uid?

Detectors write data to disk directly and return an asset datum_id in their readings, as they have from the beginning.

danielballan:
I cherry-picked these commits into #236 to take advantage of some fixes there.

PR #247, opened 2017-08-14
TST: use random port in all client/server tests

### LINKED ISSUE #221: make client/server tests more reliable
The databroker tests with the client randomly fail on starting the server

https://travis-ci.org/NSLS-II/databroker/jobs/263097538

We should either shorten the existing timeout and then re-try starting the server or watch the stdout/return from the process to know if it has started or failed.

tacaswell:
closes #221


danielballan:
It would be better to use `socket` to get a random _free_ port, as opposed to a random port which may not be free. But this should reduce the number of server failures by a lot so we can handle that later.

PR #252, opened 2017-08-15
FIX: Fixed location of handlers, was pulling from filestore

stuwilkins:
@tacaswell @danielballan @stuartcampbell 

I noticed that the current `register_builtin_handlers` was pulling from `filestore`. This confused me for a while because it was pulling from the old filestore package and not the handlers packaged with `databroker`. 

This fixes that. 

However there still needs to be a call to auto-register the handlers......


danielballan:
Ha! Good catch.

PR #256, opened 2017-08-15
BUG: filter name issue

licode:
Duplicated names are shown in the filter.

### DIFF
--- a/databroker/broker.py
+++ b/databroker/broker.py
@@ -361,7 +361,7 @@ class BrokerES(object):
         self.event_sources = event_sources
         # Once we drop Python 2, we can accept initial filter and aliases as
         # keyword-only args if we want to.
-        self.filters = []
+        self.filters = {}
         self.aliases = {}
         self.event_source_for_insert = self.event_sources[0]
         self.prepare_hook = wrap_in_deprecated_doct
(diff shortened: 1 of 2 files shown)


databroker/tests/test_broker.py
```diff
@@ -389,6 +389,11 @@ def test_filters(db_empty, RE):
 
     assert header['start']['uid'] == ken_calib_uid
 
+    db.clear_filters()
+    db.add_filter(start_time='2017')
+    db.add_filter(start_time='2017')
```
tacaswell: what does this do if you now add `db.add_filter(start_time='2018')`?

licode: OK, I think it is better to replace 2017 with 2018.

danielballan: I think you are misunderstanding Tom's question. This test is good as is for testing that duplicate are ignored. He is suggesting that you check that this doesn't make _all_ things ignored, only duplicates.

licode: then we will have two start_time as 2017 and 2018. I can add more tests to double check.

licode: I think we still need to consider the case 
```
db.add_filter(start_time='2017')
db.add_filter(start_time='2017', stop_time='2018). 
```
So basically we will raise error if there are duplicated keys. Users can clean the filters and search again. Also per discussion with @tacaswell

databroker/broker.py
```diff
@@ -433,7 +433,12 @@ def add_filter(self, **kwargs):
         :meth:`Broker.clear_filters`
 
         """
-        self.filters.append(dict(**kwargs))
+        d = dict(**kwargs)
+        for val in self.filters:
+            for k,v in val.items():
```
danielballan: Since you don't need the keys, `val.values()` would be better here.

databroker/broker.py
```diff
@@ -433,7 +433,12 @@ def add_filter(self, **kwargs):
...
+                if k in d.keys():
```
danielballan: `k in d` is equivalent to `k in d.keys()` and faster in Python 2 because it avoids creating an intermediate list.

licode: thanks, good to know.

databroker/broker.py
```diff
@@ -433,7 +433,12 @@ def add_filter(self, **kwargs):
...
+                    raise ValueError('Filter key already exists.')
```
tacaswell: Can you include in the error message what the filter it conflicts with is and how to fix the problem?

tacaswell:
Replaced by #269

PR #258, opened 2017-08-15
DOC: Make API doc links work in release notes.

danielballan:


danielballan:
Self-merging this trivial sphinx PR. One random Travis failure on server startup.

PR #261, opened 2017-08-15
ENH: Made HDF5 and Mongo option extras.

### LINKED ISSUE #224: Update setup.py
We planned to use the new pip feature of selectors to install partial dependencies. That way, users who only want to use the simple sqlite backends, not a full MongoDB deployment, won't need to install pymongo.

```
pip install databroker  # minimal deps
pip install databroker[all]  # install everything
pip install databroker[mongo]  # minimal + pymongo (but no hdf5)
pip install databroker[hdf5]  # minimal + hdf5 (but no pymongo)
```

This should get done before we tag.

stuartcampbell:
As per described in #224 - Made 'mongo' and 'hdf5' options `pip` extras.

closes #224


danielballan:
This is so satisfying.

PR #262, opened 2017-08-16
WIP: Reinstante _repr_html_

### LINKED ISSUE #133: Restore _repr_html_ on new implementation of Header
It got deleted in the refactor.

danielballan:
Work in progress. Do not merge.

Copying over the `_repr_html_` from doct was enough to
get the basics right, but lists aren't working because we
probably used to rely on recursion from doct itself.

My attempt to make this work correctly led me to

```python
TypeError: Tried to call non recursive loop.  Maybe you forgot the 'recursive' modifier.
```

and I'm giving up for now.

Closes #133 (once it's ready).


danielballan:
Thanks @tacaswell for your jinja2-fu. I pushed additional commits to remove a bug (introduced by my core/broker/utils reshuffle) and add a smoke test of `_repr_html_()`.

danielballan:
Some context for @stuartcampbell: we're building this, which worked in previous releases but as broken in 0.8.x.

screen shot 2017-08-17 at 11 36 14 pm

PR #269, opened 2017-08-16
DEV: filters as dict

licode:
We can ignore https://github.com/NSLS-II/databroker/pull/256

### DIFF
--- a/databroker/_core.py
+++ b/databroker/_core.py
@@ -2029,7 +2029,7 @@ class Broker(BrokerES):
             raise ValueError("The 'plugins' argument is no longer supported. "
                              "Use an EventSource instead.")
         if filters is None:
-            filters = []
+            filters = {}
         if filters:
             warnings.warn("Future versions of the databroker will not accept "
                           "'filters' in __init__. Set them using the filters "
--- a/databroker/_core.py
+++ b/databroker/_core.py
@@ -1021,7 +1021,7 @@ class BrokerES(object):
         self.assets = assets
         # Once we drop Python 2, we can accept initial filter and aliases as
         # keyword-only args if we want to.
-        self.filters = []
+        self.filters = {}
         self.aliases = {}
         self.event_source_for_insert = self.event_sources[0]
         self.registry_for_insert = self.event_sources[0]
@@ -1108,7 +1108,7 @@ class BrokerES(object):
         Review current filters.
 
         >>> db.filters
-        [{'user': 'Dan'}, {'start_time': '2017-3'}]
+        {{'user': 'Dan'}, {'start_time': '2017-3'}}
 
         Clear filters.
 
@@ -1120,6 +1120,15 @@ class BrokerES(object):
 
         """
         self.filters.append(dict(**kwargs))
+        for k,v in kwargs.items():
+             if k in self.filters:
+                 raise ValueError('Filter name of {} already exists.\n \n
+                                  Please remove the existing filter first. \n
+                                  For example, if you want to remove filter name of start_time, just type \n
+                                  del db.filters["start_time"] '.format(k))
+             else:
+                 self.filters[k] = v
+
 
     def clear_filters(self, **kwargs):
         """
--- a/databroker/_core.py
+++ b/databroker/_core.py
@@ -1108,7 +1108,7 @@ class BrokerES(object):
         Review current filters.
 
         >>> db.filters
-        {{'user': 'Dan'}, {'start_time': '2017-3'}}
+        {'user': 'Dan', 'start_time': '2017-3'}
 
         Clear filters.
 
--- a/databroker/_core.py
+++ b/databroker/_core.py
@@ -1090,7 +1090,7 @@ class BrokerES(object):
         all future queries.
 
         ``db.add_filter(**kwargs)`` is just a convenient way to spell
-        ``db.filters.append(dict(**kwargs))``.
+        ``db.filters.update(**kwargs)``.
 
         Examples
         --------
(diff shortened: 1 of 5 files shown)


databroker/_core.py
```diff
@@ -1108,7 +1108,7 @@ def add_filter(self, **kwargs):
         Review current filters.
 
         >>> db.filters
-        [{'user': 'Dan'}, {'start_time': '2017-3'}]
+        {{'user': 'Dan'}, {'start_time': '2017-3'}}
```
tacaswell: This is not right.

tacaswell:
The docstring needs to be updated (http://nsls-ii.github.io/databroker/_as_gen/databroker.Broker.add_filter.html#databroker.Broker.add_filter is a rendered version of the current docstring) and an note needs to make it into the API changes.

danielballan:
When we talked about this on the way to the train, I thought we left it that we would just fix the list-based implementation for this release and revisit it in the next release. Did I have my wires crossed?
I still think that getting all the docs, tests, and examples fixed up by lunchtime feels like a stretch.

licode:
The good thing is that there is no API change. API stays the same.

licode:
Done

danielballan:
`Broker.filters` is part of the public API. For example, the docstring from `Broker.add_filter` says:

> ``db.add_filter(**kwargs)`` is just a convenient way to spell ``db.add_filter(**kwargs)`` is just a convenient way to spell ``db.filters.append(dict(**kwargs))``.

That usage would break. Which is OK, but it is an API change.

licode:
api note is added.

databroker/_core.py
```diff
@@ -1119,7 +1119,14 @@ def add_filter(self, **kwargs):
         :meth:`Broker.clear_filters`
 
         """
-        self.filters.append(dict(**kwargs))
+        for k,v in kwargs.items():
```
danielballan: Is this still useful? Should we just `self.filters.update(**kwargs)`?

licode: It is better to remind users that there are existing filter names. Just keep them aware of this.

doc/api-changes/db_filters.rst
```diff
@@ -0,0 +1,4 @@
+db.filters
```
danielballan: You have discovered an old `api-changes` folder that we no longer use. (I think Tom dumped some notes here a long time ago, but they aren't included in the published docs. We should remove them to avoid this confusion.) You should add a new bullet point to the bottom of the "User-Facing API Changes" section of `doc/source/whats_new/v0.9.0.txt`

databroker/headersource/shim.py
```diff
@@ -18,15 +18,15 @@ def __init__(self, mds):
 
     def __call__(self, text_search=None, filters=None, **kwargs):
         if filters is None:
-            filters = []
+            filters = {}
         else:
-            [_format_time(v, self.mds.config['timezone']) for v in filters]
+            [_format_time(v, self.mds.config['timezone']) for v in [filters]]
```
danielballan: No need for a list comprehension here anymore, right?

databroker/headersource/shim.py
```diff
@@ -18,15 +18,15 @@ def __init__(self, mds):
...
         if text_search is not None:
-            query = {'$and': [{'$text': {'$search': text_search}}] + filters}
+            query = {'$and': [{'$text': {'$search': text_search}}] + [filters]}
```
danielballan: ditto here and the next one below

databroker/_core.py
danielballan: I'm not sure. This feels like second-guessing how the language works. You are allowed to do `d = {'a': 1}; d['a'] = 2` without doing `del d['a']` in the middle. I'm not convinced filters should work differently.

licode: agree. self.filters.update(**kwargs) looks like a much simpler option.

doc/source/whats_new/v0.9.0.txt
```diff
@@ -63,6 +63,8 @@ The following changes may break old user code.
      event['descriptor'] == descriptor['uid']
      descriptor['run_start'] == run_start['uid']
      run_stop['run_start'] == run_start['uid']
+* The type of db.filters changed from list to dict.
```
danielballan: This is going to generate sphinx errors. You need a blank line between the end of the ``.. code-block`` above and this bullet point.

databroker/headersource/shim.py
```diff
@@ -18,15 +18,15 @@ def __init__(self, mds):
...
     def __call__(self, text_search=None, filters=None, **kwargs):
         if filters is None:
-            filters = []
+            filters = {}
         else:
-            [_format_time(v, self.mds.config['timezone']) for v in filters]
+            _format_time(filters, self.mds.config['timezone'])
         if text_search is not None:
-            query = {'$and': [{'$text': {'$search': text_search}}] + filters}
+            query = {'$and': [{'$text': {'$search': text_search}}] + [filters]}
```
danielballan: This works, but it's a bit more obfuscated than it needs to be since `filters` is a now dict. This is like saying `[1] + [2]` instead of `[1, 2]`. Ditto below with `query = ...`

danielballan:
Looks good. This should be merged if Travis passes.

Thanks for your patience with many rounds of review, @licode. This is definitely smarter than my original implementation with a list. I'm glad we got this in.

licode:
Thanks. In order to avoid bugs when it is deployed to beamline, we got to review each code carefully.

danielballan:
The Travis timezone bug strikes again! Otherwise this passes.

PR #272, opened 2017-08-17
MNT: move pymongo import to be local

tacaswell:
This makes pymongo actually an optional dependency


tacaswell:
failure is a server start

PR #280, opened 2017-08-18
ENH: Support correct location for Windows config

### LINKED ISSUE #279: CI: Windows CI
Investigate and possibly implement a Windoze CI config

stuartcampbell:
Instead of just using .config in the users folder we now use the correct Windows location
i.e. `%APPDATA%`

I haven't full tested this on Windows yet - having some issues getting all the dependencies working under Windows, namely `ujson` at the moment (I would need to install Visual Studio!).  I think for this release a code review and working for unix is good enough ?

closes #279

### DIFF
--- a/databroker/_core.py
+++ b/databroker/_core.py
@@ -840,18 +840,22 @@ class Results(object):
                         yield header
                         break
 
-# Search order is:
-# ~/.config/databroker
-# <sys.executable directory>/../etc/databroker
-# /etc/databroker
+# Search order is (for unix):
+#   ~/.config/databroker
+#   <sys.executable directory>/../etc/databroker
+#   /etc/databroker
+# And for Windows we only look in:
+#   %APPDATA%/databroker
 
-
-_user_conf = os.path.join(os.path.expanduser('~'), '.config', 'databroker')
-_local_etc = os.path.join(os.path.dirname(os.path.dirname(sys.executable)),
-                          'etc', 'databroker')
-_system_etc = os.path.join('/', 'etc', 'databroker')
-CONFIG_SEARCH_PATH = (_user_conf, _local_etc, _system_etc)
-
+if os.name == 'nt':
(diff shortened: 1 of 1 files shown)


CJ-Wright:
It's ujson not on conda-forge? You might consider putting a package request into staged-recipes.

databroker/_core.py
```diff
@@ -840,18 +840,22 @@ def __iter__(self):
...
+# Search order is (for unix):
+#   ~/.config/databroker
+#   <sys.executable directory>/../etc/databroker
+#   /etc/databroker
+# And for Windows we only look in:
+#   %APPDATA%/databroker
+
+if os.name == 'nt':
+    _user_conf = os.path.join(os.environ['APPDATA'], 'databroker')
+    CONFIG_SEARCH_PATH = (_user_conf)
```
danielballan: I think this needs to be `(_user_conf,)`.

danielballan:
> I think for this release a code review and working for unix is good enough?

Yeah, that's good enough for me. See comment for one tiny revision.

The `usjon` dependency is only needed for the mongo server/client deployment, and that should be never be done on a Windows machine. @CJ-Wright It looks like anaconda packages it (https://anaconda.org/anaconda/ujson) but in any case we'll probably need to find a better serialization method since we know it encodes floating-point numbers lossily.

danielballan:
Travis failures are a _ton_ of server startup failures, more than just a couple. Rerunning to see what happens.

stuartcampbell:
Dunno what happened with `ujson` - tried again and it worked! Glitchy computers! 😄

databroker/_core.py
stuartcampbell: Of course, silly fingers, and this typing...

danielballan:
Travis failures are the phantom German timezone bug.

databroker/_core.py
danielballan: If only it could be done without commas and parentheses!

PR #284, opened 2017-08-18
BUG: keep a copy of filters

licode:
due to the issue of https://github.com/NSLS-II/databroker/issues/283


licode:
@danielballan

danielballan:
:+1:

PR #289, opened 2017-08-22
FIX: Change db.fs -> db.reg

stuartcampbell:


danielballan:
Cherry-picked and merged locally.

PR #291, opened 2017-08-23
typo

CJ-Wright:


databroker/tests/test_broker.py
```diff
@@ -64,7 +64,7 @@ def test_uid_list_multiple_headers(db, RE):
     RE.subscribe(db.insert)
     uids = RE(pchain(count([det]), count([det])))
     headers = db[uids]
-    assert uids == [h['start']['uid'] for h in headers]
+    assert uids == tuple([h['start']['uid'] for h in headers])
```
danielballan, line 67: I think this change would make the reader wonder why `tuple()` is needed to make the test pass. In fact, of course, it is not needed: lists and tuples may be compared.

CJ-Wright: Please tell me more about this magical comparison.
https://travis-ci.org/NSLS-II/databroker/jobs/269221759#L1455

danielballan: :sheep:. Indeed:

```python
>>> >>> (1,2) == [1,2]
False
```

danielballan:
One Travis build passed; the others are server startup and random network failure. Merging.

PR #293, opened 2017-08-28
FIX: if documents are filled do nothing

CJ-Wright:


CJ-Wright:
Failing on stuff fixed in https://github.com/NSLS-II/bluesky/pull/788

CJ-Wright:
@danielballan @tacaswell @arkilic ready for review!

databroker/tests/test_broker.py
```diff
@@ -64,7 +64,7 @@ def test_uid_list_multiple_headers(db, RE):
     RE.subscribe(db.insert)
     uids = RE(pchain(count([det]), count([det])))
     headers = db[uids]
-    assert uids == [h['start']['uid'] for h in headers]
+    assert uids == tuple([h['start']['uid'] for h in headers])
```
danielballan: See my comment on #291.

databroker/tests/test_broker.py
```diff
@@ -592,7 +592,11 @@ def __call__(*args, **kwrags):
     assert not ev2['filled']['image']
     ev2_filled = db.fill_event(ev2, inplace=False)
     assert ev2_filled['filled']['image'] == ev2['data']['image']
-
+    ev_ret2 = db.fill_events([ev], h.descriptors, inplace=True)
+    ev_ret3 =  db.fill_events(ev_ret2, h.descriptors, inplace=True)
+    ev = next(ev_ret3)
+    assert ev['filled']['image'] == datum
+    assert ev['filled']['image']
```
danielballan: What is the purpose of this line?

danielballan:
> Failing on stuff fixed in NSLS-II/bluesky#788

It looks like 2/3 of the builds passed. I don't think the one failure had anything to do with  NSLS-II/bluesky#788. Databroker tests run against bluesky master, and NSLS-II/bluesky#788 was merged last week.

databroker/tests/test_broker.py
CJ-Wright: I guess that isn't needed now that we are not putting `True` in things.

danielballan: I'm just confused about 598 and 599. If 598 is true and `datum` is not false-y, then 599 we be true also.

PR #294, opened 2017-08-29
REF: Simplify sqlite test fixture using temp_config.

danielballan:


databroker/tests/utils.py
```diff
@@ -6,43 +6,24 @@
...
 
-    tf = tempfile.NamedTemporaryFile()
-    fs = Registry({'dbpath': tf.name})
+    config = temp_config()
+    tempdir = config['metadatastore']['config']['directory']
 
-    def delete_fs():
-        os.remove(tf.name)
+    def cleanup():
+        os.remove(tempdir)
```
tacaswell: It looks like this is trying to remove a (non-empty?) directory and failing.

danielballan: This should be using rmtree anyway. Force-pushed a fix.

PR #295, opened 2017-08-30
FIX: Respect fill kwarg in databroker (a regression never released)

danielballan:


tacaswell:
Thank you for cleaning up my mess :disappointed_relieved:

Issue #299, opened 2017-09-01
slower queries?


jrmlhermitte:
This might be just me, but I figure the more documentation the better so I'll write it.

A query like this:
```python
hdrs = cmsdb(start_time="2017-06-01", stop_time="2017-07-01")
next(iter(hdrs))
```
returns a result near instantaneously.
However, a restreaming as such:
```python
hdrs = cmsdb(start_time="2017-06-01", stop_time="2017-07-01")
stream = cmsdb.restream(hdrs, fill=False)
nds = next(iter(stream))
```
hangs for a long time (5 min or more).

Any idea why this may be happening?

I'm using databroker version 0.9.0+16.gd78bbfa
The broker object is constructed as follows:
```python
from databroker.broker import Broker
from databroker.headersource.mongo import MDSRO
from databroker.assets.mongo import RegistryRO

    mds_conf = {
                'host': host,
                'port': port,
                'database': mdsname,
                'timezone': 'US/Eastern',
                 }

    reg_conf = {
            'host': host,
            'port': port,
            'database': fsname
    }

    mds = MDSRO(mds_conf)
    reg = RegistryRO(reg_conf)
    db = Broker(mds, reg)
```

danielballan:
It looks like `check_fields_exist` is applied to all the headers in a separate loop, breaking the laziness of restream.

danielballan:
Fixed in #300 and merged in time for v0.9.1.

PR #300, opened 2017-09-01
PERF: Do not call check_fields_exist inside get_documents.

danielballan:
The purpose of this line was to alert the user if none of the headers
had a field that the user requested. This of course requires greedily
checking all of the headers, which breaks the laziness of Results.
Based on benchmarking by @ordirules, it can take > 5 minutes to get
the first result out of month's worth of data. Removing this line
reduces that time to < 1 second.

### DIFF
--- a/databroker/_core.py
+++ b/databroker/_core.py
@@ -1528,7 +1528,6 @@ class BrokerES(object):
         else:
             headers = [headers]
 
-        check_fields_exist(fields if fields else [], headers)
         # dirty hack!
         with self.reg.handler_context(handler_registry):
             for h in headers:
--- a/databroker/tests/test_broker.py
+++ b/databroker/tests/test_broker.py
@@ -162,9 +162,6 @@ def test_get_events_filtering_field(db, RE):
     assert len(list(db.get_events(h, fields=['det']))) == 7
     assert len(list(h.documents(fields=['det']))) == 7 + 3
 
-    with pytest.raises(ValueError):
-        list(db.get_events(h, fields=['not_a_field']))
-
     uids = RE(pchain(count([det1], num=7), count([det2], num=3)))
     headers = db[uids]
 


jrmlhermitte:
From Issue #299 
Thanks @danielballan ! :-)

jrmlhermitte:
Here is comparison of 1 month's worth of data at cms:
```python
# old method
# commit d78bbfab6e21b6b7aa0eff61fec9ad24b7faac1e (currently 'master')
In [1]: import time

In [2]: hdrs = cmsdb(start_time="2017-06-02", stop_time="2017-07-01")

In [3]: stream = cmsdb.restream(hdrs)

In [4]: t1=time.time();nds = next(iter(stream));t2=time.time()
print(t2-t1)

In [5]: print(t2-t1)
453.14725065231323


# new method:
# commit : 81b4c3cccf5daa8edb33d1e48c93cd957c9a795c (currently Dan's branch)
In [1]: import time

In [2]: hdrs = cmsdb(start_time="2017-06-02", stop_time="2017-07-01")

In [3]: stream = cmsdb.restream(hdrs)

In [4]: t1=time.time();nds = next(iter(stream));t2=time.time()

In [5]: print(t2-t1)
0.2851831912994385
```
(didn't use `timeit`because results get cached after first run it seems). So that's 7min 33s or so versus .29 s...

EDIT : Timed 1 month worth of data instead for effect.

jrmlhermitte:
EDIT : deleted this comment as it is out of scope (has to do with a bad document). Will post as a new issue soon

jrmlhermitte:
The other issue I mentioned is resolved by #303 (still to be discussed).

I had noticed some other odd behaviour, but now I'm convinced this is fine. Basically, my search had a `start` document that had no `stop`. Before this PR, my `restream` would raise a `TypeError` upon the `restream` call, since the operation wasn't lazy. After this PR, `restream` only raises an error upon iterating on the bad document.

In other words, this looks good to me and works testing with my data. I don't think a test can be written for this one?

PR #301, opened 2017-09-01
DOC: Change DataBroker to db in docstrings for self-consistency.

danielballan:


danielballan:
Self-merging trivial docs fix

PR #303, opened 2017-09-01
API: return {} instead of None if missing stop document

tacaswell:
This changes the HeaderSource behavior to match the default in the
Header constructor.

This needs tests + discussion

atn @ordirules

### DIFF
--- a/databroker/headersource/shim.py
+++ b/databroker/headersource/shim.py
@@ -62,4 +62,4 @@ def safe_get_stop(hs, s):
     try:
         return hs.stop_by_start(s)
     except hs.NoRunStop:
-        return None
+        return {}


jrmlhermitte:
This resolves the `TypeError` issue. I tested on our data and it works now.
As a general question, is there a plan to eventually allow different `start`s and `stop`s to be interleaved? (So multiple starts may be received and processed before receiving any stops). If yes, then defining whether there is a missing stop may tricky?

For tests, I'm not very familiar, is something like this sufficient maybe (tests/test_broker.py)?
```python
def test_no_stop(db):
    stream = [
        ('start', {'uid' : 'fdasfsa', 'time' : 1000}),
        ('start', {'uid' : 'fdasfsa1', 'time' : 1100}),
    ]
    for name, doc in stream:
        db.insert(name, doc)
    stream = list(iter(db.restream(db(start_time=0, stop_time=1200)))
    assert stream[0][0] == 'start'
    assert stream[1][0] == 'stop'
    assert stream[1][1] == {}
```

tacaswell:
Currently the run engine can not have more than one open run at a time so at generation time you are guaranteed never see interleaved documents if you are just watching the event stream.  DataBroker is similar (because that is the easiest way to implement it or you have a `Header` object which only has documents associated with the current Start.

There are mid-to-long term plans to allow multiple runs to be open in the run engine so the live stream will have interleaved runs.  The callbacks consuming the documents will have to be able to cope with that extra complexity.

jrmlhermitte:
thanks for info, this is something @CJ-Wright may also be interested to know about

CJ-Wright:
I think that some of this discussion is over in https://github.com/NSLS-II/bluesky/issues/831

jrmlhermitte:
note : still needs a test @danielballan

PR #305, opened 2017-09-11
FIX/TST/DOC: Handler registration via config file

danielballan:


danielballan:
This feature was added just before v0.9.0 but has not yet been used in the wild. It was evidently broken, so I added a test. We should tag a v0.9.2 when this is merged.

tacaswell:
@stuwilkins Did you see this go by?   Moving away from automatic registration and using this is probably the best path forward.

stuwilkins:
Yes @tacaswell i did!

This is a nice way of going - i think - becuase it allows configuration of the correct handler on configuration of `Databroker`. I am using this to use the DASK handler at CSX.

However, currently Databroker has an `autoregister` function to register handlers _out of the box_.
When I started putting the DASK and SWMR handlers in we now have the issue that we have multiple handlers with the same SPECS which the autoregister errors on. 

So, given that we now have multiple handlers for a SPEC how do we proceed? 
We could:

1. Remove the autoregister and rely on setting up handlers on a per config file basis.
2. Use the solution in PR #308 which has a class member `autoregister` which if `autoregister == True` causes the handler to be registered on databroker config. 
3. Hardcode the registration 

While I like 1, I fear that it is going to break quite a lot. Which way do we want to proceed? I am keen to get this cleaned up and moving ... Thanks!

tacaswell:
Option 2 on the way to option 1

stuwilkins:
Ok thanks @tacaswell , any chance someone could look at PR #308 then, this should meet Option 2

danielballan:
I just submitted a review on #308.

PR #306, opened 2017-09-11
Fix fill

CJ-Wright:


CJ-Wright:
Error is dues to timeout ready for review!

danielballan:
You mentioned in person that you didn't have time to write a test. A little more information about the bug that this is fixing would help me write one. For example, does it raise a KeyError?

CJ-Wright:
Yes it raises a key error

danielballan:
OK, I think I got it. My new test...

```
$ python run_tests.py databroker/tests/test_broker.py::test_fill_and_multiple_streams[sqlite]
```

raises like this when run against the master branch:

```python
        while True:
            if not inplace:
                ev = _sanitize(ev)
                ev = copy.deepcopy(ev)
                ev = self.prepare_hook('event', ev)
            data = ev['data']
            filled = ev['filled']
            desc_id = ev['descriptor']
            for k, v in six.iteritems(fill_map):
                for dk in v:
>                   if not filled[dk]:
E                   KeyError: 'image1'

databroker/_core.py:1998: KeyError
```

and your commit resolves that failure. I squashed and added a commit to eliminate the outer loop, which is unneeded. It would still be good to check that this fixes the particular case you found at XPD.

PR #308, opened 2017-09-14
Refactor HDF5 handlers and add DASK handler

stuwilkins:
This is an attempt to refactor the HDF5 handlers and add handlers for returning DASK objects

### DIFF
--- a/databroker/assets/handlers.py
+++ b/databroker/assets/handlers.py
@@ -248,6 +248,7 @@ class AreaDetectorHDF5SWMRHandler(AreaDetectorHDF5Handler):
         number of frames to return as one datum, default 1
     """
     specs = {'AD_HDF5'} | AreaDetectorHDF5Handler.specs
+    autoregister = False
 
     def __init__(self, filename, frame_per_point=1):
         super(AreaDetectorHDF5SWMRHandler, self).__init__(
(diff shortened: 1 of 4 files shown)


stuwilkins:
@danielballan @tacaswell @stuartcampbell 

This is now ready for review. I have tried to restructure the HDF5 stuff. 

The question is what to do with autoregister, one possibility is what is here. I was also considering the possibility of having a member of the class be something like `autoregister=True` to configure what comes out of the box. Or no autoregister and configure this in the yaml config files for all beamlines. Thoughts? I am happy to do the work!

databroker/assets/handlers.py
```diff
@@ -6,54 +6,13 @@
...
+import dask.array as da
+
 from .handlers_base import HandlerBase
 from .readers.spe import PrincetonSPEFile
-from pims import FramesSequence, Frame
 
 logger = logging.getLogger(__name__)
 
-
-# The ImageCube class is used for a per event representation of
```
tacaswell: Why did these get deleted?

stuwilkins:
@tacaswell Sorry, over aggressive refactoring, I was intending on dropping PIMS but that is probably not a good idea ... I will add an HDF5 Handler to just return the h5py object

stuwilkins:
@danielballan done... travis fails though (not on the handlers)

danielballan:
The failure is due to pims. It has been fixed and released upstream, but the release has not yet made its way through conda-forge.

PR #310, opened 2017-09-20
FIX: DeprecatedDoct improperly warned on _repr_html_().

### LINKED ISSUE #309: Allow internal usage of dot
I noticed this a couple days ago but neglected to create an issue. Reported by @stuwilkins.

pastedgraphic-1 1

danielballan:
Closes #309


danielballan:
There seems to be an import failure on 2.7 coming in through pims. It's totally unrelated to this PR but should be sorted out...

tacaswell:
There is already a pims bug report for this https://github.com/soft-matter/pims/issues/276

PR #314, opened 2017-11-02
TST: update for ophyd/bluesky version bump

tacaswell:
Not done, but want travis to spin on what is there.


test-requirements.txt
```diff
@@ -1,2 +1,3 @@
 coverage
 pytest
+matplotlib
```
danielballan: I'll push a commit alphabetizing the requirements.

PR #315, opened 2017-11-02
added swimmer parameter doc

jrmlhermitte:
very small detail but figured i'd as I go

### DIFF
--- a/databroker/assets/handlers.py
+++ b/databroker/assets/handlers.py
@@ -1,7 +1,5 @@
 from __future__ import (absolute_import, division, print_function,
                         unicode_literals)
-import six
-
 import logging
 import numpy as np
 import os.path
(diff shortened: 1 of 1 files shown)


databroker/assets/handlers.py
```diff
@@ -142,6 +142,8 @@ class HDF5DatasetSliceHandler(HandlerBase):
         key of the single HDF5 Dataset used by this Handler
     frame_per_point : integer, optional
         number of frames to return as one datum, default 1
+     swmr : bool, optional
```
danielballan: Is there a stray space before `swmr` here? That will probably break sphinx formatting.

databroker/assets/handlers.py
```diff
@@ -142,6 +142,8 @@ class HDF5DatasetSliceHandler(HandlerBase):
...
+        Open the hdf5 file in SWMR read mode. Only used when mode = 'r'.
```
danielballan: Could you note, `False by default.`?

danielballan:
👍 👍 👍 👍 👍 👍  for correcting "small details" in our docs! Warms my heart.

jrmlhermitte:
cool thanks! :)

PR #321, opened 2017-12-07
DOC: added all projects toctree to databroker

mrakitin:
## Description

Add all projects toctree to the databroker documentation.

## Motivation and Context

Motivated by https://github.com/NSLS-II/docs/issues/73.

## How Has This Been Tested?

Tested locally by building the documentation.

### DIFF
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -23,3 +23,18 @@ Index
     archiver
     whats_new
     DBEP/index
+
+.. toctree::
+   :hidden:
+   :caption: Bluesky Project
+
+   Bluesky Project <https://blueskyproject.io>
+
+.. toctree::
+   :hidden:
+   :caption: GitHub Links
+
+   Bluesky Project Repositories <https://github.com/bluesky/>
+   NSLS-II Repositories <https://github.com/NSLS-II/>
+   Wish List <https://github.com/NSLS-II/wishlist/issues>
+   Bug Reports <https://github.com/NSLS-II/Bug-Reports/issues>
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -29,11 +29,10 @@ Index
    :caption: Bluesky Project
 
    Bluesky Project <https://blueskyproject.io>
+   Bluesky Project Repositories <https://github.com/bluesky>
 
 .. toctree::
    :hidden:
-   :caption: GitHub Links
+   :caption: Getting Help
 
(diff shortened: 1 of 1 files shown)


jrmlhermitte:
this looks good to me (and I'm trusting that it builds when you say it does)
@danielballan is this good to go?

danielballan:
Let's get Travis's BUILD_DOCS build to pass before we merge. I just kicked it to re-run. Merge at will if/when it passes.

mrakitin:
Probably it needs rebasing and manual resolution of the conflict.

danielballan:
I think we should revisit whether and how to do this. The number of packages is growing large. Instead of linking to them all, maybe just link to blueskyproject.io? What do you think, @mrakitin?

mrakitin:
I agree with that approach.

mrakitin:
I rebased on master and changed all the project documentation links to a single URL. Does it make sense, @danielballan?

danielballan:
Thanks! Do you think we should remove the NSLS-II link as well, just to give the right impression? And maybe replace BugReports with the gitter, which is probably a better first place to go for help?

mrakitin:
It's available via 560185a.

doc/source/index.rst
```diff
@@ -23,3 +23,16 @@ Index
...
+   :caption: Bluesky Project
+
+   Bluesky Project <https://blueskyproject.io>
+   Bluesky Project Repositories <https://github.com/bluesky>
+
+.. toctree::
+   :hidden:
+   :caption: Getting Help
+
+   Gitter <https://gitter.im/NSLS-II/DAMA>
```
danielballan, line 38: Actually --- In the spirit of scrubbing facility-specific branding out, should we move to a bluesky gitter? The way projects do this (e.g. jupyterlab moved from jupyter/jupyterlab -> jupyterlab/jupyterlab) is to post a big message in the gitter with a link to the new one. There is no way to set up an actual redirect.

danielballan: If this were _mainly_ an NSLS-II support channel I would say we should leave it as is, but it seems that a fair fraction of the questions we get are from other facilities that are evaluating bluesky.

stuartcampbell: I agree

danielballan:
Good enough for me. We might want to resolve the gitter question before we update _all_ the repos though, to save effort.

PR #322, opened 2017-12-11
API: asset registry refactor

tacaswell:
This writes up discussions with @jrmlhermitte @danielballan and @licode from the last few work days.


doc/source/DBEP/DBEP02.rst
```diff
@@ -0,0 +1,167 @@
...
+       more than once.  Do we want to allow the RE to emit them more
+       than once? **[QUESTION]**
+
+ 2. in ``save`` before the *Event* document is emitted the *RunEngine*
+    will acquire and emit any *AssetRegistry* Documents.
+
+    a. in ``save`` the *RunEngine* knows what objects in the bundle, call
+       ``get_asset_document_gen`` **[RENAME]** method ::
+
+            def gadg(self) -> Iterator[Tuple[str, Dict[str, Any]]]:
```
danielballan: As names and abbreviations go, you have surpassed yourself. :- )

doc/source/DBEP/DBEP02.rst
```diff
@@ -0,0 +1,167 @@
...
+
+
+Detailed description
+====================
+
+An odd asymmetry in how databroker works is that the documents for
+*HeaderSource* and *EventSource* are emitted by the *RunEngine* and
+can be subscribed to by one or more consumers.  Each consumer is
+notionally independent, each receive all of the documents, and do not
+need to coordinate in any way (or even be aware of each others
```
danielballan: "each others" -> "each other's" or strictly "one another's" for N > 3

doc/source/DBEP/DBEP02.rst
```diff
@@ -0,0 +1,167 @@
...
+
+Currently, *ophyd* is responsible for collecting all of the values for
+the *Resource* and *Datum* documents except for the uids.  The uids
+are generated by calls to ``reg.register_*`` and the datum uids are
+subsequently returned to the *RunEngine* via ``obj.read``.  The
+proposed change is:
+
+ 1. *ophyd* objects would be responsible for generating the full *Resource*
+    and *Datum* documents and providing them to the *RunEngine* to be
+    emitted (likely should provide some helpers).
```
danielballan: Suggestion: cut the parenthetical phrase or elaborate or add some pronouns. I think you mean that the ophyd library could probably some helper functions which device implementations can be use for generating these documents.

doc/source/DBEP/DBEP02.rst
```diff
@@ -0,0 +1,167 @@
...
+
+    a. Similar to the current documents, a *Resource* must be emitted
```
danielballan: I know what this means but I think I wouldn't understand it if I didn't already. "... a *Resource* must be emitted before any *Datum* that refers to it. A *Datum* can only refer to a *Resource* that has been emitted since the last *Start* document.

doc/source/DBEP/DBEP02.rst
danielballan: Brainstorming:

* `assets()`
* `collect_assets()`
* `collect_asset_docs()`

doc/source/DBEP/DBEP02.rst
```diff
@@ -0,0 +1,167 @@
...
+       before any *Datum* that refers to it per *Start* / *Stop* pair
+    b. an identical (including uid) *Resource* maybe emitted more than
+       once, the consumers will need to handle this.
+    c. an identical *Datum* document may be returned the *RunEngine*
```
danielballan: My answer to the question here: We could attempt to de-duplicate within a given plan, but we can't de-duplicate globally without caching an unbounded set of UUIDs. Since consumers are potentially long-running and don't know the scope of a plan (they only see runs) I don't it's useful to attempt de-duplication at all.

doc/source/DBEP/DBEP02.rst
```diff
@@ -0,0 +1,167 @@
...
+-----
+
+ 1. implement new document generation methods on all devices that have
+    external data.
+
+
+bluesky
+-------
+
+ 1. implement above logic in ``RunEngine.save``
```
danielballan: It's `RunEngine._save`.

danielballan:
Another thought that I want to record here: it is currently difficult to make a callback that uses external data, such as a live image viewer. There are two forms on latency that have to be dealt with:

1. Waiting for the asset registry to 'know' about a new datum_id
2. Waiting for the file to be available on disk

This does nothing to solve (2) but it obviates (1) entirely. A callback will directly receive all the information it needs, in order, to display externally-stored data to the user as soon as that externally-stored data is accessible on disk (which we cannot ever control).

doc/source/DBEP/DBEP02.rst
```diff
@@ -0,0 +1,167 @@
...
+notionally independent, each receive all of the documents, and do not
+need to coordinate in any way (or even be aware of each others
+existence). In contrast, the *Resource* and *Datum* documents are
+inserted directly into an *AssetRegistry* by the *ophyd* objects.
+This breaks the separation we have between the data collection process
+/ hardware, the generation of the documents, and the consumers of
+those documents and leads to several unfortunate situations:
+
+ - *ohpyd* objects hold an instance of an *AssetRegisty*
+ - we need to keep track of _which_ *AssertRegistry* things were
```
jrmlhermitte: I think it's ``**which**`` for emphasis?

doc/source/DBEP/DBEP02.rst
jrmlhermitte: I like `collect_asset_docs`.

doc/source/DBEP/DBEP02.rst
jrmlhermitte: I would also be tempted to leave them as is. Duplicate `datums` sounds like a bug to me and something that shouldn't happen. So if it did (and it will) having traces of it at the document level may be easier to debug than searching around at the ophyd level.
(Correct me if I misunderstood. I understood this as "should we curate/clean documents as they the `RE` receives them?" I would say no)

doc/source/DBEP/DBEP02.rst
```diff
@@ -0,0 +1,167 @@
...
+
+ 3. consumers will now have access to all relevant documents and can
+    do what ever they want with them (insert into an asset registry,
+    live processing / display, copy files else where)
+
+event-model
+-----------
+
+ 1. add schema for *Resource* and *Datum*
+ 2. assert that datum_id must be of the form ``{resource_id}/{N}``.
```
jrmlhermitte, line 121: I don't know if this was the previous way of writing it but this sounds very sensible.

doc/source/DBEP/DBEP02.rst
```diff
@@ -0,0 +1,167 @@
...
+    This is required to support columnar stores where the *Datum*
+    documents are group by *Resource* id.
+
+
+databroker
+----------
+
+ 1. teach ``insert`` how to deal with the additional documents.
```
jrmlhermitte, line 129: We may want to rename the new `insert` to something else. Seems like the new `insert` will be doing a little more (correct me if wrong).

doc/source/DBEP/DBEP02.rst
tacaswell: The issue I am thinking of here is if we have a device that returns from read things that look like

```python
{'image': UID1,
 'background': UID2,
 'flatfield': UID3}
```

where the 'background' and 'flaitfield' may stay constant of many calls to `read`.  Updated this with some requirements on the latch logic around this.

doc/source/DBEP/DBEP02.rst
tacaswell: previously there was no constraints so it was just uuid4 values.

doc/source/DBEP/DBEP02.rst
tacaswell: I think that the name `insert` is still ok here, it is just a broader class of documents being inserted into the databases.  I tend to think of the innards of databroker as "some way of persisting documents to get them back later".  This glosses over all of the search on the headers aspect of this, but....

doc/source/DBEP/DBEP02.rst
```diff
@@ -0,0 +1,186 @@
...
+    a. in ``save`` the *RunEngine* knows what objects in the bundle, call
+       ``collect_asset_docs`` method ::
+
+            def collect_asset_docs(self) -> Iterator[Tuple[str, Dict[str, Any]]]:
+                ...
+
+       which will yield the ``(name, doc)`` pairs for anything that
+       was just read.
+
+    b. these documents will be emitted **before** the *Event*
```
CJ-Wright, line 111: Does this mean that the raw stream will be:
1. Start
1. Descriptor
1. Resource
1. Datum
1. Event
1. Datum
1. Event
1. Stop

Does that mean that there will be two sources of the "data", one in the Datums and one in the Events?

doc/source/DBEP/DBEP03.rst
```diff
@@ -0,0 +1,59 @@
...
+====================
+
+For both *EventTable* and *DatumTable* the schema is identical to the single
+version except in place of scaler values there is an array of scaler values.
+These can be represented as:
+
+- nested `dict` with `numpy.array` or `list` at the bottom
+- a `dict` of `pandas.DataFrame` with simple column index and a shared row index
+- a `pandas.DataFrame` with a MultiIndex
+
```
ronpandolfi: On the access side, getting back a single array (or array-like object) containing all event data for a particular field would be an amazing feature. It should be noted though that for long scans I wouldn't want to transmit all frames at once when accessing that table. While I would want an object describing the shape, min, max, and size of the array to represent the contents, I might only access a subset of that array (typically slicing on the event/frame axis). Can I also have the option to get back an array-like object with 'lazy' access to each frame?

doc/source/DBEP/DBEP02.rst
tacaswell: Yes, that is the order.



> Does that mean that there will be two sources of the "data", one in the Datums and one in the Events?

"data" is too ambiguous in this sentence :)  

Currently, there is the values in the event and then a magical asset registry (nee filestore) which lets you trade tokens for (possibly large) arrays stored outside of the database.   This is just surfacing the information that is in the asset registry in a way that all clients can see (and plugs the _huge_ hole in our conceptual model that having ophyd objects know about an asset registry).  This should make it strait forward to write a consumer that ex-filtrates everything (documents and files) to an external system (for example Simon's AWS account :smiling_imp: ). 

From the point of view of 'source' in the descriptor there will still only be one value (as it is now).

doc/source/DBEP/DBEP03.rst
tacaswell: Yes to all of those :smile: 

 - we should be able ask for the tables as chunks (ex 100 rows at a time).  This fits very well with what apache arrow supports.
 - we should be able to ask for any sub-set of the columns.  we already have that on some of the DB api, but needs to be everywhere.  There is also a lot of space for optimization in this area.
 - the shape should be known from the descriptor (the type is _not_ well known past 'number', that is something else that we need to look into improving), but I don't think we compute any statistics on the raw-data (but that maybe something we should be grabbing at collection time).
 - the `data` method on `Header` will give you a generator of a specific field (https://nsls-ii.github.io/databroker/_as_gen/databroker.Header.data.html#databroker.Header.data)
 - There is already a way to get a `pims` object which will do lazy asset registry (nee filestore) retrievals and several proto-type dask-based

doc/source/DBEP/DBEP02.rst
CJ-Wright: Sorry in this example "data" are the images coming from an area detector.
Maybe a better question is will I still be able to/have to trade tokens for arrays?
For example does `LiveImage` still look for events or datums?
Do all callbacks which want large arrays need to cache datums and then wait for their event to come down?

doc/source/DBEP/DBEP03.rst
danielballan: The dask prototype is [here](https://github.com/NSLS-II-CHX/eiger-io/blob/master/eiger_io/fs_handler_dask.py) but it's a convoluted example for a couple reasons (the Eiger data format is unusual, and we're working around some historical baggage here...). In principle, dask should win us the 'lazy along any axis' feature, modulo what HDF5 can support.

doc/source/DBEP/DBEP02.rst
danielballan: `LiveImage` now has two options.

1. It can continue to work the way it does today, ignoring the new documents (datum and resource) that flow through and trading tokens for arrays via a `Broker`. A downside of this approach is that sometimes the RunEngine hands `LiveImage` an Event with some datum_id _before_ the `Registry` knows about it. (MongoDB takes some time to reflect new information in queries.) To be robust, `LiveImage` has to be prepared to poll `Broker` about a `datum_id` more than once, waiting for things to sync.
2. Alternatively, `LiveImage` can cache the datum and resource documents and use a Handler directly to load the data. In this scenario, `LiveImage` doesn't need a `Broker` (there doesn't even technically need to _be_ one!), but it does now need a Handler.

tacaswell: This discussion points on the handler registry, management, and caching should probably be re-factored to make option 2 easier.

danielballan:
Implementation:

- https://github.com/NSLS-II/event-model/pull/14/
- https://github.com/NSLS-II/bluesky/pull/999
- https://github.com/NSLS-II/databroker/pull/350
- https://github.com/NSLS-II/ophyd/pull/519

Integration test:

```python
from databroker import Broker
db = Broker.named('temp')
from bluesky.plans import count
from bluesky import RunEngine
from ophyd.sim import img, NumpySeqHandler
RE = RunEngine({})
db.reg.register_handler('NPY_SEQ', NumpySeqHandler)
RE.subscribe(lambda name, doc: print(name))
RE.subscribe(db.insert)

# Test 1: Ophyd object has no Registry; all Resource/Datum insert happens via RE.
uid, = RE(count([img]))
header = db[uid]
arr, = header.data('img')
print(arr)

# Test 2: Ophyd object has a Registry, as in old configurations
# Resource/Datum insert happens from ophyd AND then from RE.
# The RE's insert is ignored and a warning is shown.
img.reg = db.reg
uid, = RE(count([img]))
header = db[uid]
arr, = header.data('img')
print(arr)
```

Output:

```python
$ ipython -i test.py
Python 3.6.4 |Anaconda, Inc.| (default, Jan 16 2018, 12:04:33)
Type 'copyright', 'credits' or 'license' for more information
IPython 6.2.1 -- An enhanced Interactive Python. Type '?' for help.
start
descriptor
resource
datum
event
stop
[[ 1.  1.  1.  1.  1.  1.  1.  1.  1.  1.]
 [ 1.  1.  1.  1.  1.  1.  1.  1.  1.  1.]
 [ 1.  1.  1.  1.  1.  1.  1.  1.  1.  1.]
 [ 1.  1.  1.  1.  1.  1.  1.  1.  1.  1.]
 [ 1.  1.  1.  1.  1.  1.  1.  1.  1.  1.]
 [ 1.  1.  1.  1.  1.  1.  1.  1.  1.  1.]
 [ 1.  1.  1.  1.  1.  1.  1.  1.  1.  1.]
 [ 1.  1.  1.  1.  1.  1.  1.  1.  1.  1.]
 [ 1.  1.  1.  1.  1.  1.  1.  1.  1.  1.]
 [ 1.  1.  1.  1.  1.  1.  1.  1.  1.  1.]]
start
descriptor
resource
/Users/dallan/Documents/Repos/databroker/databroker/assets/core.py:195: UserWarning: Ignoring attempt to insert Datum with duplicate datum_id, assuming that both ophyd and bluesky attempted to insert this document. Remove the Registry (`reg` parameter) from your ophyd instance to remove this warning.
  warnings.warn("Ignoring attempt to insert Datum with duplicate "
datum
/Users/dallan/Documents/Repos/databroker/databroker/assets/core.py:150: UserWarning: Ignoring attempt to insert Resource with duplicate uid, assuming that both ophyd and bluesky attempted to insert this document. Remove the Registry (`reg` parameter) from your ophyd instance to remove this warning.
  warnings.warn("Ignoring attempt to insert Resource with duplicate "
event
stop
[[ 1.  1.  1.  1.  1.  1.  1.  1.  1.  1.]
 [ 1.  1.  1.  1.  1.  1.  1.  1.  1.  1.]
 [ 1.  1.  1.  1.  1.  1.  1.  1.  1.  1.]
 [ 1.  1.  1.  1.  1.  1.  1.  1.  1.  1.]
 [ 1.  1.  1.  1.  1.  1.  1.  1.  1.  1.]
 [ 1.  1.  1.  1.  1.  1.  1.  1.  1.  1.]
 [ 1.  1.  1.  1.  1.  1.  1.  1.  1.  1.]
 [ 1.  1.  1.  1.  1.  1.  1.  1.  1.  1.]
 [ 1.  1.  1.  1.  1.  1.  1.  1.  1.  1.]
 [ 1.  1.  1.  1.  1.  1.  1.  1.  1.  1.]]
```

danielballan:
Notes on back-compat:

* If you have don't have new ophyd or any devices that implement the new `collect_asset_docs` method, bluesky won't hit the codepaths that would cause it to need the new event-model, and any combination of new or old databroker, bluesky, and event_model work fine.
* If you have new ophyd and a device that implements the new `collect_asset_docs` but old bluesky, bluesky will ignore `collect_asset_docs` (and ophyd will still insert the old way) so any combination of new or old databroker, bluesky, and event_model work fine.
* If you have new ophyd and a device that implements the new `collect_asset_docs` and new bluesky, bluesky will try to emit Datum and Resource documents, so you will need new databroker and event_model as well.

PR #323, opened 2017-12-15
FIX/TST: Fix a bug in the mongoquery backend.

danielballan:
This came up in yesterday's tutorial. The bug is isolated to the mongoquery
backend, which we only use for testing/teaching purposes so far, so it should
not affect the floor.

### DIFF
--- a/databroker/headersource/mongoquery.py
+++ b/databroker/headersource/mongoquery.py
@@ -1,5 +1,6 @@
 from __future__ import absolute_import
 
+import copy
 import os
 import json
 from mongoquery import Query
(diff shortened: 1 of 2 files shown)


tacaswell:
I suspect this may fix a confusing bug @CJ-Wright has reported.

CJ-Wright:
yey! (which one was that?)

tacaswell:
@CJ-Wright has reported a couple about duplicate calls not behaving correctly

CJ-Wright:
ah yes now I remember.

Issue #274, opened 2017-08-17
Document 'install_sentinels' in the configuration section.


danielballan:

jrmlhermitte:
closed by #324

PR #324, opened 2017-12-19
Doc Add version sentinels

jrmlhermitte:
I think we need this somewhere (discussion about version sentinels). This is a suggestion, feel free to overtake it. 

When I build by the way, I get a few unrelated warnings:
(for ex: doc/source/_as_gen/databroker.Header.values.rst: WARNING: document isn't included in any toctre)

@danielballan

### DIFF
--- a/doc/source/creating.rst
+++ b/doc/source/creating.rst
@@ -1,0 +1,29 @@
+***********************
+Creating a New Database
+***********************
+
+In some cases, you may want to create a new database. Currently, storage and
+retrieval is implemented in MongoDB so instructions for MongoDB will follow.
+The implementation for other types of databases is likely similar.
+
+Creation in MongoDB
+-------------------
+
+The creation of a database in MongoDB is simple. Simply create a configuration
+as described in :ref:`configuration`. The only requirement is that the MongoDB
+instance is running on the machine specified. The database need not exist yet.
+The next step is to install a version sentinel. This is done as follows:
+
+.. code-block:: python
+
+    # Instantiate databroker instance
+    from databroker import Broker
+    db_analysis = Broker.named(config_name)
+
+    # install sentinels
+    databroker.assets.utils.install_sentinels(db_analysis.reg.config,
+                                              version=1)
+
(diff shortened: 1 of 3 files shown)


doc/source/creating.rst
```diff
@@ -0,0 +1,29 @@
...
+    # Instantiate databroker instance
+    from databroker import Broker
+    db_analysis = Broker.named(config_name)
+
+    # install sentinels
+    databroker.assets.utils.install_sentinels(db_analysis.reg.config,
+                                              version=1)
+
+where ``config_name`` is the name of your configuration and ``version=1`` refers
+to the version of filestore you are using (it is currently ``1`` as of this
```
danielballan: version of filestore -> version of asset registry. We haven't been great at sticking to this name change in our conversations, but we do _intend_ to drop "filestore" from the lexicon.

doc/source/creating.rst
```diff
@@ -0,0 +1,29 @@
...
+
+In some cases, you may want to create a new database. Currently, storage and
+retrieval is implemented in MongoDB so instructions for MongoDB will follow.
+The implementation for other types of databases is likely similar.
+
+Creation in MongoDB
+-------------------
+
+The creation of a database in MongoDB is simple. Simply create a configuration
+as described in :ref:`configuration_doc`. The only requirement is that the MongoDB
```
danielballan: If you just want to point to the configuration.rst file generically (rather than a specific section within it, you can use the directive

```
:doc:`configuration`
```

with no need to create a specific reference.

danielballan:
About those sphinx warnings: Header provides some of the standard dict methods (keys, values, items, etc.). We should add an autosummary table to document those as well.

doc/source/creating.rst
jrmlhermitte: great thanks for the advice!

doc/source/creating.rst
jrmlhermitte: oops yes, thanks!

danielballan:
Closes #274

PR #328, opened 2017-12-21
BUG: removed invalid kwarg to get_images

jrmlhermitte:
@yugangzhang's code still uses `get_images` in some places. We're working to move away from that. However, in the meantime, could we fix the deprecated function? `handler_override` is no loonger an acccepted kwarg thanks!

### DIFF
--- a/databroker/pims_readers.py
+++ b/databroker/pims_readers.py
@@ -20,9 +20,6 @@ def get_images(headers, name, handler_registry=None):
         field name (data key) of a detector
     handler_registry : dict, optional
         mapping spec names (strings) to handlers (callable classes)
-    handler_override : callable class, optional
-        overrides registered handlers
-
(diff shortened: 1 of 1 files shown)


tacaswell:
It is used in the function though...

mrakitin:
Also, the docstring should be updated as well to exclude the corresponding argument, if we decide to remove it.

jrmlhermitte:
@tacaswell sorry this was a separate edit made from the function i tested on the chx server. updated.
@mrakitin thanks removed the doc string

danielballan:
Test coverage of `get_images` is bad. We are moving away it from it, though.

jrmlhermitte:
thanks @danielballan ! I agree, probably not worth adding tests. We'll discourage its use.

Issue #325, opened 2017-12-19
extend config to allow for root_map


jrmlhermitte:
currently, configuration allows for the setting of many parameters, including the file handlers.
It would be nice to have a feature that allows the setting of a global `root_map`

reference : http://nsls-ii.github.io/databroker/configuration.html

jrmlhermitte:
@tacaswell @danielballan @mrakitin

danielballan:
relevant to @CJ-Wright's interests as well -- I wonder if we can tempt him to hammer out a quick PR ;-)

CJ-Wright:
I don't know if any of my PRs in recent history have been *quick*. I will not be able to get to it this week most likely, I'm doing a deep dive in SHED.

jrmlhermitte:
closed by #332

PR #332, opened 2018-01-02
Added root_map to config

jrmlhermitte:
Added `root_map` to config.
@danielballan 

I still need to review it (this was a stashed commit done earlier).

### DIFF
--- a/databroker/_core.py
+++ b/databroker/_core.py
@@ -497,7 +497,6 @@ def register_builtin_handlers(reg):
                              cls.specs)
 
 
-
 def get_fields(header, name=None):
     """
     Return the set of all field names (a.k.a "data keys") in a header.
@@ -1867,8 +1866,8 @@ class BrokerES(object):
                 datums = self.reg.datum_gen_given_resource(uid)
                 for datum in datums:
                     db.reg.insert_datum(new_res,
-                                       datum['datum_id'],
-                                       datum['datum_kwargs'])
+                                        datum['datum_id'],
+                                        datum['datum_kwargs'])
         return file_pairs
 
     def export_size(self, headers):
@@ -2148,15 +2147,19 @@ class Broker(BrokerES):
         # if 'root_map' in config, set the root_map
         if 'root_map' in config:
             root_map = config['root_map']
-            try:
-                if len(root_map) == 2:
-                    # change to dict
-                    root_map = {root_map[0] : root_map[1]}
-                    db.reg.set_root_map(root_map)
-            except TypeError:
-                print("root_map is not right type in config file "
-                      "Must be tuple of length 2 (see documentation)")
-                raise
+            if not isinstance(root_map, dict):
+                print("root_map is not right type in config file. "
+                      "It must be tuple of length 2 (see documentation)")
+                raise TypeError
+            # TODO : Allow multiple mappings, need to update
+            # db.reg.set_root_map first
+            if len(root_map) != 1:
+                errormsg = "Received {} ".format(len(root_map))
+                errormsg += "root_map definitions. Only 1 is currently"
+                errormsg += " supported"
+                raise ValueError(errormsg)
+            for oldroot, newroot in root_map.items():
+                db.reg.set_root_map({oldroot: newroot})
         return db
 
     @classmethod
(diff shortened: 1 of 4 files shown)


jrmlhermitte:
addresses #325

databroker/_core.py
```diff
@@ -2145,6 +2145,17 @@ def from_config(cls, config, auto_register=True):
...
+        # if 'root_map' in config, set the root_map
+        if 'root_map' in config:
+            root_map = config['root_map']
+            try:
+                if len(root_map) == 2:
+                    # change to dict
+                    root_map = {root_map[0] : root_map[1]}
+                    db.reg.set_root_map(root_map)
+            except TypeError:
+                warnings.warn("root_map is not right type in config file"
```
danielballan: I think this should be a `TypeError` instead of a warning, but it is good to catch the plain `TypeError` and re-raise one with more detail, as you have done. Note that this needs a trailing space so it does not read "config fileMust".

tacaswell:
Agree with @danielballan This should fail hard (and fast) if there is a root map and it can not make sense of it.

jrmlhermitte:
I agree. Sorry for missing this (meant to do this a lot earlier)

databroker/_core.py
```diff
@@ -2145,6 +2145,18 @@ def from_config(cls, config, auto_register=True):
         for spec, handler in config.get('handlers', {}).items():
             cls = load_cls(handler)
             db.reg.register_handler(spec, cls)
+        # if 'root_map' in config, set the root_map
+        if 'root_map' in config:
+            root_map = config['root_map']
+            try:
```
tacaswell: This will pass `root_map == 'foobar'`

doc/source/configuration.rst
```diff
@@ -74,6 +74,15 @@ classes that load externally stored data.
             module: 'databroker.assets.path_only_handlers'
             class: 'RawHandler'
 
+Optionally, you may set a root_map, which comes in handy when the handlers
+involves mounted files that have been moved to a different mount point in the
+file system.
+.. code-block:: yaml
+    root_map : ['old_root', 'new_root']
```
tacaswell: I think we need to be able to support many re-mappings, not just one.  Why not let rootmap be a dict?

jrmlhermitte:
This is ready for review, but py3.6 keeps failing with an unrelated Server time out.

Looking ahead, we may need to start implementing the acceptance of multiple `root_map`s. (This is out of scope of this PR) We currently need this at xpd because the mounted file system has changed location, which removes the freedom of using this feature for small tests/fixes/patches.

databroker/_core.py
```diff
@@ -2145,6 +2144,22 @@ def from_config(cls, config, auto_register=True):
         for spec, handler in config.get('handlers', {}).items():
             cls = load_cls(handler)
             db.reg.register_handler(spec, cls)
+        # if 'root_map' in config, set the root_map
+        if 'root_map' in config:
+            root_map = config['root_map']
+            if not isinstance(root_map, dict):
+                print("root_map is not right type in config file. "
+                      "It must be tuple of length 2 (see documentation)")
```
tacaswell: This error message is wrong now.  Can you put this string in the exception rather than printing?

databroker/_core.py
```diff
@@ -2145,6 +2144,22 @@ def from_config(cls, config, auto_register=True):
...
+                raise TypeError
+            # TODO : Allow multiple mappings, need to update
+            # db.reg.set_root_map first
+            if len(root_map) != 1:
```
tacaswell: Why this restriction?    I think this entire block can just be `db.reg.set_rootmap(root_map)`

jrmlhermitte:
Thanks @tacaswell . I should have read further. I understood multiple `root_map`s not being supported. If that is the case, it should be tested. I added an extra test for that in `assets/tests/test_fs_mutate.py`
Let me know if this is good

databroker/assets/tests/test_fs_mutate.py
```diff
@@ -232,3 +233,16 @@ def local_handler(rpath):
         path = fs.retrieve(dm['datum_id'])
 
     assert path == os.path.join('baz', 'foo')
+
+    # test two root maps work
+    res = fs.insert_resource('root-test', 'foo', {}, root='baz2')
```
tacaswell: should this be `bar2`?

jrmlhermitte: oops yes. tests passed but to be consistent with the logic flow, it should be `bar2`. thanks!

jrmlhermitte:
thanks @danielballan ! @tacaswell ?

jrmlhermitte:
thnanks!

PR #335, opened 2018-01-12
FIX: ensure there is only 1 version of DuplicateHandler

### LINKED ISSUE #334: Duplicate DuplicateHandler exceptions
We should remove one. Here they are:
https://github.com/NSLS-II/databroker/blob/master/databroker/assets/handlers_base.py#L27
and
https://github.com/NSLS-II/databroker/blob/master/databroker/assets/base_registry.py#L19

Probably remove the former. Noticed this when trying to catch a DuplicateHandler exception, and not catching it!

tacaswell:
- Let the ground truth be in handler_base
 - use the exception class stashed on the BaseRegistryRO object to
   allow customization by sub-classes and so a registry instance
   bring's its exception classes with it

closes #334

### DIFF
--- a/databroker/assets/base_registry.py
+++ b/databroker/assets/base_registry.py
@@ -14,10 +14,7 @@ from ..utils import ensure_path_exists
 from pkg_resources import resource_filename
 import json
 from .utils import _ChainMap
-
-
-class DuplicateHandler(RuntimeError):
-    pass
+from .handlers_base import DuplicateHandler
 
 
 logger = logging.getLogger(__name__)
@@ -189,7 +186,7 @@ class BaseRegistryRO(object):
         if (not overwrite) and (key in self.handler_reg):
             if self.handler_reg[key] is handler:
                 return
-            raise DuplicateHandler(
+            raise self.DuplicateHandler(
                 "You are trying to register a second handler "
                 "for spec {}, {}".format(key, self))
 


jrmlhermitte:
there are lots of seemingly unrelated timing errors in Travis. do we know where they come from?

tacaswell:
It is so that the registry can bring it's exceptions with it so in user code you can do

```
try:
    db.reg.register_handler(...)
except db.reg.DuplicateHandler:
    ...
```

and now you don't have to import `databroker` in your user code (if you assume one will be passed into you).   This is a pattern that came from before the big split, but gives us a layer of indirection if we (or one of our collaborators) want to implement a completely independent version of `Databroker` that just duck-types correctly.

tacaswell:
There is a bug with timezones which due to some things being naive and localized, our configuration being US/New_York, and the time on the travis machines being Berlin time means that from about 9pm - midnight our time the tests don't pass.  Restarted them now that it is morning.

The other class of failures is the server for mdservice sometimes does not start and we are not sure why.

jrmlhermitte:
thanks. useful information.
and that's great, thanks for the information. yes, duck typing is more pythonic

jrmlhermitte:
im restarting the failed py36 build (due to time out) and can merge once it passes
(not sure what the timeout is either, if you know I'd be interested to hear)

jrmlhermitte:
```
        from ophyd import sim
>       db1 = broker_factory()
```

```
        while True:
            if time.time() - startup_time > TIMEOUT:
>               raise Exception("Server startup timed out.")
E               Exception: Server startup timed out.
```

keeps happening. any idea?

tacaswell:
Suspect it is something todo with port randomization, I think @arkilic had some theories.

jrmlhermitte:
This looks good to me. I'll merge it. thanks @tacaswell !

PR #337, opened 2018-01-18
Rename start_time, end_time -> since, until.

### LINKED ISSUE #320: start_time, end_time -> since, until?
`Broker.__call__` generally passes keyword arguments through to build to a Mongo query, but it gives special significance to `start_time` and `end_time` (if present) to build a time-based query. These names are confusing because they suggest the start and end of the _run_, when in fact we are filtering on RunStart time alone. How about we use "since" and "until" instead, borrowing these names from `git log`?

danielballan:
These names of confusing, especially 'end_time' because people think
it filters on the time that a run ends. The names 'since' and 'until'
come from `git log`.

Closes #320

I also found some code that should have been deleted in the
"octomerge".

### DIFF
--- a/databroker/_core.py
+++ b/databroker/_core.py
@@ -1128,8 +1128,8 @@ class BrokerES(object):
 
     def _format_time(self, val):
         "close over the timezone config"
-        # modifies a query dict in place, remove keys 'start_time' and
-        # 'stop_time' and adding $lte and/or $gte queries on 'time' key
+        # modifies a query dict in place, remove keys 'since' and
+        # 'until' and adding $lte and/or $gte queries on 'time' key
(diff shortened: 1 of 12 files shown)


jrmlhermitte:
:+1:
I really like how `since` and `until` are less likely to be accidentally used as metadata as opposed to `start_time` etc

danielballan:
Is this good to merge then? Tests pass; the coverage complaints are with regard to the archiver plugin, which is its own separate can of worms.

jrmlhermitte:
I'm fine with everything, but I was not part of the "octomerge". Maybe have @tacaswell  sign off on the removal of that vestigial code?

PR #338, opened 2018-01-22
Revert "Refactor HDF5 handlers and add DASK handler"

tacaswell:
Reverts NSLS-II/databroker#308

This breaks sub-classes which passing in 'key' as a string (as is still documented).

### DIFF
--- a/databroker/assets/tests/test_handlers.py
+++ b/databroker/assets/tests/test_handlers.py
@@ -10,10 +10,8 @@ import tifffile
 
 from ..handlers import AreaDetectorHDF5Handler
 from ..handlers import AreaDetectorHDF5SWMRHandler
-from ..handlers import AreaDetectorHDF5DaskHandler
 from ..handlers import AreaDetectorHDF5TimestampHandler
 from ..handlers import AreaDetectorHDF5SWMRTimestampHandler
-from ..handlers import AreaDetectorHDF5PyHandler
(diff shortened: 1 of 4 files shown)


danielballan:
Will merging this break eiger-io, @jrmlhermitte?

danielballan:
I answered my own question: everything in eiger-io derives from `HandlerBase`, so this should be fine.

@stuwilkins Do you have the bandwidth to take another pass at #308 and do it in a backward-compatible way? It's OK if not; just let us know if we should pick this up.

danielballan:
Randomly metadataserver failure, per usual. Merging!

databroker/assets/tests/test_handlers.py
```diff
@@ -39,18 +37,12 @@ def fs_with_handlers(request, fs_cls):
     fs.register_handler('AD_HDF5_TS', AreaDetectorHDF5TimestampHandler)
     fs.register_handler('AD_HDF5_SWMR_TS',
                         AreaDetectorHDF5SWMRTimestampHandler)
-    fs.register_handler('AD_HDF5_DASK',
```
jrmlhermitte, line 42: this did not look like good practice here

jrmlhermitte: (Specs should be agnostic of file formats or how they're read right?)

PR #339, opened 2018-01-24
Add get_config

jrmlhermitte:
before i forget, i think this may be useful.
@danielballan and I discussed this.

If we want to pass minimal information about the databroker connection between processes, this could be useful. Still need to discuss serialization of headers with @tacaswell 

PS : this is not very clean. If we were to add more features to the config file, we need to remember to modify `get_config` and the test. If anyone has suggestions id like to hear

### DIFF
--- a/databroker/tests/test_config.py
+++ b/databroker/tests/test_config.py
@@ -58,13 +58,9 @@ def test_from_config():
     reg_config = config['assets']
     root_map_config = config['root_map']
 
-    for key, val in mds_config.items():
-        assert mds_example[key] == val
-    for key, val in reg_config.items():
-        assert reg_example[key] == val
-    for key, val in root_map_config.items():
-        assert root_map_example[key] == val
-
+    assert mds_example == mds_config
+    assert reg_example == reg_config
+    assert root_map_example == root_map_config
 
 
 def test_handler_registration():
--- a/databroker/tests/test_config.py
+++ b/databroker/tests/test_config.py
@@ -46,7 +46,25 @@ EXAMPLE = {
 
 
 def test_from_config():
-    Broker.from_config(EXAMPLE)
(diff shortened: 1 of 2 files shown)


databroker/tests/test_config.py
```diff
@@ -46,7 +46,25 @@
...
+    # we explicitly test for parts we know should be accepted
+    mds_example = EXAMPLE['metadatastore']['config']
+    reg_example = EXAMPLE['assets']['config']
+    root_map_example = EXAMPLE['root_map']
+
+    mds_config = config['metadatastore']
+    reg_config = config['assets']
+    root_map_config = config['root_map']
+
+    for key, val in mds_config.items():
```
danielballan: This check won't catch the case where `config` is missing an item. It's probably simpler to just do `assert mds_config == mds_example`. Pytest provides some introspection that will give an element-by-element view of any differences, which I think the goal of the way you have factored this here.

jrmlhermitte: ah i was worried this would be false:
```py
In [1]: a = dict(a=1)

In [2]: b = dict(a=1)

In [3]: a==b
Out[3]: True
```

thanks ill add it in

jrmlhermitte:
Thanks @danielballan . For the record, this is not a `classmethod` since you need to probe an instance. I'm sure that's what you meant but just wanted to be sure. (i.e. `Broker.get_config(broker_instance)` is not possible but `broker_instance.get_config()` is)

danielballan:
Oh, yeah, definitely. I was being a little bit loose in calling tn `Broker.get_config()`.

jrmlhermitte:
so @danielballan and @tacaswell  is this good?

jrmlhermitte:
awesome thanks!

PR #346, opened 2018-02-14
Fix version

jrmlhermitte:


jrmlhermitte:
this fixes doc issue temporarily....

doc/source/tutorial.rst
```diff
@@ -15,7 +15,10 @@ concurrent "streams" of data in an organized way.
 
     import os
     import yaml
+    import epics
+    epics.__version__ = "0.0.0"
     from databroker import temp_config
+    from bluesky.examples import det, motor
```
tacaswell: This should be `from ophyd.sim`

jrmlhermitte: thanks

tacaswell:
The `_as_gen` files should not be under source control.

danielballan:
I rebased to remove generated files from PR. I discovered that `.gitignore` thought these files were under the directory `generated/` but in fact they were going in to `_as_gen/`. In the docs, I updated the autosummary directives to output to `generated/` which is consistent with our other projects.

danielballan:
Thanks for kicking this off @jrmlhermitte!

jrmlhermitte:
thanks @danielballan !

PR #347, opened 2018-02-22
Add stream_name option to fields

jrmlhermitte:
This looks like it's missing (currently trying to read fields only for a certain stream).
tested locally at QAS, still needs test.

### DIFF
--- a/databroker/tests/test_broker.py
+++ b/databroker/tests/test_broker.py
@@ -514,7 +514,6 @@ def test_stream_name(db, RE, hw):
     # subscribe db.insert
     RE.subscribe(db.insert)
 
-
     # custom plan that will generate two streams
     @run_decorator()
     def myplan(dets):
@@ -522,9 +521,10 @@ def test_stream_name(db, RE, hw):
 
         Meant for test only.
         '''
-        yield trigger_and_read([dets[0]], name='primary')
-        yield trigger_and_read([dets[1]], name='secondary')
-
+        for msg in trigger_and_read([dets[0]], name='primary'):
+            yield msg
+        for msg in trigger_and_read([dets[1]], name='secondary'):
+            yield msg
 
     # this test is meaningless (will always pass)
     # if our two detectors have the same name. Ensure this is true
--- a/databroker/tests/test_broker.py
+++ b/databroker/tests/test_broker.py
@@ -546,11 +546,13 @@ def test_stream_name(db, RE, hw):
 
     db.mds.insert_event(descriptor=desc_uid2, time=ttime.time(), uid=event_uid2,
                         data={'image2': datum_id2},
-                        timestamps={'image': ttime.time()}, seq_num=0)
+                        timestamps={'image2': ttime.time()}, seq_num=0)
 
     h = db[rs_uid]
 
-    assert h.fields() == {'image', 'image2'}
+    assert h.fields() == {'image', 'image2', 'det'}
+    assert h.fields(stream_name='primary') == {'det'}
+    assert h.fields(stream_name='injected') == {'image'}
     assert h.fields(stream_name='injected2') == {'image2'}
 
 
--- a/databroker/tests/test_broker.py
+++ b/databroker/tests/test_broker.py
@@ -510,6 +510,51 @@ def test_configuration(db, RE):
 
 
 @py3
+def test_stream_name(db, RE, hw):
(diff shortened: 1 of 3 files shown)


jrmlhermitte:
my local tests pass, this is ready to go (waiting on Travis)

databroker/tests/test_broker.py
```diff
@@ -509,6 +509,53 @@ class SynWithConfig(Device):
...
+                           'external': 'FILESTORE://simulated'}}
+    data_keys2 = {'image2': {'dtype': 'array', 'source': '', 'shape': [5, 5],
+                           'external': 'FILESTORE://simulated'}}
+    db.mds.insert_descriptor(run_start=rs_uid, data_keys=data_keys,
+                             time=ttime.time(), uid=desc_uid,
+                             name='injected')
+    db.mds.insert_descriptor(run_start=rs_uid, data_keys=data_keys2,
+                             time=ttime.time(), uid=desc_uid2,
+                             name='injected2')
+    db.mds.insert_event(descriptor=desc_uid, time=ttime.time(), uid=event_uid,
```
jrmlhermitte: insert_event is not really necessary. i can remove if needed

tacaswell: It is probably simpler to write a custom plan that uses trigger_and_read to add a second stream.

jrmlhermitte: @tacaswell This is a different PR. (Maybe I misunderstood)
I was trying to find a way to find fields versus stream_name. I could not find a way without this

jrmlhermitte: (this is completely unrelated to streaming for CHX)

jrmlhermitte: sorry i understand you now, wrote too quick... will do, thanks!

jrmlhermitte: actually, i would argue against this one. This would add in a bluesky dependency. I don't think we should add one if we don't have to. Is this okay as is?

tacaswell: We already have a bluesky dependency in this test (the `RE` fixture).

danielballan: Yep, to reiterate some stuff we talked about for the record here...

Originally, all the databroker tests inserted documents manually like this. It was verbose and hard to maintain, and sometimes didn't quite match up with the actual documents generated by bluesky as things changed. Therefore, we decided to use the RunEngine to generate all documents used in databroker tests. We use as a `RE` figure to do that.

A downside of this is that databroker still supports py2, while bluesky only support py3, so tests that use the `RE` fixture cannot be tested on py2. Such tests as decorated with the `@py3` decorator so that the py2 tests skip them. This obviously hurts on py2 test coverage. One possible solution is to generate all the document in Python 3, stash them somewhere online as JSON, and download them for use in the py2 tests. We hope to get around to doing that some day and boosting our py2 test coverage.

jrmlhermitte: thanks. i wrote a custom plan that writes to two different streams. tests pass on the test but  `test_find_by_string_time` seems to fail right now. any idea why this could be?

tacaswell:
We really need to fix that 'search by time' bug that I am pretty sure is caused by timezone being the *worst*.  Restarted the tests.

jrmlhermitte:
python2 won't parse the file because of `yield from`. changed to `yield`

databroker/tests/test_broker.py
```diff
@@ -522,8 +522,8 @@ def myplan(dets):
 
         Meant for test only.
         '''
-        yield from trigger_and_read([dets[0]], name='primary')
-        yield from trigger_and_read([dets[1]], name='secondary')
+        yield trigger_and_read([dets[0]], name='primary')
```
danielballan: Oh, I see what went wrong here. The Python 2 version of `yield from` is:

```py
for msg in trigger_and_read([dets[0]], name='primary')
    yield msg
```

Doing `yield trigger_and_read([dets[0]], name='primary')` yields the generator instance itself, not the messages can issue.

jrmlhermitte: thanks!

danielballan:
The test failures and random and unrelated ("Server startup timeout"). Thanks, @jrmlhermitte!

jrmlhermitte:
great thanks @danielballan !

PR #350, opened 2018-03-01
ENH: Implement Broker.insert for 'datum' and 'resource'

danielballan:
Implements databroker's pieces of
https://github.com/NSLS-II/databroker/pull/322, where Resource and
Datum docs are inserted the same as everything else.

This is designed to work with old ophyd configurations (where ophyd inserts
Resource/Datum docs directly into a Registry) running new ophyd objects (where
the RunEngine collects Resource/Datum docs and emits them). In this case, the
databroker will see two documents with the same uid, which would normally cause
it to error. To ease the transition, we the databroker will emit a warning if
it sees a duplicate insert but then just ignore it. This behavior is
configurable by a new kwarg, which is strict by default in the low-level API.
It is tolerant by default in the high-level ``Broker.insert`` API, only
temporarily to ease this transition.

### DIFF
--- a/databroker/_core.py
+++ b/databroker/_core.py
@@ -1097,10 +1097,25 @@ class BrokerES(object):
         doc : dict
             Document
         """
-        if name in {'start', 'stop'}:
+        if name in {'event', 'bulk_events', 'descriptor'}:
+            return self.event_source_for_insert.insert(name, doc)
+        # We are transitioning from ophyd objects inserting directly into a
+        # Registry to ophyd objects passing documents to the RunEngine which in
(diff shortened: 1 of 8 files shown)


databroker/_core.py
```diff
@@ -1097,10 +1097,25 @@ def insert(self, name, doc):
         doc : dict
             Document
         """
-        if name in {'start', 'stop'}:
+        if name in {'event', 'bulk_event', 'descriptor'}:
```
tacaswell: It looks like the key is 'bulk_events'

PR #356, opened 2018-04-27
ENH: pass the name of Broker.named into the broker

### LINKED ISSUE #354: expose databroker name
Would it be possible to expose the databroker's name after the object has been created?

tacaswell:
closes #354


danielballan:
I'm late to this party, but wouldn't it be better to include the name as an optional key in the configuration itself? We already have a description key that works this way, and I can't think of a good reason to treat 'name' differently.

In the current PR, `Broker.named('chx')` gets a name but `Broker.from_config('/etc/databroker/chx.yml')` does not. That seems less than ideal.

mrakitin:
Good point, @danielballan!

CJ-Wright:
It might be valuable to know if the databroker was loaded from a configuration or via a named broker. I could produce a configuration which *says* it's `chx` but actually has some subtle issues in the configuration which makes it not the same as the `/etc/...` databroker.

tacaswell:
If you want to give it a name you can pass the name through `from_config`.

One case for treating 'name' differently is that it is already in the yaml file name and via `Broker.named` so if we put it in the configuration what and it was _different_ which one would we use?

We definitely do not want to go down the path of recording if the configuration came from a path or a dictionary or ... as where the configuration comes from is not important and may lie (ex, say we recorded a path, then pickled and shipped the broker to another computer with a different file system).

danielballan:
OK, I've come full circle on this. I think the PR as it is is the best approach available to us.

mrakitin:
Travis is not happy:
```py
Warning, treated as error:
/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/databroker/_core.py:docstring of databroker.Broker.add_event_source:1:duplicate object description of databroker.Broker.add_event_source, other instance in /home/travis/build/NSLS-II/databroker/doc/source/_as_gen/databroker.Broker.add_event_source.rst, use :noindex: for one of them
make: *** [html] Error 2
```

danielballan:
Closer....

Completely lost as to where [this error](https://travis-ci.org/NSLS-II/databroker/jobs/374991500#L1760) is coming from. Works fine locally on latest version of sphinx.

tacaswell:
```
>>>-------------------------------------------------------------------------
Exception in /home/travis/build/NSLS-II/databroker/doc/source/assets.rst at block ending on line 290
Specify :okexcept: as an option in the ipython:: block to suppress this message
KeyErrorTraceback (most recent call last)
<ipython-input-34-37f44c86eaf3> in <module>()
----> 1 api_key = os.environ['FORECAST_IO_API_KEY']
~/virtualenv/python3.6.3/lib/python3.6/os.py in __getitem__(self, key)
    667         except KeyError:
    668             # raise KeyError with the original key value
--> 669             raise KeyError(key) from None
    670         return self.decodevalue(value)
    671 
KeyError: 'FORECAST_IO_API_KEY'
<<<-------------------------------------------------------------------------

```

Is the issue that we are missing an env?

tacaswell:
Yeah, secrets are not available in PRs (or I could put in a PR to print out your secrets!).

mrakitin:
The reason the build fails is `-W` passed to sphinx, which turns warnings into errors. It was introduced in https://github.com/NSLS-II/databroker/pull/322/files?utf8=%E2%9C%93&diff=unified#diff-f3644bba251750c9ec4a2361d6cc59ebL5.

mrakitin:
Not sure what's going on with `FORECAST_IO_API_KEY`, I recreated it with my API key (tested locally, worked fine). Still failing...

tacaswell:
The env is encrypted using some travis tools (you use their public key to encrypt it, they have the private key to decrypt it) so you can commit secrets to git without leaking them (good!), but if they were available on PR builds (which run when _anyone_ opens a PR) then an attacker could open a PR that echos your env variable and hence capture your keys (bad!).

I am not a huge fan of requiring an API key to build the docs....

tacaswell:
@danielballan can you please do something about that example?

danielballan:
The forecast example is not what is breaking the build -- IPython's warnings don't trigger `-W`. See previous comment for the warning that is actually blocking us here. I still cannot understand it.

mrakitin:
Great job, @tacaswell, now the docs building passes on Travis!

tacaswell:
For those following at home, the source of the warning was an update to one of the sphinx libraries that generated entries for properties / attributes

PR #359, opened 2018-04-27
updated ArchiveEventSource

malitsky:


tacaswell:
You don't need to make new PRs, you can just keep pushing new commits to your branch and the PR will automatically update.

databroker/eventsource/archiver.py
```diff
@@ -73,65 +119,175 @@ def descriptors_given_header(self, header):
                         'external_url': self.url}
                 descs.append(desc)
             self._descriptors[run_start_uid] = descs
-            return [ d for d in self._descriptors]
+            
+            return self._descriptors[run_start_uid]
```
tacaswell: We probably want to return `list(self._descriptorts[run_start_uid)` because lists are mutable and if the user mutates the return value we want to be protected from that.

tacaswell:
Is there a plan to test this like we test the other event sources?

malitsky:
I expect to have the corresponding tests by the end of May. Therefore, I will temporarily close this pull request and maintain the updated version on my fork repository

tacaswell:
We need this functionality ready to deploy for next cycle.  Lets get these changes reviewed and merged and delay testing until we have a bit more time.

tacaswell:
Can you please squash this to one or two commits with good commit messages?

databroker/eventsource/archiver.py
```diff
@@ -73,65 +119,155 @@ def descriptors_given_header(self, header):
                         'external_url': self.url}
                 descs.append(desc)
             self._descriptors[run_start_uid] = descs
-            return [ d for d in self._descriptors]
+            
+            return list(self._descriptors[run_start_uid])      
+            # return [ d for d in self._descriptors]
```
danielballan: You may as well delete this line and the commented-out lines below it.

databroker/eventsource/archiver.py
```diff
@@ -73,65 +119,155 @@ def descriptors_given_header(self, header):
...
             req = requests.get(self.archiver_addr, params=params, stream=True)
             req.raise_for_status()
             raw, = req.json()
+            
             timestamps = [x['secs'] for x in raw['data']]
             data = [x['val'] for x in raw['data']]
-            for seq_num, (d, t) in enumerate(zip(data, timestamps)):
-                doc = {'data': {name: d},
+            
+            for seq_num, (v, t) in enumerate(zip(data, timestamps)):
```
danielballan: We start `seq_num` at `1` -- which is unfortunate in the sense that it's not very Python, but at this point we'd better be consistent. You can do `enumerate(..., start=1)` to start counting from `1`.

malitsky: sure

malitsky:
The updated version preserves the input parameters of the Archiver Appliance constructor, updates several methods (fields_given_header, descriptors_given_header, and doc_given_header) to be consistent with EventSourceShim and implementes a new version of table_given_header. Specifically, table_given_header joins the time series of different PVs based on their timestamps and then enumerates rows of the resulting DataFrame

malitsky:
The updated version includes documentation, the ArchiverReader and a new tables_given_times method addressing the CSX team's requirements.

danielballan:
If I recall correctly, `tables_given_times` is not part of the established API for event sources. Can you explain what requirements of the CSX team's this is aiming to meet? I wonder if we can find a way to meet them that satisfies the standard interface, or if we can agree on a change that needs to be made to that interface.

databroker/eventsource/archiver.py
```diff
@@ -65,77 +107,285 @@ def descriptors_given_header(self, header):
...
-
     Return
     ------
     time
         as ISO-8601 format
     """
     t = datetime.fromtimestamp(t)
     return timezone.localize(t).replace(microsecond=0).isoformat()
+
+class ArchiverReader(object):
```
tacaswell: Can this be merged into https://github.com/NSLS-II/databroker/blob/master/databroker/_core.py#L2086 ?  I don't think we need a separate configuration system for this.

doc/source/archiver.rst
```diff
@@ -0,0 +1,122 @@
...
+After integrating ArchiverEventSources with :class:`Broker`, PV data can
+be retrieved with the standard :meth:`Broker.get_table` method:
+
+.. code-block:: python
+
+   # select header
+   hdr = db[69209]
+
+   stream_name = 'archiver_pv1'
+   df = db.get_table(hdr, stream_name)
```
tacaswell: This can just be `hdr.table(stream_name)`  [ fixed typo]

databroker/eventsource/archiver.py
```diff
@@ -37,18 +39,58 @@ def __init__(self, url, timezone, pvs):
...
+        Parameters
+        ----------
+        header: Header
+            not used. stream_names are generated from pvs
+        Returns
+        -------
+        list: e.g., [archiver_<user-defined PV name>, ...]
+        """
+            
         return ['archiver_{}'.format(name) for name in self.pvs]
```
tacaswell: Can we not tack archiver on the front?  Given that they provided us with nice names on configuration, might as well just use them.

doc/source/archiver.rst
```diff
@@ -0,0 +1,122 @@
...
+
+For example, the following code illustrates the configuration of
+two :class:`ArchiverEventSource` instances:
+
+.. code-block:: python
+
+    # CSX ArchiverEventSource
+    aes_config_csx = {
+		'name' : 'csx',
+		'url' : 'http://xf23id-ca.cs.nsls2.local:17668',
```
tacaswell: Do we have a list someplace of all of the archiver URLs?  That should probably go into playbooks.

doc/source/archiver.rst
mrakitin: @tacaswell, `db.table(stream_name)` or `hdr.table(stream_name)`?

tacaswell: sorry! `hdr.table(stream_name)`

malitsky:
The ArchiverReader was introduced to provide a configuration procedure
for the existing ArchiverEventSource class without any changes to the databroker
files. Now, you are suggesting to change the _core.py. This task can be implemented in multiple ways and would require a meeting to discuss the trade-offs among the different views.

tacaswell:
> The ArchiverReader was introduced to provide a configuration procedure
> for the existing ArchiverEventSource class without any changes to the databroker
> files. Now, you are suggesting to change the _core.py. This task can be implemented in multiple ways and would require a meeting to discuss the trade-offs among the different views.

I thought the plan was to integrate this with databroker as tightly as was practical.

tacaswell:
Can you please rebase this so the tests will run?

malitsky:
Can you elaborate the issue?

danielballan:
@malitsky Your branch cannot be automatically merged into the master branch because it needs to [rebased](https://git-scm.com/docs/git-rebase). This issue is blocking our CI tests from running, because they attempt to check out a temporary branch, _automatically merge the PR branch into the master branch_, and run the tests on that. This is a way of checking in advance whether actually merging the PR branch into master will break any tests.

malitsky:
I can make it by tomorrow morning.

danielballan:
I think [this comment](https://github.com/NSLS-II/databroker/pull/359#discussion_r187463138) was suggesting that we need to make the config system more flexible (so that Archiver configuration can be passed in) but not in such a way that the token `archiver` gets hard-coded anywhere in`_core.py`. Extending the existing pattern, users can pass in a module name, class name, and dictionary of keyword arguments.

databroker/_core.py
```diff
@@ -2063,7 +2070,7 @@ class Broker(BrokerES):
         to do all registration manually.
 
     """
-    def __init__(self, mds, reg=None, plugins=None, filters=None,
+    def __init__(self, mds, reg=None, plugins=None, filters=None,  archivers = None,
```
danielballan: We should definitely not add an `archivers` keyword argument to `Broker`. The Archiver should be "just another data source".

databroker/eventsource/shim.py
```diff
@@ -39,9 +42,9 @@ def stream_names_given_header(self, header):
         return set(d.get('name', 'primary') for d in
                    self.descriptors_given_header(header))
 
-    def fields_given_header(self, header, stream_name=ALL):
+    def fields_given_header(self, header):
```
danielballan: What's the purpose of this change?

malitsky:
_core.py extended to integrate archiver event sources. The following script and configuration file demonstrates this extension:

https://github.com/malitsky/nsls2/blob/master/areventsource/notebooks/archiver-reader.db.5.13.18.ipynb

https://github.com/malitsky/nsls2/blob/master/areventsource/notebooks/csx_with_archivers.yml

databroker/_core.py
```diff
@@ -2100,8 +2107,11 @@ def __init__(self, mds, reg=None, plugins=None, filters=None,
             warnings.warn("Future versions of the databroker will not accept "
                           "'filters' in __init__. Set them using the filters "
                           "attribute after initialization.", stacklevel=2)
+        ess = [EventSourceShim(mds, reg)]
+        if event_sources:
+            ess += archivers.event_sources()
```
tacaswell: I think this will fail with a local name error?

databroker/eventsource/archiver.py
malitsky: yes, it is used as (1) a factory for ArchiverEventSources:
https://github.com/malitsky/nsls2/blob/master/areventsource/notebooks/csx_with_archivers.yml

and as (2) a stand-alone class for the CSX extensions
https://github.com/malitsky/nsls2/blob/master/areventsource/notebooks/archiver-reader.db.5.13.18.ipynb

databroker/_core.py
```diff
@@ -2087,7 +2094,7 @@ class Broker(BrokerES):
     name : str, optional
         The name of the broker
     """
-    def __init__(self, mds, reg=None, plugins=None, filters=None,
+    def __init__(self, mds, reg=None, plugins=None, filters=None, event_sources = None,
```
tacaswell: put `event_sources` last so that we do not beak API of people using many positional arguements.

tacaswell: This should be a list of EventSources or None

databroker/eventsource/archiver.py
tacaswell: As discussed in person, please make this class private.

malitsky:
updated archiver.py and _core.py addressing suggestions

stuartcampbell:
I really don't see what `ArchiveReader()` adds as a benefit for the down side of adding another place to configure things from.  The `._from_config()` static method should really be public on the `ArchiveEventSource`.  

I really want to see the config be done in one place for all the Event Sources - there shouldn't be something specific for archivers - they are just another event source. 

I don't think we should be adding `_` private methods for something like this.   As in I don't see why we are making archivers special for our particular epics use case.  We may (and almost certainly will) want other archivers and eventsources in the future.

databroker/eventsource/shim.py
```diff
@@ -24,6 +24,10 @@ class EventSourceShim(object):
     into the implementations
     '''
 
+    @property
```
danielballan, line 27: What is this for?

malitsky: it is added to be consistent with the db.ES[name] for accessing event sources

malitsky: _ArchiverReader works as a (internal/private) factory for configuring multiple archiver event sources under the Broker API. This approach has been already blessed by Tom.

stuartcampbell: I know - but I disagree with Tom on this.

danielballan:
I share @stuartcampbell's concerns about introducing a special, private ` _ArchiverReader` class. I think the right path here is to work this into the core config-parsing system.

I think we need to consider postponing this to the next release.
* It doesn't fix a bug; it's an enhancement.
* Doing it properly requires touching the core config-parsing system, which risks breaking critical existing functionality if we get it wrong.
* As this is the first external "event source," this will set some precedents that we will have to stick with forever (or suffer through an API break). I see a handful of interface and naming choices that I would like time to weigh in on, once this is in a working form. I don't think there is time for that today.

malitsky:
it is used for configuring multiple archiver event sources (accelerator and beamline) with one entry in the databroker configuration file:

archivers:
    module: databroker.eventsource.archiver
    class: _ArchiverReader
    config:
        arch_csx: 
             url: 'http://xf23id-ca.cs.nsls2.local:17668'
        arch_acc:
             url: 'http://arcapp01.cs.nsls2.local:17668'
             pvs:
                  test_watT: 'UT:SB1-Cu:1{}T:Prmry-I'

malitsky:
metadatastore:
    module: databroker.headersource.mongo
    class: MDS
    config:
        host: 'xf23id-ca'
        port: 27017
        database: 'datastore2'
        timezone: 'US/Eastern'
assets:
    module: databroker.assets.mongo
    class: Registry
    config:
        host: 'xf23id-ca'
        port: 27017
        database: 'filestore'
archivers:
    module: databroker.eventsource.archiver
    class: _ArchiverReader
    config:
        arch_csx: 
             url: 'http://xf23id-ca.cs.nsls2.local:17668'
        arch_acc:
             url: 'http://arcapp01.cs.nsls2.local:17668'
             pvs:
                  test_watT: 'UT:SB1-Cu:1{}T:Prmry-I'

danielballan:
Gothca, but:

> The ._from_config() static method should really be public on the ArchiveEventSource.

tacaswell:
I suggested renaming to `_ArchiverReader` as a temporary state so that we could get functionality in and put off sorting out if we want to keep it or not.  

Long-term I do not think we need `ArchiverReader` and all of the configuration file parsing / ingest should happen in `Broker.from_config`.  If @stuartcampbell would like the configuration centralized now, then I am 100% on board with that.

malitsky:
So, I will keep _ArchiverReader as a (internal) temporary mechanism for configuring multiple archiver event sources defined in the databroker configuration file.

databroker/_core.py
```diff
@@ -1102,6 +1102,13 @@ def __init__(self, hs, event_sources, assets, external_fetchers,
         self.registry_for_insert = self.event_sources[0]
         self.prepare_hook = wrap_in_deprecated_doct
 
+    @property
```
danielballan: The name `ES` strikes me as a bit cryptic. Why not `event_sources`, which is more self-describing and also symmetric with the name of the argument to `__init__`?

databroker/_core.py
```diff
@@ -2087,8 +2094,8 @@ class Broker(BrokerES):
     name : str, optional
         The name of the broker
     """
-    def __init__(self, mds, reg=None, plugins=None, filters=None,
-                 auto_register=True, external_fetchers=None, name=None):
+    def __init__(self, mds, reg=None, plugins=None, filters=None,
```
danielballan: whitespace issues here

databroker/_core.py
```diff
@@ -2157,8 +2167,14 @@ def from_config(cls, config, auto_register=True, name=None):
         # Instantiate component classes.
         mds = mds_cls(config['metadatastore']['config'])
         assets = assets_cls(config['assets']['config'])
+        evsources = None
+        if 'archivers' in config:
```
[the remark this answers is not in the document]
malitsky: Again, _ArchiverReader helps to use one entry/one class to support multiple ArchiverEventSources. In any case, we need some working solution for future generalization

databroker/_core.py
danielballan: No worries, it wasn't a personal criticism (doesn't matter who suggested it). Can we come up with a more  self-describing name? Is there a good reason why `event_sources` in `__init__` is a list and not a dict with names?

databroker/_core.py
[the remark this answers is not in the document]
danielballan: The generalization I am thinking of is not generalization over multiple ArchiverEventSources but rather generalization over any kind of EventSources. If the token `'archivers'` enters the `Broker` class, then that is a sign to me that we have not been as general as I would like us to be.

doc/source/archiver.rst
```diff
@@ -0,0 +1,61 @@
...
+    archivers:
+      module: databroker.eventsource.archiver
+      class: _ArchiverReader
+      config:
+        arch_csx: 
+             url: 'http://xf23id-ca.cs.nsls2.local:17668'
+        arch_acc:
+             url: 'http://arcapp01.cs.nsls2.local:17668'
+             pvs:
+	       test_watT: 'UT:SB1-Cu:1{}T:Prmry-I'
```
danielballan: Indentation here. Also, just curious, what is `watT`?

doc/source/archiver.rst
```diff
@@ -26,7 +26,8 @@ two ArchiverEventSources, arch_csx and arch_acc:
         arch_acc:
              url: 'http://arcapp01.cs.nsls2.local:17668'
              pvs:
-	       test_watT: 'UT:SB1-Cu:1{}T:Prmry-I'		
+               test_watT: 'UT:SB1-Cu:1{}T:Prmry-I'
+
```
danielballan: It looks like this commit removed some indentation and added new indentation.

malitsky:
ok. I moved the factory method into ArchiverEventSource. The configuration works. Just trying to find some database, since CSX is disconnected ...

danielballan:
Wait for v0.11.0? That ship has [sailed](https://github.com/NSLS-II/databroker/tree/v0.11.0). We originally planned to do this Friday, and we couldn't postpone deployments any longer.

It would make worried to deploy this without unit tests anyway. Previous attempts at addressing this use case have come and gone from the source code, and each time we accidentally broke it because we had no automated tests on it. (Or at least, people reported that it was broken, and we had no way to be sure!) Better to do it right, this time, in my opinion.

malitsky:
Sure. I plan to add unit tests by the end of May.

malitsky:
archiver.py and _core.py extended to support the new databroker configuration file including event sources. The following script and configuration file demonstrates this extension:

https://github.com/malitsky/nsls2/blob/master/areventsource/notebooks/archiver-reader.db.5.18.18.ipynb

https://github.com/malitsky/nsls2/blob/master/areventsource/notebooks/csx_with_archivers.yml

danielballan:
I noticed a couple merge commits ("Merge branch 'master' of https://github.com/NSLS-II/databroker") go by here. We follow [GitHub Flow](https://guides.github.com/introduction/flow/), wherein we _rebase_ on master rather than merging it in, which keeps the project's commit history simpler.

If you need help with that process, please don't hesitate to ask. As I have mentioned on here before, I received copious help with git surgery myself when I first contributed to a project on GitHub, and I'm more than happy to pay it forward. :- )

databroker/_core.py
```diff
@@ -1105,6 +1105,13 @@ def fetch_external(self, start, stop):
         return {k: func(start, stop) for
                 k, func in self.external_fetchers.items()}
 
+    @property
+    def ES(self):
```
danielballan: Any better ideas for how to name this? We have `event_sources`, a list of event sources, and we have this -- a dict mapping the event source _name_ to the event source. Not sold on `ES`. attn @NSLS-II/dama

databroker/_core.py
```diff
@@ -2156,8 +2167,17 @@ def from_config(cls, config, auto_register=True, name=None):
...
         mds = mds_cls(config['metadatastore']['config'])
         assets = assets_cls(config['assets']['config'])
+        # Instantiate event sources.
+        event_sources = []
+        ess = []
+        if 'event_sources' in config.keys():
+            ess = config['event_sources']
+        for es in ess:
+            es_cls = load_cls(es)
+            event_sources.append(es_cls._from_config(es['config']))
```
danielballan: I think this can just be `es_cls(es['config'])` and the `ArchiverEventSource.from_config` method can be removed. The reason that `Broker` has `from_config` is that it can be handed live `mds` and `reg` objects (via `__init__`) or configuration telling it how to go build them (via `from_config`). In this situation, I think it is fine (and obviously simpler) to only support the `__init__` path.

danielballan:
I see some test commits. This might be a good time to re-up my suggestion posted [above](https://github.com/NSLS-II/databroker/pull/359#pullrequestreview-116907450) and at various times previously. The solution in this PR is runnable without standing up a real archiver, which is good, but it would need to be maintained separately in the event of a change to Archiver. The standard "VCR" approach records actual request/response cycles from a real Archiver, so it is guaranteed to be correct to the byte, and it is easy to re-recoded if the Archiver is changed in future.

malitsky:
Great. I replaced the HTTP server with the VCR approach. The corresponding unit test works on all the databroker platforms.

danielballan:
That looks like a good start with VCR. I will take a more detailed look when I get the chance. In the meantime, it looks like there are some unaddressed comments above.

databroker/_core.py
tacaswell: `event_source_by_name`?

`sources_by_name`

`named_sources`

`named_es`

`es_by_name`

`names`

test-requirements.txt
```diff
@@ -2,3 +2,5 @@ coverage
 glueviz
 matplotlib
 pytest
+vcrpy
```
danielballan: Note that this list is alphabetized to make it easy to check whether something is present. (It's admittedly of little use in such as short list, but it's our system for all the projects.)

doc/source/archiver.rst
```diff
@@ -0,0 +1,61 @@
...
+-------------
+
+ArchiverEventSources is configured via :class:`Broker` by extending the databroker
+configuration file with the archivers entry. For example, the following
+cxs_with_archiver.yml file illustrates the configuration of
+two ArchiverEventSources, arch_csx and arch_acc:
+
+.. code-block:: yaml
+
+    archivers:
```
danielballan: The archiver shouldn't get its own special name in configuration. It is "just another event source."

doc/source/archiver.rst
```diff
@@ -0,0 +1,61 @@
...
+      module: databroker.eventsource.archiver
+      class: _ArchiverReader
```
danielballan: The config shouldn't need to refer to private classes. Whatever class is referenced here should be public.

databroker/eventsource/tests/conftest.py
```diff
@@ -0,0 +1,2 @@
+import pytest
```
danielballan: Is this accidental?

test-requirements.txt
```diff
@@ -2,3 +2,5 @@ coverage
...
+pathlib
```
tacaswell: Why do we need pathlib?

databroker/eventsource/eventsource.py
```diff
@@ -0,0 +1,64 @@
+
+from ..utils import ALL
+
+import sys
+import abc
+
+# Version 3.4+
+
+def define_base_30():
```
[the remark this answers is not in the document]
danielballan: Not as far as I know.

test-requirements.txt
malitsky: pathlib was added after updating the unit test to be consistent with NSLS-II/pyolog2/test_olog.py
 and in order to support the Python 2.7 test environment

malitsky:
The ArchiverEventSource commits are rebased. This PR deals with the clear-cut implementation task. Subsequently, AbstractEventSource and the ES method have been removed for being submitted with a new PR.

doc/source/api.rst
```diff
@@ -300,9 +300,10 @@ Configuration Utilities
 
 See also the Broker methods :meth:`Broker.from_config` and
 :meth:`Broker.named`.
-
+
```
danielballan: accidental whitespace

doc/source/archiver.rst
```diff
@@ -0,0 +1,66 @@
+.. currentmodule:: databroker
+
+
+**********************************
+Connection with Archiver Appliance
+**********************************
+
+The Archiver Appliance is an EPICS logging system developed in a collaboration of SLAC, BNL, and MSU for collecting and storing measurements from various control devices. Data retrieval is based on the client-server interface using HTTP requests. Large-scale accelerators and experimental facilities may maintain multiple archivers. Within the databroker application, each archiver is represented by an ArchiverEventSource that needs to be configured and registred with the Broker. The following sections highlight two major topics: ArchiverEventSource configuration and data retrival based on the databroker interface.
```
danielballan: Should be line-wrapped at 79 chars

databroker/eventsource/archiver.py
```diff
@@ -10,45 +10,113 @@
...
+
+        if 'timezone' in config:
+            timezone = config['timezone']
+        else:
+            timezone = 'US/Eastern'
         self.tz = pytz.timezone(timezone)
+
+        if 'pvs' in config:
+            pvs = config['pvs']
+        else :
```
danielballan: There is an extra space here.

databroker/eventsource/archiver.py
```diff
@@ -10,45 +10,113 @@
...
         self.url = url
         self.archiver_addr = self.url + "retrieval/data/getData.json"
+
+        if 'timezone' in config:
+            timezone = config['timezone']
+        else:
+            timezone = 'US/Eastern'
         self.tz = pytz.timezone(timezone)
+
+        if 'pvs' in config:
```
danielballan: Is there a compelling use case for making `pvs` optional? Should we just require it?

databroker/eventsource/archiver.py
```diff
@@ -10,45 +10,113 @@
...
+        self.pvs.update(dict(new_pvs))
+        return dict(self.pvs)
+
+    def get_pvs(self):
+        return dict(self.pvs)
+
     def insert(self, name, doc):
-        raise NotImplementedError()
+         """
+         Not supported, data archiving is managed via the EPICS Archiver Appliance Toolkit
```
danielballan: Should be line-wrapped at 79 chars

databroker/eventsource/tests/test_archiver.py
```diff
@@ -0,0 +1,53 @@
...
+    
+    hdr = get_header_for_archiver()
+    pvs = aes.stream_names_given_header(hdr)
+    df = aes.table_given_header(hdr, stream_name = pvs[0])
+    
+    values = df[pvs[0]].values.tolist()
+    assert len(values) == 61
+
+
+def test_pkg_resources():
```
danielballan: How is `pkg_resources` involved?

databroker/eventsource/tests/utils.py
```diff
@@ -0,0 +1,26 @@
+import time
+
+import six
+
+from databroker.eventsource.archiver import ArchiverEventSource
+
+arch_csx_url = 'http://xf23id-ca.cs.nsls2.local:17668'
```
danielballan: @tacaswell Opinions on hard-coding this into databroker?

databroker/eventsource/archiver.py
```diff
@@ -10,45 +10,113 @@
...
 
     def stream_names_given_header(self, header):
         # We actually don't use the header in this case.
-        return ['archiver_{}'.format(name) for name in self.pvs]
+        """
+        Return a list of user-defined PV names
+
+        Parameters
+        ----------
+        header: Header
```
danielballan: This docstring won't render properly in HTML without a space before the colon, like `header : Header`.

doc/source/archiver.rst
```diff
@@ -0,0 +1,66 @@
...
+* name: user-defined name of the Archiver Appliance archiver
+* url: address of the Archiver Appliance Retrieval server
+* pvs: dictionary mapping user-defined names to EPICS PVs
+
+A pvs dictionary can be extended via :class:`Broker` as:
+
+.. code-block:: python
+  
+   db = Broker.named('csx_with_archivers')
+   arch_csx  = db.ES['arch_csx']
```
danielballan: Is this still true? Can't find `ES` now...

danielballan:
@malitsky I notice that the VCR tests were run against a local beamline. It would be better to run them against some reproducible deployment, such as a VM or something that our collaborators at other facilities could stand up on their own. How much trouble would this be to do?

malitsky:
PR is updated to address the recent comments.

databroker/eventsource/archiver.py
malitsky: pvs is optional since their choice depends on data analysis and can be updated via the add_pvs method.

malitsky:
The EPICS Archiver Appliance data storage and retrieval interface are generic. Specifically, data are organized as a flat collection of PV datasets and the corresponding interface uses only the PV name and a time interval for reading time series of PV values. From this perspective, any beamline can be considered as a reference and generic use case. Furthermore, everything (including the url address) is captured in the VCR file that can be generated by the simulation program without the EPICS Archiver.

danielballan:
> [...] any beamline can be considered as a reference and generic use case. Furthermore, everything (including the url address) is captured in the VCR file that can be generated by the simulation program without the EPICS Archiver.

That's true, but it doesn't solve the whole problem. Imagine this realistic situation: A collaborator from the APS makes a change that involves sending a different request to the Archiver. Thus, the relevant VCR "recording(s)" need to be updated. That collaborator will run the tests against their own PV(s) using their own Archiver deployment. When they commit the results, all the VCR recordings will be changed to reflect the APS PV(s). Later, when someone at NSLS-II makes a change, all the recordings will be changed back. This thrashing will accumulate large diffs in the git history over time.

It could be avoided by capturing the recordings against a standard Archiver deployment that anyone can stand up using a bash script. We could include that script to make it easy for others to contribute. Could you try following the [Archiver Quickstart](https://slacmshankar.github.io/epicsarchiver_docs/quickstart.html) instructions to see how much trouble it would be a run a standard deployment?

The VCR layer still has value because you'll be able to run the tests without any Archiver deployment as long you aren't changing the requests. That will lower the barrier to entry for the vast majority of contributions, and it will make CI simpler/faster.

danielballan:
P.S. For what it's worth, we are taking this exactly same approach with http://github.com/NSLS-II/pyolog2 and have used a similar strategy with http://github.com/NSLS-II/caproto.

It takes some work to set up initially, but creating a low barrier to entry for subsequent contributors is a critical part of our strategy for making databroker a multi-facility project with many contributors.

malitsky:
VCR is a good idea, especially for testing this generic retrieval interface. Then, why would the APS people generate a new VCR file. As I said, everything is generic, trivial, and encapsulated in the corresponding VCR file, more precisely in one line of this file:

"uri": "http://xf23id-ca.cs.nsls2.local:17668/retrieval/data/getData.json?pv=XF%3A23ID-ID%7BBPM%7DVal%3APosXS-I&from=...

Should we replace our realistic-looking url and PV name with http://fake_url_name:17668 and fake-epics-name? Again, thanks to VCR, the unit test checks the integration of the generic Archiver output from one retrieval method which is the same for other PVs

danielballan:
The response that VCR captures will depend what data has been stored in that PV. I am proposing that we write a script that populates a freshly deployed archiver with a small amount of known data and then test against that. It won't be a "fake" URL; it will be `localhost` where VCR will capture the responses of an Archiver with known output.

malitsky:
In case you are not familiar with Archiver, here is a sequence of required actions:  (1) deploy a new Archiver three-tier system (like xf23id-ca ), (2) create a new IOC with simulated “known data” of some PV (like  XF:23ID-ID{BPM}Val:PosXS-I), (3) access the Archiver MGMT service for registering the corresponding PV, (4) connect the Archiver Engine with the new IOC for collecting and storing the PV time series into the Archiver backend, (5) connect to the Archiver Retrieval service for accessing the collected time series and (6) finally create the VCR file (including another url name). Will this file with the SAME INFORMATION provide any NEW VALUE for testing the Archiver Retrieval interface? Is this the topic that is preventing the PR from being merged?

databroker/eventsource/archiver.py
```diff
@@ -10,45 +10,113 @@
...
         self.pvs = pvs
+        
         self._descriptors = {}
 
+
+    @property
+    def name(self):
+        return self._name
+
+    def add_pvs(self, new_pvs):
```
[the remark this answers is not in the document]
tacaswell: There is something to be said for returning a copy so that the user can not accidentally mutate it someplace else in the code, but `es.pvs.update({...})` is _also_ very nice.  I am not sure of anywhere else we go to such lengths to protect the user from them selves (and in fact in the case of `read_attrs` we go to great lengths to fake returning a mutable thing!).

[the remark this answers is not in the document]
tacaswell: We also had a very nasty bug with the hkl library where we were doing

```python
config = HKLEngine.get_config()
config.set_stuff()
```

but the object returned was a copy (not a proxy) for the underlying c++ object so various constraints were not actually being set so the computations were wrong.  This API has echos of that.

databroker/eventsource/archiver.py
```diff
@@ -10,45 +10,119 @@
...
+        """
+        
+        return dict(self.pvs)
+
     def insert(self, name, doc):
-        raise NotImplementedError()
+         """
+         Not supported, data archiving is managed via the EPICS Archiver
+         """
+         raise TypeError("Not supported, data archiving is managed via the EPICS Archiver Appliance Toolkit")
```
tacaswell: Why change the exception type?

databroker/eventsource/archiver.py
```diff
@@ -10,45 +10,119 @@
...
+        ----------
+        header : Header
+            not used. fields are generated from pvs
+
+        Returns
+        -------
+        set : set of user-defined PV names
+        """
+               
+        return set(self.pvs)
```
tacaswell: why change the return type?

danielballan:
Niceties like what I am proposing are an important difference between developing code for a collaborative open source project and developing research code or code suitable for a technical demo. Creating a reproducible process for new contributors or contributors at other facilities is worth paying an up-front cost; it's the only way to be sustainable. That's the "new value" I am concerned with.

My original question was:

> I notice that the VCR tests were run against a local beamline. It would be better to run them against some reproducible deployment, such as a VM or something that our collaborators at other facilities could stand up on their own. How much trouble would this be to do?

You made the case that indeed it would be a significant amount of trouble:

> In case you are not familiar with Archiver, here is a sequence of required actions: (1) deploy a new Archiver three-tier system (like xf23id-ca ), (2) create a new IOC with simulated “known data” of some PV (like XF:23ID-ID{BPM}Val:PosXS-I), (3) access the Archiver MGMT service for registering the corresponding PV, (4) connect the Archiver Engine with the new IOC for collecting and storing the PV time series into the Archiver backend, (5) connect to the Archiver Retrieval service for accessing the collected time series and (6) finally create the VCR file (including another url name). Will this file with the SAME INFORMATION provide any NEW VALUE for testing the Archiver Retrieval interface? Is this the topic that is preventing the PR from being merged?

In case you didn't know, all-caps is generally read as "yelling" and is good to avoid in online discussions generally, including here.

In any case, if it's a lot of trouble to get something stood up, one could argue that that's all the _more_ reason to provide a clear and streamlined process for doing so, in the context of a collaborative open-source project. Could the effort you describe be captured in a VM image? Launching the Olog is similarly complex, and that's how they handle it.

We might also conclude that it's simply not worth the trouble and shouldn't block merging. I'm not yet personally convinced, but open to hearing others' arguments on this.

databroker/eventsource/archiver.py
tacaswell: The field name and stream names are identical?

databroker/eventsource/shim.py
tacaswell: This should probably be removed now?

databroker/eventsource/tests/utils.py
tacaswell: :-1:  It can we replace that with a dummy address?

doc/source/archiver.rst
```diff
@@ -0,0 +1,79 @@
...
+ - config:
+      name: 'arch_acc'
+      url: 'http://arcapp01.cs.nsls2.local:17668'
+      timezone: 'US/Eastern'
+      pvs:
+        pv2: 'UT:SB1-Cu:1{}T:Prmry-I'
+   module: 'databroker.eventsource.archiver'
+   class: 'ArchiverEventSource'
+ - config:
+     name: 'arch_csx'
```
tacaswell, line 40: this has no PVs?

tacaswell:
I agree with @danielballan that having a script / VM / what ever to generate reproducible synthetic data is the right thing to do.  From @malitsky 's comments it seems like we know how to do it and it is straight forward process (particularly if you use a caproto IOC).  However, I do not think that work should hold up this PR, I will create another issue for that.

@malitsky can you also add some more tests to push the coverage of the new code up?  codecov will tel you what lines are missing.

danielballan:
That works for me. We wanted to have this on the floor a long time ago, so I'm OK with merging this (once coverage is good and tests pass) as long as we don't lose momentum and track further progress on #374.

malitsky:
I am using all-caps to highlight my points, such as the BEAUTY of the standard VCR approach that perfectly addresses the scope of the current PR.

danielballan:
VCR is an important part of the solution, but only part of the solution. It streamlines the common case. I think "perfectly addresses the scope of this PR" is overstating the case.

databroker/eventsource/tests/utils.py
```diff
@@ -11,16 +13,98 @@
...
+    config = {'module': 'databroker.assets.mongo',
+              'class': 'Registry',
+              'config': {
+                  'host': 'localhost',
+                  'port': '27017',
+                  'database': 'fs'}}
+    return config
+
+def get_db_config_for_event_sources():
+	archiver_config =  get_config_for_archiver()
```
danielballan: Something must be off about your editor's configuration. There are couple 5-spaces indents, mixed with 4-spaces indents.

doc/source/archiver.rst
malitsky: Name (in shim.py) is used by BrokerES::event_sources_by_name. 

PVs (in archiver.rst) are optional since they can be added via API as demonstrated in this document.

malitsky:
PR is updated to address the recent comments

databroker/eventsource/archiver.py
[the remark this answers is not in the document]
danielballan: @malitsky Do you agree then that `add_pvs` and `get_pvs` should be removed?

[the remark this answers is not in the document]
malitsky: As already explained in the documentation’s comments, add_pvs and get_pvs provide a user with the optimal mechanism for extending the Databroker configuration file with application-specific pvs (note: beamline archivers maintain tens of thousands pvs):

https://github.com/malitsky/databroker/blob/master/doc/source/archiver.rst

Let me know when you plan to merge this PR.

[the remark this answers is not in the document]
malitsky: These methods follow a conventional approach for separating the class interface from its members. The optimal interface however is a trade-off among objective and subjective factors, so these methods per se can be removed.

malitsky:
PR is updated to address the recent comments

danielballan:
Tests failed due to server startup. Re-running that job to be sure we can get a fully-green build.

danielballan:
I was giving it one more read and found standing unresolved comments -- [example](https://github.com/NSLS-II/databroker/pull/359/files#r208736550)

databroker/eventsource/shim.py
malitsky: Name (in shim.py) is used by BrokerES::event_sources_by_name.

malitsky:
This comment was already addressed in the same conversation and in another place: “Name (in shim.py) is used by BrokerES::event_sources_by_name”. Nonetheless, the same reply has been added to this comment.

PR #360, opened 2018-04-30
Added DBEP for mutating documents

jrmlhermitte:
This is  beginning for the discussion of document mutation for documents stored by data broker.


doc/source/DBEP/DBEP07.rst
```diff
@@ -0,0 +1,60 @@
...
+   this information together on the way out.
+   #* Pros : This could likely be easily turned on or off, since the addition of
+     the changes is performed upon request.
+   #* Cons : The changes must be made as the documents are accessed. A new
+     function must be supported/debugged for accessing data.
+
+
+Resources/similar work
+======================
+
```
tacaswell, line 57: This should to point to how we are currently doing this with `Resource` documents.

doc/source/DBEP/DBEP07.rst
```diff
@@ -0,0 +1,60 @@
...
+
+
+Proposed Implementations
+====================
+There are three ways that have been discussed for handling this.
+Which one should be used has not been decided on as of now.
+
+
+#. Copy the original data into a backup data store. Create a new start document
```
tacaswell, line 34: Changing the UID is :-1: .  We have promised that they will always be able to get the data back if they give us that string.

doc/source/DBEP/DBEP07.rst
```diff
@@ -0,0 +1,60 @@
...
+   with a new uid with the updated metadata. Lutz didn't like this because this
+   isn't backwards compatible with his previous measurements.
+   #* Pros: Data is never mutated, less risk for data corruption
+   #* Cons: This is not very backwards compatible for beamline scientists. I.e.
+     retrieval of data by uid must be updated with the newest uid used.
+#. Modify the existing start document, and save the changes made in a separate
```
tacaswell, line 40: :+1:

doc/source/DBEP/DBEP07.rst
```diff
@@ -0,0 +1,60 @@
...
+   changes database.
+   #* Pros : The user is unaffected by the changes
+   #* Cons : Once the change is made, it is a little difficult to view the
+     documents without the change.
+#. Create a separate database (mongo collection, sqlite database, spreadsheet,
```
tacaswell, line 45: We are going to want to get the corrected version far more often than we are going to want the original so it makes more sense to store the store the final state and compute the diff when we want it than to store the original + diffs and compute the current everytime we access it.

Interestingly, this is the ways that git (store full version, compute diffs) and svn (store diffs, compute full version) work.

doc/source/DBEP/DBEP07.rst
```diff
@@ -0,0 +1,60 @@
...
+
+
+Branches and Pull requests
+==========================
+
+
+Abstract
+========
+The motivation of this DBEP is the handling of document mutation.
+There is a need for post-acquisition modification of metadata.
```
tacaswell, line 22: Which documents will we allow users to modify?

doc/source/DBEP/DBEP07.rst
tacaswell: and now that they are in the same code base we should look into making them share a code base (independent of how we implement this, all of the documents should be handled in the same way.

jrmlhermitte: good idea! I added it in before I forget! (I looked found it earlier but forgot to note it down)

doc/source/DBEP/DBEP07.rst
jrmlhermitte: Good point. Only permitted keys in the start and stop documents to begin with.
However, one could imagine mutating the descriptor configuration for ex: the wrong wavelength was given

doc/source/DBEP/DBEP07.rst
jrmlhermitte: I agree with this. I will move the two points to "other implementations" and keep the second point (which is basically the implementation performed for resources)

doc/source/DBEP/DBEP07.rst
tacaswell: The things in the configuration are usually read from epics.  How do you draw the line between 'correcting' and 'falsifying' data?

CJ-Wright: User defined masks could be generated and go into the analysisDB. The analysisDB is mostly disposable and has a many to many relationship with the data. If a new mask is needed we can just make a new one.

Wavelengths could also be a generated parameter, since usually some amount of analysis is needed to turn a motor position/xtal scattering into a wavelength.

danielballan:
Now that _insert_ is moving out of scope for databroker, this discussion continues in https://github.com/bluesky/suitcase-mongo/issues/25.

PR #361, opened 2018-05-09
ENH: add hook to also fetch external data to go with a header

tacaswell:
- can register a (NAME, function) that takes a start and stop document and
   returns an object
 - payload is available on h.ext.NAME


tacaswell:
This is ready to go and passes against other master branches.

PR #362, opened 2018-05-14
DOC: Release notes for v0.11.0

danielballan:


stuartcampbell:
Still not happy with the name `ext` 😄

danielballan:
Just noting that I consider this an experimental feature, pre-1.0, and I reserve the right to argue (with @stuartcampbell in my corner, looks like) that `external` would be clearer. :- )

PR #363, opened 2018-05-17
CLN: remove print namespace in _core

mrakitin:
Previously observed by @tacaswell and @jrmlhermitte at ISS, found today at TES.


danielballan:
Bah. I merged it. My fault too. Sorry, all.

PR #372, opened 2018-07-23
FIX: Do not assume Descriptors have a 'name' field.

danielballan:
The 'name' field is not required by the schema, and we cannot assume that it is
present. It happens that the bluesky RunEngine always includes this field now,
but older documents do not have it, and it is still optional from the point of
view of the schema.

### DIFF
--- a/databroker/_core.py
+++ b/databroker/_core.py
@@ -290,7 +290,7 @@ class Header(object):
         for d in sorted(self.descriptors, key=lambda d: d['time']):
             config = d['configuration'].get(device_name)
             if config:
-                result[d['name']].append(config['data'])
+                result[d.get('name')].append(config['data'])
         return dict(result)  # strip off defaultdict behavior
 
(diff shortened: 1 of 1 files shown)


databroker/_core.py
```diff
@@ -1657,7 +1657,7 @@ def get_table(self,
 
                 # get the first descriptor for this event stream
                 desc = next((d for d in h.descriptors
-                             if d['name'] == stream_name),
+                             if d.get('name') == stream_name),
```
jrmlhermitte, line 1660: just a little devil's advocate, what if somehow `stream_name = None` was passed? Shouldn't be possible, but might be safer
```py3
if d.get('name') == stream_name and d.get("name") is not None
```
EDIT: or maybe some sentinel: `d.get('name', NONE)`

danielballan: Stream names, if present, must be strings.

PR #375, opened 2018-10-12
add a file handler to be used with AD hdf5 plugin in Single mode

awalter-bnl:
## Description:
This PR adds a new filestore handler that should be used with the new filestore mixin defined in the PR [here](https://github.com/NSLS-II/ophyd/pull/623)

## Motivation:
This was motivated by a custom plugin which has an array with one dimension determined only after running the plugin The result is that each `trigger` then creates a different sized array to the previous one, and so the `Streaming` mode which sets the array size prior to any `trigger`  did not correctly save the data.

## Testing:
This has been tested in the field as submitted here (NSLS-II, SIX beamline, rixscam detector, XIP plugin) and is working there.

### DIFF
--- a/databroker/assets/handlers.py
+++ b/databroker/assets/handlers.py
@@ -281,9 +281,8 @@ class AreaDetector_HDF5SingleHandler_DataFrame(AreaDetector_HDF5SingleHandler):
                  column_names=None, frame_per_point=1):
         # I have included defaults for `key` and 'column_names' for back
         # compatibility with existing files at SIX.
-        super(AreaDetector_HDF5SingleHandler_DataFrame, self).__init__(
-              fpath=fpath, template=template, filename=filename, key=key,
-              frame_per_point=frame_per_point)
+        super().__init__(fpath=fpath, template=template, filename=filename,
+                         key=key, frame_per_point=frame_per_point)
         self._path = os.path.join(fpath, '')
         self._fpp = frame_per_point
         self._template = template
--- a/databroker/assets/handlers.py
+++ b/databroker/assets/handlers.py
@@ -2,6 +2,7 @@ from __future__ import (absolute_import, division, print_function,
                         unicode_literals)
 import logging
 import numpy as np
+import pandas as pd
 import os.path
 from .handlers_base import HandlerBase
 from .readers.spe import PrincetonSPEFile
@@ -202,8 +203,10 @@ class AreaDetectorHDF5Handler(HDF5DatasetSliceHandler):
             frame_per_point=frame_per_point)
 
 
-class HDF5SingleHandler(HandlerBase):
-    '''Handler for hdf5 data stored 1 image per file.
+class AreaDetector_HDF5SingleHandler(HandlerBase):
+    '''Handler for hdf5 data stored 1 image per file and returned as a
+    numpy.array.
+
     Parameters
     ----------
     fpath : string
@@ -214,11 +217,16 @@ class HDF5SingleHandler(HandlerBase):
         filename
     key : string
         the 'path' inside the file to the data set.
+    column_names : list[str]
+        The column names of the table
     frame_per_point : float
         the number of frames per point.
     '''
+    specs = {'AD_HDF5_SINGLE'} | HandlerBase.specs
 
-    def __init__(self, fpath, template, filename, key, frame_per_point=1):
+    def __init__(self, fpath, template, filename, key='/entry/data/data',
+                 frame_per_point=1):
+        # I have included defaults for `key` for back compatibility.
         self._path = os.path.join(fpath, '')
         self._fpp = frame_per_point
         self._template = template
@@ -232,12 +240,12 @@ class HDF5SingleHandler(HandlerBase):
             yield self._template % (self._path, self._filename, j)
 
     def __call__(self, point_number):
-        import h5py
         ret = []
+        import h5py
         for fn in self._fnames_for_point(point_number):
-            f = h5py.File(fn, 'r')
-            data = f[self._key]
-            ret.append(data)
+            with h5py.File(fn, 'r') as f:
+                array = np.array(f[self._key][:])
+            ret.append(array)
         return ret
 
     def get_file_list(self, datum_kwargs):
@@ -247,8 +255,11 @@ class HDF5SingleHandler(HandlerBase):
         return ret
 
 
-class AreaDetectorHDF5SingleHandler(HDF5SingleHandler):
-    '''Handler for hdf5 data stored 1 image per file by areadetector
+class AreaDetector_HDF5SingleHandler_DataFrame(AreaDetector_HDF5SingleHandler):
+    '''Handler for hdf5 data stored 1 image per file and returned as a
+    Pandas.DataFrame.
+    This will work with all hdf5 files that are a mxn arrays and the data is
+    'table like' where m is the number of columns and n is the number of rows.
     Parameters
     ----------
     fpath : string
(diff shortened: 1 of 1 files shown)


awalter-bnl:
The Travis error here is a caproto IOC 'server startup timed out` error on the nightly build, I don't think this is related to the current changes.

awalter-bnl:
Based on some changes made in a version of `NSLS-II-SIX.sixtools` which has been merged and tested. I have also added 2 versions, one which returns a `numpy.array` (for image like data) and one which returns a `pandas.DataFrame` (for tabular style data).

databroker/assets/handlers.py
```diff
@@ -203,6 +203,105 @@ def __init__(self, filename, frame_per_point=1):
...
+
+    def _fnames_for_point(self, point_number):
+        start = int(point_number * self._fpp)
+        stop = int((point_number + 1) * self._fpp)
+        for j in range(start, stop):
+            yield self._template % (self._path, self._filename, j)
+
+    def __call__(self, point_number):
+        ret = []
+        import h5py
```
mrakitin, line 244: Maybe import is once in the beginning of the file, so that it won't be necessary to do it twice.

databroker/assets/handlers.py
```diff
@@ -203,6 +203,105 @@ def __init__(self, filename, frame_per_point=1):
...
+        # I have included defaults for `key` for back compatibility.
+        self._path = os.path.join(fpath, '')
+        self._fpp = frame_per_point
+        self._template = template
+        self._filename = filename
+        self._key = key
+
+    def _fnames_for_point(self, point_number):
+        start = int(point_number * self._fpp)
+        stop = int((point_number + 1) * self._fpp)
```
mrakitin: For readability, this can be replaced with `stop = start + self._fpp` (maybe with `int()`).

databroker/assets/handlers.py
```diff
@@ -203,6 +203,105 @@ def __init__(self, filename, frame_per_point=1):
...
+        for j in range(start, stop):
+            yield self._template % (self._path, self._filename, j)
```
mrakitin: Why not to use `self._template.format(...)`? Same effect, newer syntax.

databroker/assets/handlers.py
```diff
@@ -203,6 +203,105 @@ def __init__(self, filename, frame_per_point=1):
...
+    specs = {'AD_HDF5_SINGLE_DATAFRAME'} | HandlerBase.specs
+
+    def __init__(self, fpath, template, filename, key='/entry/data/data',
+                 column_names=None, frame_per_point=1):
+        # I have included defaults for `key` and 'column_names' for back
+        # compatibility with existing files at SIX.
+        super(AreaDetector_HDF5SingleHandler_DataFrame, self).__init__(
+              fpath=fpath, template=template, filename=filename, key=key,
+              frame_per_point=frame_per_point)
+        self._path = os.path.join(fpath, '')
```
mrakitin, line 286: Not sure why we need it here and above. Is it to get the absolute path? `os.path.abspath()` can be used then.

databroker/assets/handlers.py
```diff
@@ -203,6 +203,105 @@ def __init__(self, filename, frame_per_point=1):
...
+        self._fpp = frame_per_point
+        self._template = template
+        self._filename = filename
+        self._key = key
+        self._column_names = column_names
+
+    def __call__(self, point_number):
+        ret = []
+        import h5py
```
mrakitin, line 295: See my comment about importing above.

databroker/assets/handlers.py
```diff
@@ -203,6 +203,105 @@ def __init__(self, filename, frame_per_point=1):
...
+    frame_per_point : float
+        the number of frames per point.
+    '''
+    specs = {'AD_HDF5_SINGLE_DATAFRAME'} | HandlerBase.specs
+
+    def __init__(self, fpath, template, filename, key='/entry/data/data',
+                 column_names=None, frame_per_point=1):
+        # I have included defaults for `key` and 'column_names' for back
+        # compatibility with existing files at SIX.
+        super(AreaDetector_HDF5SingleHandler_DataFrame, self).__init__(
```
mrakitin: I don't understand how it works here, can you explain why `AreaDetector_HDF5SingleHandler_DataFrame` is used within itself, and then the `self` is added?

databroker/assets/handlers.py
awalter-bnl: done

databroker/assets/handlers.py
awalter-bnl: I am following the style from all the previous handlers which is to import the filehandler module just before it is first used, for example see [here](https://github.com/NSLS-II/databroker/blob/f4fbfd352d6ab6ffbd494e9e8d948764423abee4/databroker/assets/handlers.py#L171).

If we want to change this I am happy to do it.

databroker/assets/handlers.py
awalter-bnl: It appears they are not equivalent:
` print (self._template % (self._path, self._filename, 1))`
` print (self._template.format(self._path, self._filename, 1))`
returns
`/XF02ID1/RIXSCAM/DATA/2018/10/10/4a45e3b9-0ac2-4196-aec0_000001.h5`
`%s%s_%6.6d.h5`

databroker/assets/handlers.py
awalter-bnl: see response above about importing

databroker/assets/handlers.py
awalter-bnl: On linux the difference between the two is:
`fpath = '/XF02ID1/RIXSCAM/DATA/2018/10/10'`
`self._path = '/XF02ID1/RIXSCAM/DATA/2018/10/10/'`
The goal of this line is to add a directory seperator to the end of fpath. This ensures that is done OS independently.

databroker/assets/handlers.py
awalter-bnl: I was following the structure [here](https://github.com/NSLS-II/databroker/blob/f4fbfd352d6ab6ffbd494e9e8d948764423abee4/databroker/assets/handlers.py#L201). After testing I found that it was not necessary to add that so it has been removed.

awalter-bnl:
@mrakitin thanks for the prompt feedback, please see responses above.

awalter-bnl:
The failure of the checks is in the 3.6 and nightly builds only. 
The error for the 3.6 build is:
RuntimeError: Non Expected exception in `/home/travis/build/NSLS-II/databroker/doc/source/assets.rst` line 77
THe error for the nightly build is in databroker/tests/test_broker.py and is:
 Exception: Server startup timed out.

I don't think that either of these are related to the current changes, let me know if they are and I will investigate further.

databroker/assets/handlers.py
mrakitin: I see, thanks for the reference. This importing within a function makes it harder to track all external dependencies, so I'd vote for moving it to the top of the file.

databroker/assets/handlers.py
mrakitin: I see, thanks for the explanation. That seems to be the easiest way to get the separator at the end.

databroker/assets/handlers.py
danielballan: Right, AD using traditional format tokens (`%s`, `%d`, ...) and Python's `.format` method only fills in new-style token like `{}`, `{<int>}`, or `{<keyword>}`.

databroker/assets/handlers.py
danielballan: I think this was our half-baked attempt to isolate dependencies so that, as `handlers.py` grows, we don't accumulate a ton of required dependency in `databroker`. The "proper" fix, in my view, is to follow the suitcase/intake pattern and create many small packages, as we have started to do with `eiger-io`.

I vote to follow @mrakitin here.

danielballan:
Let's try to get this across the finish line.

databroker/assets/handlers.py
mrakitin: I like the idea of using separate packages for specific handlers in the future.

danielballan:
Bump

awalter-bnl:
This is ready to go in based on my last comment some time ago

danielballan:
The blocking issue is review.

awalter-bnl:
I am guessing the fact that this has received no feedback even after @danielballan bump that it is no longer required. It was mentioned in a meeting that it will be superseded by the upcoming changes to databroker anyway. Therefore I am closing it out.

awalter-bnl:
re-opening at the request of @danielballan

PR #378, opened 2018-12-05
publish event documents in time order

### LINKED ISSUE #365: hdr.documents doesn't return the documents in their timestamp order
It seems that `hdr.documents()` doesn't return the documents in their timestamp order? Is this correct? Would it be possible for the databroker to report the documents in the same order they came out of the run engine?

CJ-Wright:
Closes #365


CJ-Wright:
@NSLS-II/databroker ready for review!

tacaswell:
This code is included in #379

PR #379, opened 2018-12-13
report resources and datums

CJ-Wright:


databroker/tests/test_broker.py
```diff
@@ -1262,3 +1263,62 @@ def test_monitoring(db, RE, hw):
     sd.monitors.append(hw.rand)
     RE(bp.count([hw.det], 5, delay=1))
     assert len(db[-1].table('rand_monitor')) > 1
+
+
+@py3
```
tacaswell: does this really require py3?

databroker/tests/test_broker.py
```diff
@@ -1262,3 +1263,62 @@ def test_monitoring(db, RE, hw):
...
+
+    names = set()
+    for (n1, d1), (n2, d2) in zip(db[uid].documents(), L):
+        print(n1)
+        names.add(n1)
+        assert n1 == n2
+        # It seems that some of the documents don't have key parity
+        if n1 == 'resource':
+            d1.pop('id')
+            d2.pop('run_start')
```
tacaswell: if 'run_start' is not coming back out of the databroker than is a bug someplace else..

databroker/assets/core.py
```diff
@@ -37,7 +37,7 @@ def _get_datum_from_datum_id(col, datum_id, datum_cache, logger):
     try:
         datum = datum_cache[datum_id]
     except KeyError:
-        keys = ['datum_kwargs', 'resource']
+        keys = ['datum_kwargs', 'resource', 'datum_id']
```
tacaswell: I assume this is so that you can get the full datum out later?

databroker/eventsource/shim.py
```diff
@@ -5,18 +5,52 @@
...
+    val : tuple
+        The next (name, dict) pair in time order
+
+    """
+    iters = [iter(g) for g in gens]
+    heap = []
+
+    def safe_next(indx):
+        try:
+            val = next(iters[indx])
```
tacaswell: could this be `val = next(gen)`?  I guess you are using the index to break the tie when two things have _exactly_ the same time?

databroker/eventsource/shim.py
```diff
@@ -5,18 +5,52 @@
...
-
 logger = logging.getLogger(__name__)
 
 
+def interlace_gens(*gens):
+    """Take generators and interlace their results by timestamp
+
+    Parameters
+    ----------
+    gens : generators
```
tacaswell, line 26: These docstrings are wrong, the code expects just the dictionaries.

databroker/assets/core.py
CJ-Wright: No this was so we have parity between what the `RunEngine` puts out and what the `Broker` puts out. On line https://github.com/NSLS-II/databroker/pull/379/files#diff-153ae1bc82a20d75fad0da96762b32d5R47 the only things in the datum are in those keys.

databroker/eventsource/shim.py
CJ-Wright: No I think this takes in the generators, unless `self.mds.get_events_generator(d)` reports dictionaries.

databroker/eventsource/shim.py
CJ-Wright: What is `gen`?

tacaswell: `gen` should be `iter_gen` (then things you keep in `iters`).

databroker/eventsource/shim.py
tacaswell: I miss-clicked, this should be on the line below.

databroker/assets/core.py
tacaswell: That is what I meant.  Most of this code was written assuming that datum/resource were not going to make it out to the world.  The reason the `datum_id` was not in this originally is that if you were asking for the resource and datum_kwargs, you already had the datum_id so why store it again.

Might be worth looking at just using `edoc` directly (and popping off `_id`) rather than using `keys` to construct a new dictionary.

databroker/eventsource/shim.py
```diff
@@ -204,9 +240,9 @@ class defined in this module). The default name is
             # if localizing to 'local' time
             if localize_times:
                 times = (times
-                         .dt.tz_localize('UTC')     # first make tz aware
-                         .dt.tz_convert(timezone)   # convert to 'local'
-                         .dt.tz_localize(None)      # make naive again
+                         .dt.tz_localize('UTC')  # first make tz aware
```
tacaswell, line 243: I think the even comments makes this clearer but meh.

databroker/eventsource/shim.py
CJ-Wright: the `indx` is in there so we don't need to store the entire generators inside the heap. The `indx` is also returned in the `heappop`. This allows us to draw from a generator until it finally falls behind another entry.

databroker/tests/test_broker.py
CJ-Wright: ~I thought RE was only py3k~

databroker/tests/test_broker.py
CJ-Wright: On my local sqlite and hdf5 don't have a `run_start` with their resources.

databroker/eventsource/shim.py
tacaswell: Isn't it an object either way?

```python
heap = []
def safe_next(itr, indx):
    try:
        val = next(itr)
    except StopIteration:
        return
    heapq.heappush(heap, (val['time'], indx, val, itr))

for j, itr in enumerate(iters):
    safe_next(itr, j)
while heap:
    _, j, val, itr = heapq.heappop(heap)
   yield val
   safe_next(itr, j)

```

would also work.

tacaswell: where here the index is serving as the tie-breaker.

databroker/tests/test_broker.py
tacaswell: That is a bug in those asset registries, does it work with mongo?

CJ-Wright: It does work with mongo (on my local)

CJ-Wright: example: https://travis-ci.org/NSLS-II/databroker/jobs/467728070#L1943

CJ-Wright: It seems that the sqlite asset registry doesn't have a column for `start_doc` (and hdf5 uses sqlite)

danielballan:
@CJ-Wright I happened to need this doing some live-coding at a beamline today, and I was so happy to find your PR ready to use. It worked for me interactively. Will take a close look tomorrow.

databroker/tests/test_broker.py
danielballan: @tacaswell and I are both on board with changing the schema for this. Does that work for you?

CJ-Wright: Maybe, I don't know what that means in concrete terms, though.

tacaswell:
I was suggesting to put the number in the table name.  When we open up a sqlite file first look for `resource_v2` and failing that look for `resource`.

CJ-Wright:
Ah ok (sorry about the names, I had committed that before your comment)

databroker/tests/test_broker.py
```diff
@@ -1262,3 +1263,59 @@ def test_monitoring(db, RE, hw):
...
+        db.reg.register_handler(spec, NumpySeqHandler)
+    L = []
+    RE.subscribe(db.insert)
+    RE.subscribe(lambda *x: L.append(x))
+
+    uid, = RE(count([hw.img], num=7, delay=0.1))
+
+    names = set()
+    for (n1, d1), (n2, d2) in zip(db[uid].documents(), L):
+        print(n1)
```
CJ-Wright:
```suggestion
```

databroker/tests/test_broker.py
```diff
@@ -1262,3 +1263,58 @@ def test_monitoring(db, RE, hw):
...
+@py3
+def test_order(db, RE, hw):
+    from ophyd import sim
+    RE.subscribe(db.insert)
+    d = sim.SynPeriodicSignal(name='d', period=.5)
+    uid, = RE(monitor_during_wrapper(count([hw.det], num=7, delay=0.1), [d]))
+
+    t0 = None
+    for name, doc in db[uid].documents():
+        print(name, doc['time'])
```
CJ-Wright:
```suggestion
```

databroker/assets/sqlite.py
```diff
@@ -104,7 +113,8 @@ def reconnect(self):
                 c.execute(CREATE_DATUMS_TABLE)
                 c.execute(CREATE_RESOURCE_UPDATES_TABLE)
         else:
-            EXPECTED_TABLES = ['Resources', 'Datums', 'ResourceUpdates']
+            EXPECTED_TABLES = ['Resources_{}'.format(RESOURCE_VERSION),
```
tacaswell: This will fail to read a non-versioned table.

databroker/assets/sqlite.py
```diff
@@ -33,26 +37,31 @@
...
+
+SELECT_RESOURCE = "SELECT * FROM Resources_{} WHERE uid=?;".format(
+    RESOURCE_VERSION)
+OLD_SELECT_RESOURCE = "SELECT * FROM Resources WHERE uid=?;"
 SELECT_DATUM_BY_UID = "SELECT * FROM Datums WHERE datum_id=?;"
 SELECT_DATUM_BY_RESOURCE = "SELECT * FROM Datums WHERE resource=?;"
+
 UPDATE_RESOURCE = """
-UPDATE Resources
+UPDATE Resources_{}
```
tacaswell, line 62: this only works on the new files, not the old.

CJ-Wright:
failing on a server timeout

danielballan:
This is some stellar work, @CJ-Wright. Thank you.

PR #380, opened 2018-12-14
allow for reversed databroker searches

CJ-Wright:
This will be helpful when trying to rerun an entire series of experiments
in order.

### DIFF
--- a/databroker/headersource/core.py
+++ b/databroker/headersource/core.py
@@ -777,7 +777,9 @@ def bulk_insert_events(event_col, descriptor, events, validate):
 
 # DATABASE RETRIEVAL ##########################################################
 
-def find_run_starts(run_start_col, run_start_cache, tz, **kwargs):
+def find_run_starts(run_start_col, run_start_cache, tz,
+                    order='descending',
+                    **kwargs):
     """Given search criteria, locate RunStart Documents.
 
     Parameters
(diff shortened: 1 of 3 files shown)


tacaswell:
This effectively squats a kwarg that is currently usable for meta-data (which we happen to be testing!).

Defer to @danielballan , but I am mostly :-1: on adding a new kwarg to `__call__`.

It is probably better to put reversing logic on the `Results` object.

CJ-Wright:
The results object might be too late, at that point the query has already been issued so we can't ask the database to return the results in reversed order. If we do this on the results we'll need to exhaust the generator first which could be rather expensive.

danielballan:
You are right that doing this on the Results object is too late. However, I agree that we ought not to shadow “order” as a possible field in a RunStart document that people want to search on. Granted, you are only following _our example_ here — we squatted “since” and “until” — but I now believe we made a mistake when we chose to do that.

I believe the correct way forward here is sketched out in what we have recently started calling the  “databroker 2.0 API”:

db.search(query_dict)

as opposed to

db(**query_dict)

That is: by accepting the query as a dict rather than as unpacked keyword arguments, we allow room for extra keyword arguments to search() like “order” and “limit” without ambiguity/collision with any such terms in the query itself.

I think it’s best to leave __call__() as it is and do this in search(). My work on the new API is in this repo. https://github.com/danielballan/intake-bluesky I will make it a priority to discuss this at length after the break. It is in the exploratory stage.

danielballan:
Note that this could potentially provide a path for another feature you have been (sensibly!) lobbying for:

db.search({}, descriptor_query={...})

danielballan:
Closing in favor of the approach I outlined above. #426 will track that work.

PR #385, opened 2019-01-08
support old resources

CJ-Wright:


databroker/assets/sqlite.py
```diff
@@ -255,14 +255,18 @@ def find_one(self, query):
         # it up in an older one.
         for select in [SELECT_RESOURCE, OLD_SELECT_RESOURCE]:
             with cursor(self._conn) as c:
-                c.execute(select, (query['uid'],))
-                raw = c.fetchone()
+                raw = None
+                try:
+                    c.execute(select, (query['uid'],))
```
tacaswell, line 260: Can you explain this change a bit more?

CJ-Wright: It seems that if this runs on an old databroker it raises an exception rather than just returning `None` as expected. This now runs a try except so it can cycle through all the table schemas until it finds the right one or outright fails.

danielballan: Just to be sure we understand the theory of operation, what is the text of the exception that gets raised?

databroker/assets/sqlite.py
```diff
@@ -255,14 +255,18 @@ def find_one(self, query):
...
+                    raw = c.fetchone()
+                except sqlite3.OperationalError:
+                    pass
```
danielballan, line 263: Could this be `continue`?

danielballan: ...and then
```
else:
    break
```

I think that would make this easier to follow, rather than using the `None` as a proxy for "Did it raise?"

databroker/assets/sqlite.py
CJ-Wright: `sqlite3.OperationalError: no such table: Resources_v2`

databroker/assets/sqlite.py
CJ-Wright: Maybe, are we expecting a `None`? It seems that the `if raw is None` is original to the code. See: https://github.com/NSLS-II/databroker/blame/2ff5d8a2cae8c76651b3924dc7e8b2b5ec374072/databroker/assets/sqlite.py

danielballan:
I don't understand this yet, and I think it's probably not worth anyone's time at this point as we're deprecating sqlite support in favor of formats with more first class support for objects (msgpack, JSON).

PR #386, opened 2019-01-17
Support round trip of resource

CJ-Wright:


CJ-Wright:
@NSLS-II/dama thoughts?

danielballan:
attn @tacaswell

databroker/assets/base_registry.py
```diff
@@ -588,6 +588,7 @@ def bulk_ingest_datum(self, resource_uid, datum_doc_iterable,
     # ## OLD API
     def insert_resource(self, spec, resource_path, resource_kwargs, root=None,
                         path_semantics='posix', uid=None, run_start=None,
+                        id=None,
```
mrakitin: Won't it collide with Python's built-in [`id`](https://docs.python.org/3/library/functions.html#id)?

Second question: is it used at all in this PR or anywhere else in the code?

Also, formatting suggestion: this line can be merged with the next line to save space and make it look nicer.

databroker/assets/core.py
```diff
@@ -168,6 +168,7 @@ class _PrivateException(Exception):
 def insert_resource(col, spec, resource_path, resource_kwargs,
                     known_spec, root, path_semantics='posix', uid=None,
                     run_start=None,
+                    id=None,
```
mrakitin: Maybe worth to combine it with the previous line?

mrakitin:
Also, rebasing on master would make sense to resolve the conflict.

databroker/assets/base_registry.py
CJ-Wright: It should only collide within the function, where is isn't used.
This is used so we can pass resources in without causing issues, since resources have an id field.

databroker/assets/core.py
```diff
@@ -168,7 +168,38 @@ class _PrivateException(Exception):
...
                     ignore_duplicate_error=False, duplicate_exc=None):
+    """Insert resource into a databroker
```
mrakitin: To make it a complete sentence:
```suggestion
    """Insert resource into a databroker.
```

databroker/assets/core.py
```diff
@@ -168,7 +168,38 @@ class _PrivateException(Exception):
...
+
+    Parameters
+    ----------
+    col :
```
mrakitin: Do you mind to specify types here as well?

databroker/assets/core.py
```diff
@@ -168,7 +168,38 @@ class _PrivateException(Exception):
...
+    resource_path : str
+        The path to the resource files
+    resource_kwargs : dict
+        The kwargs for the resource
+    known_spec : set
+        The known specs
+    root : str
+        The root of the file path
+    path_semantics : str, optional
+        The name of the path semantics, eg ``posix`` for linux systems
```
mrakitin:
```suggestion
        The name of the path semantics, e.g. ``posix`` for Linux systems
```

databroker/assets/base_registry.py
mrakitin: @CJ-Wright, thanks for the explanation here and in person. Do you also mind to add the docstring here as well so that the both participating parties have the documentation? Thanks.

databroker/assets/core.py
mrakitin: Do you mind to reformat the signature, so it's more compact, as suggested above or even better?

databroker/assets/core.py
danielballan: `pymongo.Collection`

databroker/assets/core.py
mrakitin: Addressed.

mrakitin:
Local tests pass, the problem with Travis is that there are timeout errors for some mongo tests. Decided to merge after discussion with @danielballan and @awalter-bnl.

danielballan:
It's worth clarifying that the timeout errors were all related to server startup, which is a known issue with the server test fixture, not a problem with "mongo tests" at large.

PR #392, opened 2019-02-21
Refactor databroker on top of intake.

danielballan:
This work is just begun, but I am opening a WIP PR so it can proceed publicly.

As communicated previously in various conversations and on the #databroker
Slack channel, the plan is:

- Refactor databroker to use intake as much as possible with little or no API
  breakage. Provide a second API that is basically just intake, which users can
  explore. This second API should not be considered stable.
- Tag 1.0.
- Once the new, intake-like API is stable, tag 2.0, and start guiding users to that API.
- Profit.


danielballan:
Rebased. This branch has gone a bit stale and some new failures have crept in due to changes in intake-bluesky. Current status:

```
28 failed, 519 passed, 97 skipped, 2 xfailed, 25 warnings
```

databroker/tests/test_urls.py
```diff
@@ -10,14 +11,18 @@
                           ('run_stop', 'find_run_stops'),
                           ('event', 'find_events'),
                           ('event_descriptor', 'find_descriptors')])
+
+@pytest.mark.skipif(os.environ.get('INCLUDE_V0_SERVICE_TESTS') != '1',
```
tacaswell: You can put this at the module level so just 

```python
pytest.mark.skipif(...)
```

which will skip the whole module.

databroker/v1.py
```diff
@@ -1608,38 +1608,58 @@ def _get_mongo_client(host, port):
     return client
 
 
-class _GetDocumentsRouter(event_model.DocumentRouter):
-    "This is used by Broker.get_documents."
+class _GetDocumentsRouter:
```
tacaswell, line 1611: promote this to event_model?

danielballan: Good idea. https://github.com/bluesky/event-model/issues/106

databroker/tests/test_urls.py
gwbischof: updated

tacaswell: For future referenece, it is 

```python
pytestmark = pytest.mark.skipif(...)
```

PR #394, opened 2019-03-11
DOC: release notes for v0.12.2

mrakitin:


danielballan:
Note for posterity: more server stratup failures, but the sphinx build passed.

PR #408, opened 2019-05-22
fixes #400

### LINKED ISSUE #400: YAML deprecation warning triggered by using load() without specifying the loader
Using databroker to load a YAML file will trigger the following warning:

```
databroker/_core.py:962: YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.
  return yaml.load(f)
```

the easiest fix would probably be changing `return yaml.load(f)` to `return yaml.load(f, Loader=yaml.FullLoader)`

prjemian:


danielballan:
The CI failures on this PR are due to general bit rot on the `master` branch, not this change. Merged. Thank you!

PR #410, opened 2019-06-04
DOC: fix db.process example

tacaswell:
process is a method on Broker, not a top-level function


mrakitin:
Do we know why the tests are failing?

danielballan:
Hmm, `pip` says it cannot find `event-model>=1.10.0`, but it's clearly there on PyPI. Perhaps a transient issue with PyPI. I have rerun the builds.

PR #411, opened 2019-06-05
be kinder about checking data keys and be kinder about yaml usage

CJ-Wright:
Just in case one doesn't have the newest yaml installed or somehow the data_keys and event data keys don't match

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,6 @@ before_install:
 matrix:
   fast_finish: true
   include:
-    - python: 3.5
     - python: 3.6
     - python: 3.7
       dist: xenial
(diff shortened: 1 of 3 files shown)


CJ-Wright:
@bluesky/maintainers I think this is ready for review. (test failures are either sphinx issues or server timeouts)

danielballan:
I think the YAML stuff is good to merge, but the data keys stuff is a very different issue and probably requires more thought. If you want to get the YAML in soon, a separate PR would help.

CJ-Wright:
Hmm ok. I have run into the data key stuff in production at PDF (see the 'PDF' databroker `1f529bb6-3a66-45d3-a122-37362b9ce82c`)

danielballan:
We should make validation stricter in order to catch that and then we should fix the broken EventDescriptors in the database.

PR #421, opened 2019-07-26
FIX: make asset registry thread safe

tacaswell:
This also has a few other clean ups to account for h5py
and Python moving under us.

### DIFF
--- a/databroker/assets/base_registry.py
+++ b/databroker/assets/base_registry.py
@@ -472,12 +472,14 @@ class RegistryTemplate(BaseRegistryRO):
 
         col = self._resource_col
 
-        return self._api.insert_resource(col, spec, rpath,
-                                         rkwargs,
-                                         self.known_spec,
-                                         root=root,
-                                         path_semantics=path_semantics,
-                                         **rs_kwarg)['uid']
+        return self._api.insert_resource(
+            col, spec, rpath,
+            rkwargs,
+            self.known_spec,
+            root=root,
+            path_semantics=path_semantics,
+            duplicate_exc=self.DuplicateKeyError,
+            **rs_kwarg)['uid']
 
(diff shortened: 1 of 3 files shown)


databroker/assets/sqlite.py
```diff
@@ -137,6 +192,62 @@ def reconnect(self):
...
+                # queue.
+                continue
+            try:
+                finished_event, name, args, kwargs, ret = item
+            except ValueError:
+                print("did not get the right number of values in {item}".
+                      format(item=item))
+                continue
+            try:
+                print(f'handling {name}')
```
danielballan: Stray print.

danielballan: I pushed a commit to fix this since I have checked it out anyway. :-)

tacaswell: Thanks.

databroker/assets/base_registry.py
```diff
@@ -472,12 +472,14 @@ def register_resource(self, spec, root, rpath, rkwargs,
...
-                                         root=root,
-                                         path_semantics=path_semantics,
-                                         **rs_kwarg)['uid']
+        return self._api.insert_resource(
+            col, spec, rpath,
+            rkwargs,
+            self.known_spec,
+            root=root,
+            path_semantics=path_semantics,
+            duplicate_exc=self.DuplicateKeyError,
```
danielballan, line 481: Good catch.

danielballan:
I checked the CI failures, and they are all related to server startup timeouts, safe to ignore.

databroker/assets/base_registry.py
tacaswell: This was failing on Python master (as it was trying to do `except None` and at some point throwing / catching things that do not derive from BaseException will be forbidden).

PR #423, opened 2019-07-30
CLN: remove dask dependency

mrakitin:
To address https://github.com/NSLS-II/PyXRF/issues/159#issuecomment-504426739.

This change may be temporary, as dask may be introduced in the later major versions.


mrakitin:
Thank you @danielballan!

PR #424, opened 2019-07-30
MNT/CI: rename <prefix>-requirements.txt to requirements-<prefix>.txt

mrakitin:
This is to add convenience when attempting to type `requirements` + <kbd>Tab</kbd>x2, so that it can show all the files at once.


danielballan:
Just one random server startup failure on one build.

mrakitin:
Thank you!

PR #425, opened 2019-07-30
Polish v0.13.2 release notes.

danielballan:


danielballan:
Per conversation with @mrakitin, I am adding a link to the v0.13.2 GitHub milestone. This is a good way to provide developers with a _full_ list of the changes, some of which are too low-level to be of interest in the published release notes for users.

Let's get in the habit of getting our milestones up to date when we release and adding links like this to all future release notes.

PR #427, opened 2019-08-01
REF: replace deprecated unordered bulk op

klauer:
* `ujson` appears to be a dependency of a dependency, which was not picked up when I installed the requirements in a pre-existing environment
* Deprecation information: https://api.mongodb.com/python/current/api/pymongo/collection.html?highlight=unordered%20bulk#pymongo.collection.Collection.initialize_unordered_bulk_op
* Replaced with: https://api.mongodb.com/python/current/api/pymongo/collection.html?highlight=unordered%20bulk#pymongo.collection.Collection.bulk_write
* bulk_write requires pymongo >= 3.0

### DIFF
--- a/databroker/assets/mongo_core.py
+++ b/databroker/assets/mongo_core.py
@@ -68,12 +68,12 @@ def bulk_insert_datum(col, resource, datum_ids,
                          datum_kwargs=dict(d_kwargs))
             yield datum
 
-    bulk = col.initialize_unordered_bulk_op()
     d_uids = deque()
+    bulk = deque()
     for dm in datum_factory():
-        bulk.insert(dm)
+        bulk.append(pymongo.InsertOne(dm))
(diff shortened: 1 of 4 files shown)


danielballan:
> bulk_write requires pymongo >= 3.0

Should the requirements be updated to that effect? Note that the relevant code is in `extras_require` in `setup.py`, not in `requirements.txt`.

danielballan:
And thanks for the databroker fix-up PR, @klauer!

klauer:
No problem, @danielballan. Modified the requirement and discovered/fixed a second deprecated bulk operation call.

Should I be worried about the test suite stability or primarily `test_*bulk*`?

danielballan:
The server startup tests are known to be very unstable, and all of those code is about to pulled out in #392 anyway. All others tests should be pretty stable.

PR #432, opened 2019-09-06
Add root_map

danielballan:
Currently ``root_map`` can only be set after `__init__`, via

```
cat = SomeCatalogClass(...)
cat.filler.root_map = {...}
```

and, moreover, it does not propagate into serach results. It obviously
should propagate into search results.

And it should be settable via `__init__` for consitency with
``handler_registry`` and so that it can be set in configuration files.

### DIFF
--- a/databroker/_drivers/mongo_embedded.py
+++ b/databroker/_drivers/mongo_embedded.py
@@ -207,7 +207,7 @@ class BlueskyMongoCatalog(Broker):
             handler_registry = {}
         parsed_handler_registry = parse_handler_registry(handler_registry)
         self.filler = event_model.Filler(
-            parsed_handler_registry, root_map=None, inplace=True)
+            parsed_handler_registry, root_map=root_map, inplace=True)
         super().__init__(**kwargs)
 
     def _get_event_pages(self, descriptor_uid, skip=0, limit=None):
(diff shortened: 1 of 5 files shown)


databroker/_drivers/mongo_embedded.py
```diff
@@ -204,7 +206,8 @@ def __init__(self, datastore_db, *, handler_registry=None,
         if handler_registry is None:
             handler_registry = {}
         parsed_handler_registry = parse_handler_registry(handler_registry)
-        self.filler = event_model.Filler(parsed_handler_registry, inplace=True)
+        self.filler = event_model.Filler(
+            parsed_handler_registry, root_map=None, inplace=True)
```
tacaswell:
```suggestion
            parsed_handler_registry, root_map=root_map, inplace=True)
```

gwbischof: should this be root_map=root_map?

databroker/_drivers/mongo_embedded.py
```diff
@@ -261,6 +264,7 @@ def search(self, query):
             datastore_db=self._db,
             query=query,
             handler_registry=self.filler.handler_registry,
+            root_map=self.filler.root_map,
```
gwbischof, line 267: wont this always be None with self.filler=Filler(root_map=None)?

PR #433, opened 2019-09-09
Fix fetch info of BlueskyRun not its parent catalog.

danielballan:
This is a subtle bug, made difficult by the deep class hierarchy
of intake Catalogs---something we might revisit in intake someday.

The RemoteBlueskyRun was fetching the full, un-paginated /v1/info of its
*parent* catalog (e.g. BlueskyMongoCatalog). This was "harmless" in the
sense that everything still worked, but it was very slow because the
parent catalog can be large.

This is partly of a relic of the early days of intake-bluesky dev when
my mental model of intake was not so good.


danielballan:
This is a hot-fix and not especially reviewable by anyone who hasn't lived in the depths of intake for awhile, so I'm going to self-merge.

PR #435, opened 2019-09-10
Discover handlers via 'databroker.handlers' entrypoint.

danielballan:
Continuing our theme of using entrypoints for discovery, this adds a new
``'databroker.handlers'`` entrypoint which packages can use to declare that
they ship some classes that may be used as handlers for certain specs.

This means that a handler registry can be specified:

1. directly, by passing a dict into the Catalog's `__init__`
2. via a v0-style configuration file that specifies a dict _to be passed_ into `__init__`
3. via an intake-style `catalog.yml` file with the same effect as (2)
4. implicitly, by discovering all installed handlers via the `'databroker.handlers'` entrypoint

Notice that we avoid any potentially-confusing configuration "merging" here.
Approach (4) is only activated if none of the other approaches are used and
thus ``handler_registry=None`` is passed in to the Catalog's ``__init__``.
But Approach (4) is convenient and likely sufficient for many users. If they

```
pip install area-detector-handlers  # new package, just added to PyPI
```

then databroker can discover the handlers that it declares

```py
In [1]: import databroker.core                                                                                                                                                                

In [2]: databroker.core.discover_handlers()                                                                                                                                                   
Out[2]: 
{'AD_CBF': area_detector_handlers.handlers.PilatusCBFHandler,
 'AD_HDF5': area_detector_handlers.handlers.AreaDetectorHDF5Handler,
 'AD_HDF5_SWMR': area_detector_handlers.handlers.AreaDetectorHDF5SWMRHandler,
 'AD_HDF5_SWMR_TS': area_detector_handlers.handlers.AreaDetectorHDF5SWMRTimestampHandler,
 'AD_HDF5_TS': area_detector_handlers.handlers.AreaDetectorHDF5TimestampHandler,
 'AD_SPE': area_detector_handlers.handlers.AreaDetectorSPEHandler,
 'AD_TIFF': area_detector_handlers.handlers.AreaDetectorTiffHandler,
 'XSP3': area_detector_handlers.handlers.Xspress3HDF5Handler}
```

and it will use this as its ``handler_registry`` unless a different one is
specified.

This was intentionally done at the databroker level, not in
``event_model.Filler`` so that the `Filler` always gets an explicit
``handler_registry`` and does not get involved in configuration management.

### DIFF
--- a/databroker/core.py
+++ b/databroker/core.py
@@ -1,5 +1,6 @@
 import collections
 import copy
+import entrypoints
 import event_model
 from datetime import datetime
 import dask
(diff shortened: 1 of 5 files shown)


danielballan:
Note the `1.0.0a2` tag on this branch---a somewhat unusual move in order to make this easy for alpha users to try it in the wild. (It works well on CSX data! :tada:) This branch should not be rebased.

databroker/core.py
```diff
@@ -1065,6 +1066,53 @@ def xarray_to_event_gen(data_xarr, ts_xarr, page_size):
...
+    group_all = entrypoints.get_group_all(entrypoint_group_name)
+    if len(group_all) != len(group):
+        # There are some name collisions. Let's go digging for them.
+        for name, matches in itertools.groupby(group_all, lambda ep: ep.name):
+            matches = list(matches)
+            if len(matches) != 1:
+                winner = group[name]
+                warnings.warn(
+                    f"There are {len(matches)} entrypoints for the "
+                    f"databroker handler spec {name!r}."
```
mrakitin:
```suggestion
                    f"databroker handler spec {name!r}. "
```

mrakitin:
@danielballan, how can I use the tag from the PR branch?

danielballan:
@mrakitin A tag on a PR branch alone doesn’t do much, but in this case, I uploaded a prerelease to PyPI from this tag so that alpha users can `pip install --pre databroker` and obtain a distribution that works. I would never upload a final release like this, but a prerelease feels OK in these circumstances.

PR #436, opened 2019-09-10
Added EntrypointsCatalog

gwbischof:


danielballan:
For anyone who happens to be following along, @gwbischof proposed this for intake in https://github.com/intake/intake/issues/404. This prototype will be pushed upstream shortly.

PR #442, opened 2019-09-17
Use vendored intake-xarray.

danielballan:
The motivation is to avoid taking on all the format-specific dependencies
in intake-xarray and just use the server--client piece.


gwbischof:
Looks good!

PR #456, opened 2019-09-25
API: canonical(...) always returns EventPage, never Event.

danielballan:
Thus, all *new* consumer code, using the `canonical(...)` method, will be
written to expected EventPages and thus do things in an efficient, vectorized
way.

Note that DatumPages were already handled correctly, so no changes were needed
for Datum.

This change exposed some typos in codepaths that were not previously covered.

### DIFF
--- a/databroker/core.py
+++ b/databroker/core.py
@@ -760,7 +760,8 @@ class BlueskyRun(intake.catalog.Catalog):
                             datum_ids |= set(datum_page['datum_id'])
                 if fill == 'yes':
                     self._fill(event)  # in place (for now)
-                payload.append(('event', event))
+                event_page = event_model.pack_event_page(event)
+                payload.append(('event_page', event_page))
             if i == self.npartitions - 1 and self._run_stop_doc is not None:
                 payload.append(('stop', self._run_stop_doc))
         for _, doc in payload:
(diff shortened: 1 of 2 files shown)


tacaswell:
:+1: in principle.

databroker/core.py
```diff
@@ -760,7 +760,8 @@ def read_partition(self, partition):
                             datum_ids |= set(datum_page['datum_id'])
                 if fill == 'yes':
                     self._fill(event)  # in place (for now)
-                payload.append(('event', event))
+                event_page = event_model.pack_event_page(event)
```
gwbischof, line 763: thought: this could also be done by changing interlace_event_pages to interlace_event_page_chunks. But that may require more work, maybe it is better to do that later.

gwbischof:
Is this a good time to add a chunk_size kwarg to canonical?

databroker/core.py
danielballan: Yeah, I just did the dumbest/simplest thing possible. There are a couple moving pieces involved in actually doing this efficiently.

danielballan:
We can add optional keyword arguments in a non-breaking fashion. I think we should tag 1.0.0b1 once this change (which is a breaking change) is in and then turn our attention to refactoring and extending functionality.

PR #457, opened 2019-09-25
Get entry on getitem

danielballan:
This is an experimental idea that I would like to _try_ in databroker. If we find that it creates more confusion than it alleviates, we should revert it before the `databroker.v2` interface is declared "stable for general users". If we find that it works well, we should explore moving it upstream to intake if @martindurant is game. (Martin, your comments on what follows are welcome here too!)

## Background and Motivation

To help users who don't have a complete mental model of intake, intake overrides `__getattr__` to blur the line between an Entry and its content. If I call `entry.<something>` when I mean `entry.get().<something>` intake will "do the right thing." Thus, I can get along not really understanding the entry/content distinction and sometimes I'm OK. But other times I can't help but bump into it. For example, this returns an `Entry`.

```py
catalog['1a86d0']
```

(And the same with our custom recency-based lookup, such as the ubiquitous `catalog[-1]`.)

Now, sometimes having an Entry is useful. I can, for example, obtain some metadata cheaply. This is typically useful in loops over large numbers of entries

```py
for key, entry in  catalog.items():
    populate_some_gui(entry.describe())
```

but when I look up a specific item via `__getitem__`, as in `catalog['1a86d0']`, what I typically want is its _content_ so I can see a nice repr and, if it's a sub-catalog, iterate over its contents. I can't do that to an Entry. I have to access the Entry's content via one of these:

```py
catalog['1a86d0'].get()
catalog['1a86d0']()
```
...which is not so obvious to our users or nice to look at. Now, if I use dot-access on it directly

```py
catalog['1a87d0'].primary.read()
```

intake automatically makes that into `catalog['1a87d0'].get().primary.get().read()`. But this doesn't help me when I want to stop at `catalog['1a87d0']` and examine it. Early feedback with our users has shown this is a major sticking point.

## Proposed Change

Above we said that the magic transforming "`.` -> `.get()`" blurs the line between an Entry and its content. Notice that it only does so in one direction. If I have an Entry, I can treat it like its content, and it will do the right thing.

```py
catalog['1a87d0'].get().primary
catalog['1a87d0'].primary  # Treating an Entry as if it were its content works.
```

But if I have an Entry's _content_ and I try to treat it like an Entry, that fails.

```py
catalog['1a87d0'].describe()  # works
catalog['1a87d0'].get().describe()  # FAILS! Can't treat content like an Entry
```

In this PR, we make that last line work. The blurriness between Entry and content now works in _both_ directions instead of only one. That is, `entry.get().<something>` falls back to `entry.<something>`, symmetric with how, `entry.<something>` falls back to `entry.get().<something>`.

This makes it possible to address the usability issue that proved to be a sticking point with our early users. Now, `__getitem__` on our Catalogs returns `entry.get()` instead of `entry` so that

```py
catalog['1a86d0']
```

returns content (a `BlueskyRun`) with a nice repr that you can iterate over. Note that `catalog.items()` still returns `(key, Entry)` pairs; the laziness of `Entry` is still accessible when you need it.

This _could_ turn out to be a terrible idea. It is certainly very magical and could leave users confused about the type of the object that they are holding. But I would argue that, if we have accepted the magical ~`__getitem__`~ `__getattr__` [edited] behavior in one direction, it is _less_ confusing to have it in both directions. At any rate it's worth trying in the wild.

Thanks to @tacaswell for helping me think this through.

### DIFF
--- a/databroker/core.py
+++ b/databroker/core.py
@@ -497,7 +497,7 @@ class RemoteBlueskyRun(intake.catalog.base.RemoteCatalog):
         try:
             start = self.metadata['start']
             stop = self.metadata['stop']
-            out = (f"Run Catalog\n"
+            out = (f"BlueskyRun\n"
                    f"  uid={start['uid']!r}\n"
                    f"  exit_status={stop.get('exit_status')!r}\n"
                    f"  {_ft(start['time'])} -- {_ft(stop.get('time', '?'))}\n"
--- a/databroker/core.py
+++ b/databroker/core.py
@@ -666,25 +666,25 @@ class BlueskyRun(intake.catalog.Catalog):
                 getshell=True,
                 catalog=self)
 
-    def get(self):
+    def get(self, *args, **kwargs):
         """
-        Return self.
+        Return self or, if args are provided, some new instance of type(self).
 
         This is here so that the user does not have to remember whether a given
         variable is a BlueskyRun or an *Entry* with a Bluesky Run. In either
         case, ``obj.get()`` will return a BlueskyRun.
         """
-        return self
+        return self._entry.get(*args, **kwargs)
 
-    def __call__(self):
+    def __call__(self, *args, **kwargs):
         """
-        Return self.
+        Return self or, if args are provided, some new instance of type(self).
 
         This is here so that the user does not have to remember whether a given
         variable is a BlueskyRun or an *Entry* with a Bluesky Run. In either
         case, ``obj()`` will return a BlueskyRun.
         """
-        return self.get()
+        return self.get(*args, **kwargs)
 
     def __getattr__(self, key):
         try:
--- a/databroker/core.py
+++ b/databroker/core.py
@@ -666,6 +666,28 @@ class BlueskyRun(intake.catalog.Catalog):
                 getshell=True,
                 catalog=self)
 
+    def get(self):
+        """
+        Return self.
+
+        This is here so that the user does not have to remember whether a given
+        variable is a BlueskyRun or an *Entry* with a Bluesky Run. In either
+        case, ``obj.get()`` will return a BlueskyRun.
+        """
+        return self
+
+    def __call__(self):
+        """
+        Return self.
+
+        This is here so that the user does not have to remember whether a given
+        variable is a BlueskyRun or an *Entry* with a Bluesky Run. In either
+        case, ``obj()`` will return a BlueskyRun.
+        """
+        return self.get()
+
+    def __getattr__(self, key):
+        return getattr(self._entry, key)
 
     def canonical(self, *, fill):
         """
--- a/databroker/core.py
+++ b/databroker/core.py
@@ -11,6 +11,7 @@ import heapq
 import importlib
 import itertools
 import intake.catalog.base
+import intake.catalog.local
 import errno
 import intake.container.base
 from intake.compat import unpack_kwargs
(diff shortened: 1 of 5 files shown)


databroker/core.py
```diff
@@ -647,6 +666,28 @@ def _load(self):
                 getshell=True,
                 catalog=self)
 
+    def get(self):
```
tacaswell: does this need to be able to pass `arg, kwargs` trough?

databroker/in_memory.py
```diff
@@ -7,18 +7,10 @@
 from mongoquery import Query
 
 
-from .core import parse_handler_registry, discover_handlers
+from .core import parse_handler_registry, discover_handlers, Entry
 from .v2 import Broker
 
 
-class SafeLocalCatalogEntry(intake.catalog.local.LocalCatalogEntry):
-    # For compat with intake 0.5.1.
```
tacaswell, line 15: have we pinned to make sure we have a new enough version?

danielballan: Yes, we are pinned to 0.5.2 which (I have just re-confirmed) has the fix.

databroker/core.py
danielballan: When I wrote this I intentionally did _not_ support it, but now I think that was based on a confused notion. Added in fcc489f86e0d78991c8193b442dbe19f8e9a6e1e.

martindurant:
You are right that this seems to confuse people...

If you are to make two things that act as if they are one thing, can you not just make one thing? I mean, what have been entries so far would become sources which have not been initialised, and an attempt to do any of the source-y things (including "call" and "get") would force initialisation, but you can always re-initialise with other parameter overrides.

(such a change would require API breakages)

Any any case, I would post an issue on Intake, so that others can discuss this idea too.

martindurant:
Note that this "is it initialised" already happens to some extent within sources, e.g., some stuff happens in `__init__`, but we defer getting the schema, and then getting the full data, untile explicitly requested.

danielballan:
> If you are to make two things that act as if they are one thing, can you not just make one thing?

I think this is a key point. The best architecture, if we can achieve without inflicting too much breakage on users, may be to merge `CatalogEntry` and `DataSource` into one thing. Instantiating that thing should be cheap, but calling methods like `.read()` or, if it's a Catalog, `.items()` would trigger more information to be pulled up.

databroker/core.py
```diff
@@ -666,25 +666,25 @@ def _load(self):
                 getshell=True,
                 catalog=self)
 
-    def get(self):
+    def get(self, *args, **kwargs):
         """
-        Return self.
+        Return self or, if args are provided, some new instance of type(self).
```
tacaswell, line 671: The caching happens at the entry level?

danielballan: Yes.

tacaswell:
This reminds me of the new `cached_property` (https://docs.python.org/3.8/library/functools.html?highlight=cached_property#functools.cached_property) decorator + (as @danielballan suggested in our slack) `namedtuple._replace`.  It is lazy by default, some actions cause it to cache data / become heavy weight under the hood, and there is a way to get a new instance with different initial values.

I think it might also be ok if both classes pick up a `entry` attribute so that you can always do `obj.entry.get(*args, **kwargs)` to get your self the contents with your own choice of calling arguments, but I think that pushing them together into one class is probably a good idea or this full-mirroring is a better idea.

gwbischof:
I support this idea, I also like the idea of eventually merging `CatalogEntry` and `DataSource`.  Crazy idea: I think nested catalogs are DAGs. If CatalogEntry+DataSource inherits from a Node class from a graph library, it could enable some nice features. For example DFS/BFS searches on the nested catalog, that returns a new catalog with edges pointing to the found nodes.
PR looks good to me, merging...

PR #458, opened 2019-09-26
changes to canonical and read_partition

gwbischof:
- Adding a chunk_size kwarg to canonical, this will set the size of the event_pages that are returned.
- read_partition needs to be changed around to accommodate this.
- canonical always returns the datum along with filled events.
- canonical is a snapshot and returns only the documents present when it is called. So if canonical is called in the middle of the run, you won't get all of the run, but it can be called again to more documents.
- we plan to change read_partition so that the whole unfilled run gets loaded on the first call.
- the first partition will be the header(start, descriptor, resources)


databroker/core.py
```diff
@@ -761,90 +821,59 @@ def datum_kwarg_gen():
...
+            resource_uid = self._lookup_resource_for_datum(datum_id)
+        resource = self._get_resource(uid=resource_uid)
+
+        # Use rechunk datum pages to make them into pages of size "partition_size"
+        # and yield one page per partition.  Rechunk might be slow.
+        datum_gen = self._get_datum_pages(resource['uid'])
+        partitions = [[('datum_page', datum_page)] for datum_page in
+                      event_model.rechunk_datum_pages(datum_gen, partition_size)]
+
+        # Datum not found.  There is probably a nicer way to write this.
```
gwbischof: This isn't correct, I'll fix it up soon

databroker/core.py
```diff
@@ -585,9 +642,12 @@ def __init__(self,
         self._get_event_pages = get_event_pages
         self._get_event_count = get_event_count
         self._get_resource = get_resource
+        self._get_resources = get_resources
         self._lookup_resource_for_datum = lookup_resource_for_datum
         self._get_datum_pages = get_datum_pages
-        self.filler = filler
+        self.fillers = {'yes': event_model.Filler(),
```
danielballan: These will need a handler registry. Also, not sure we want a separate `Filler` instance per `BlueskyRun`. One role of the `Filler` object is the control caching of documents and files. If two users request the same file, would be need to have one file handle.

databroker/core.py
```diff
@@ -484,7 +541,7 @@ def to_dask(self):
     def _close(self):
         self.bag = None
 
-    def canonical(self, *, fill):
+    def canonical(self, *, fill, chunk_size=1):
```
danielballan: I think our options are:
* Require the user to specify a `chunk_size` or explicitly `chunk_size='auto'`. This is how `dask.array` used to work.
* Use `chunk_size='auto'` by default. This is how `dask.array` works nowadays.

In the absence of user input, I think we're likely to get the best performance from deferring to the server rather than imposing a fixed chunk size.

databroker/core.py
```diff
@@ -1308,6 +1341,55 @@ def dataarray_page_to_dataset_page(dataarray_page):
             'filled': xarray.merge(dataarray_page['filled'].values())}
 
 
+class NoFiller(event_model.Filler):
+
+    # Overloading the function so that changes event_model wont break it.
```
danielballan, line 1634: But likewise _fixes_ to event-model won't propagate here.

danielballan:
Superseded by #469

PR #464, opened 2019-10-23
Fix typo in EntrypointsCatalog

dylanmcreynolds:
Found a very small typo in discovery.py. 

I investigated adding a pytest to stress this code, but not sure how to setup entry points in a test.

### DIFF
--- a/databroker/discovery.py
+++ b/databroker/discovery.py
@@ -53,7 +53,7 @@ class EntrypointsCatalog(Catalog):
             try:
                 self._entries[name] = EntrypointEntry(entrypoint)
             except Exception as e:
-                warings.warn(f"Failed to load {name}, {entrypoint}, {e!r}.")
+                warnings.warn(f"Failed to load {name}, {entrypoint}, {e!r}.")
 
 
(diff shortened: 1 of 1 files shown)


gwbischof:
Thanks!

We can update this test: https://github.com/bluesky/databroker/blob/master/databroker/tests/test_discovery.py

There might be another problem here, ill think about it more this afternoon.

gwbischof:
in `test_catalog_discovery()` the UserWarning is generated by `list_configs`: https://github.com/bluesky/databroker/blob/7d1682e2a94797f2d720362364bf9d1673754e13/databroker/utils.py#L254

gwbischof:
I'm not sure how to make this line raise an exception: https://github.com/bluesky/databroker/blob/master/databroker/discovery.py#L54

The __init__ of EntrypointEntry is just an assignment. The Entrypoint isn't actually loaded until `get()` is called. Then the EntrypointEntry is put in the _entries dict. I don't think this line can raise an exception.

I think this try/except isn't needed.
Thoughts?

danielballan:
Good point.

PR #465, opened 2019-10-25
Wip dask issue

dylanmcreynolds:
Fixes for an issue found where a local databroker's to_dask() returns full numpy arrays.

### DIFF
--- a/databroker/_drivers/jsonl.py
+++ b/databroker/_drivers/jsonl.py
@@ -2,6 +2,7 @@ import glob
 import json
 import os
 import pathlib
+import event_model
 
 from ..in_memory import BlueskyInMemoryCatalog
 from ..core import tail
@@ -54,8 +55,9 @@ def get_stop(filename):
 class BlueskyJSONLCatalog(BlueskyInMemoryCatalog):
     name = 'bluesky-jsonl-catalog'  # noqa
 
-    def __init__(self, paths, *,
-                 handler_registry=None, root_map=None, query=None, **kwargs):
+    def __init__(self, paths, *, handler_registry=None, root_map=None,
+                 filler_class=event_model.Filler,
+                 query=None, **kwargs):
         """
         This Catalog is backed by a newline-delimited JSON (jsonl) file.
 
@@ -68,11 +70,36 @@ class BlueskyJSONLCatalog(BlueskyInMemoryCatalog):
         paths : list
             list of filepaths
         handler_registry : dict, optional
-            Maps each asset spec to a handler class or a string specifying the
-            module name and class name, as in (for example)
-            ``{'SOME_SPEC': 'module.submodule.class_name'}``.
-        root_map : dict, optional
-            Maps resource root paths to different paths.
+            This is passed to the Filler or whatever class is given in the
+            filler_class parametr below.
+
+            Maps each 'spec' (a string identifying a given type or external
+            resource) to a handler class.
+
+            A 'handler class' may be any callable with the signature::
+
+                handler_class(resource_path, root, **resource_kwargs)
+
+            It is expected to return an object, a 'handler instance', which is also
+            callable and has the following signature::
+
+            handler_instance(**datum_kwargs)
+
+            As the names 'handler class' and 'handler instance' suggest, this is
+            typically implemented using a class that implements ``__init__`` and
+            ``__call__``, with the respective signatures. But in general it may be
+            any callable-that-returns-a-callable.
+        root_map: dict, optional
+            This is passed to Filler or whatever class is given in the filler_class
+            parameter below.
+
+            str -> str mapping to account for temporarily moved/copied/remounted
+            files.  Any resources which have a ``root`` in ``root_map`` will be
+            loaded using the mapped ``root``.
+        filler_class: type, optional
+            This is Filler by default. It can be a Filler subclass,
+            ``functools.partial(Filler, ...)``, or any class that provides the
+            same methods as ``DocumentRouter``.
         query : dict, optional
             Mongo query that filters entries' RunStart documents
         **kwargs :
@@ -86,8 +113,8 @@ class BlueskyJSONLCatalog(BlueskyInMemoryCatalog):
         self._filename_to_mtime = {}
         super().__init__(handler_registry=handler_registry,
                          root_map=root_map,
-                         query=query,
-                         **kwargs)
+                         filler_class=filler_class,
+                         query=query, **kwargs)
 
     def _load(self):
         for path in self.paths:
@@ -122,8 +149,8 @@ class BlueskyJSONLCatalog(BlueskyInMemoryCatalog):
         cat = type(self)(
             paths=self.paths,
             query=query,
-            handler_registry=self.filler.handler_registry,
-            root_map=self.filler.root_map,
+            handler_registry=self._handler_registry,
+            root_map=self._root_map,
             name='search results',
             getenv=self.getenv,
             getshell=self.getshell,
--- a/databroker/_drivers/jsonl.py
+++ b/databroker/_drivers/jsonl.py
@@ -71,7 +71,7 @@ class BlueskyJSONLCatalog(BlueskyInMemoryCatalog):
             list of filepaths
         handler_registry : dict, optional
             This is passed to the Filler or whatever class is given in the
-            filler_class parametr below.
+            ``filler_class`` parameter below.
 
             Maps each 'spec' (a string identifying a given type or external
             resource) to a handler class.
@@ -90,8 +90,8 @@ class BlueskyJSONLCatalog(BlueskyInMemoryCatalog):
             ``__call__``, with the respective signatures. But in general it may be
             any callable-that-returns-a-callable.
         root_map: dict, optional
-            This is passed to Filler or whatever class is given in the filler_class
-            parameter below.
+            This is passed to Filler or whatever class is given in the
+            ``filler_class`` parameter below.
 
             str -> str mapping to account for temporarily moved/copied/remounted
             files.  Any resources which have a ``root`` in ``root_map`` will be
@@ -111,6 +111,7 @@ class BlueskyJSONLCatalog(BlueskyInMemoryCatalog):
             paths = [paths]
         self.paths = paths
         self._filename_to_mtime = {}
+
         super().__init__(handler_registry=handler_registry,
                          root_map=root_map,
                          filler_class=filler_class,
--- a/databroker/_drivers/mongo_embedded.py
+++ b/databroker/_drivers/mongo_embedded.py
@@ -174,7 +174,8 @@ class _Entries(collections.abc.Mapping):
 
 
 class BlueskyMongoCatalog(Broker):
-    def __init__(self, datastore_db, *, query=None, **kwargs):
+    def __init__(self, datastore_db, *, handler_registry=None, root_map=None,
+                 filler_class=event_model.Filler, query=None, **kwargs):
         """
         This Catalog is backed by a MongoDB with an embedded data model.
 
@@ -224,7 +225,11 @@ class BlueskyMongoCatalog(Broker):
         else:
             self._db = datastore_db
         self._query = query or {}
-        super().__init__(**kwargs)
+
+        super().__init__(handler_registry=handler_registry,
+                         root_map=root_map,
+                         filler_class=filler_class,
+                         **kwargs)
 
     def _get_event_pages(self, descriptor_uid, skip=0, limit=None):
         if limit is None:
--- a/databroker/_drivers/mongo_embedded.py
+++ b/databroker/_drivers/mongo_embedded.py
@@ -78,7 +78,7 @@ class _Entries(collections.abc.Mapping):
             get_resource=get_resource,
             lookup_resource_for_datum=lookup_resource_for_datum,
             get_datum_pages=self.catalog._get_datum_pages,
-            filler=self.catalog.filler)
+            filler=self.catalog._get_filler())
         return Entry(
             name=run_start_doc['uid'],
             description={},  # TODO
@@ -175,6 +175,7 @@ class _Entries(collections.abc.Mapping):
 
 class BlueskyMongoCatalog(Broker):
     def __init__(self, datastore_db, *, handler_registry=None, root_map=None,
+                 filler_class=event_model.Filler,
                  query=None, **kwargs):
         """
         This Catalog is backed by a MongoDB with an embedded data model.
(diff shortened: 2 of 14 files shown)


databroker/core.py
```diff
@@ -1308,54 +1296,34 @@ def dataarray_page_to_dataset_page(dataarray_page):
...
-    def event_page(self, doc):
-
-        @dask.delayed
-        def delayed_fill(event_page, key):
-            self.fill_event_page(event_page, include=key)
-            return numpy.asarray(event_page['data'][key])
+def coerce_dask(handler_class, filler_state):
+    # If the handler has its own delayed logic, defer to that.
+    if hasattr(handler_class, 'return_type'):
+        if handler_class.return_type['delayed']:
```
tacaswell, line 1325:
```suggestion
        if handler_class.return_type.get('delayed', False):
```

danielballan: The strict lookup was intentional. Do we want to go down the path of "If the `return_type` attribute exists, it can have whatever inside?" That level of flexibility has made the hints code very messy. I would rather say, "If `return_type` exists, it must take a position on 'delayed'."

databroker/_drivers/jsonl.py
```diff
@@ -68,11 +70,36 @@ def __init__(self, paths, *,
...
+            callable and has the following signature::
+
+            handler_instance(**datum_kwargs)
+
+            As the names 'handler class' and 'handler instance' suggest, this is
+            typically implemented using a class that implements ``__init__`` and
+            ``__call__``, with the respective signatures. But in general it may be
+            any callable-that-returns-a-callable.
+        root_map: dict, optional
+            This is passed to Filler or whatever class is given in the filler_class
```
mrakitin:
```suggestion
            This is passed to Filler or whatever class is given in the ``filler_class``
```

databroker/_drivers/mongo_embedded.py
```diff
@@ -189,12 +188,29 @@ def __init__(self, datastore_db, *, handler_registry=None, root_map=None,
...
+        root_map: dict, optional
+            This is passed to Filler or whatever class is given in the filler_class
+            parameter below.
+             str -> str mapping to account for temporarily moved/copied/remounted
+            files.  Any resources which have a ``root`` in ``root_map`` will be
+            loaded using the mapped ``root``.
+        filler_class: type, optional
+            This is Filler by default. It can be a Filler subclass,
+            ``functools.partial(Filler, ...)``, or any class that provides the
+            same methods as ``DocumentRouter``.
```
mrakitin, line 214: The `handler_registry`, `root_map` and `filler_class` are not explicitly declared as arguments. Should they be defined?

databroker/_drivers/mongo_normalized.py
```diff
@@ -150,12 +150,33 @@ def __init__(self, metadatastore_db, asset_registry_db, *,
...
+        root_map: dict, optional
+            This is passed to Filler or whatever class is given in the
+            filler_class parameter below. 
+            str -> str mapping to account for temporarily
+            moved/copied/remounted files.  Any resources which have a ``root``
+            in ``root_map`` will be loaded using the mapped ``root``.
+        filler_class: type, optional
+            This is Filler by default. It can be a Filler subclass,
+            ``functools.partial(Filler, ...)``, or any class that provides the
+            same methods as ``DocumentRouter``.
```
mrakitin, line 180: Same question as above.

databroker/_drivers/msgpack.py
```diff
@@ -63,11 +66,35 @@ def __init__(self, paths, *,
         paths : list
             list of filepaths
         handler_registry : dict, optional
-            Maps each asset spec to a handler class or a string specifying the
-            module name and class name, as in (for example)
-            ``{'SOME_SPEC': 'module.submodule.class_name'}``.
-        root_map : dict, optional
-            Maps resource root paths to different paths.
+            This is passed to the Filler or whatever class is given in the
+            filler_class parametr below.
```
mrakitin:
```suggestion
            ``filler_class`` parameter below.
```

databroker/tests/test_broker.py
```diff
@@ -576,7 +576,17 @@ def test_handler_options(db, RE, hw):
...
+
+def test_external_access_with_handler(db, RE, hw):
+    from ophyd.sim import NumpySeqHandler
+
+    RE.subscribe(db.insert)
+    rs_uid, = RE(count([hw.img], 2))
+
+    h = db[rs_uid]
+
+    # For some db fixtures, this is a already registered and is therefore a
```
mrakitin:
```suggestion
    # For some db fixtures, this is already registered and is therefore a
```

databroker/v2.py
```diff
@@ -10,6 +14,32 @@ class Broker(Catalog):
 
     It includes an accessor the databroker API version 1.
     """
+
+    def __init__(self, *, handler_registry=None, root_map=None,
+                 filler_class=event_model.Filler,**kwargs):
```
mrakitin:
```suggestion
                 filler_class=event_model.Filler, **kwargs):
```

databroker/v2.py
```diff
@@ -10,6 +14,32 @@ class Broker(Catalog):
...
+        if handler_registry is None:
+            handler_registry = discover_handlers()
+        self._handler_registry = parse_handler_registry(handler_registry)
+        self.handler_registry = event_model.HandlerRegistryView(
+            self._handler_registry)
+
+        self._get_filler = partial(self._filler_class,
+                              handler_registry=self.handler_registry,
+                              root_map=self._root_map,
+                              inplace=False)
```
mrakitin: Consider aligning/indenting those 3 lines at the level of `self._filler_class,`.

databroker/core.py
```diff
@@ -301,9 +301,10 @@ def documents_to_xarray(*, start_doc, stop_doc, descriptor_docs,
             continue
         if any(data_keys[key].get('external') for key in keys):
             filler('descriptor', descriptor)
+            filled_events = []
             for event in events:
                 try:
-                    filler('event', event)
+                    filled_events.append(filler('event', event)[1])
```
mrakitin: Why do we append the element `[1]` from the filler?

databroker/core.py
```diff
@@ -366,26 +361,19 @@ def documents_to_xarray(*, start_doc, stop_doc, descriptor_docs,
...
-                    # Construct the same default dimension names xarray would.
-                    dims = tuple(f'dim_{next(dim_counter)}' for _ in range(ndim))
+                field_metadata = data_keys[key]
+                ndim = len(field_metadata['shape'])
+                # if the EventDescriptor doesn't provide names for the
+                # dimensions (it's optional) use the same default dimension
+                # names that xarray would.
+                dims = field_metadata.get(
+                    'dims',
+                    tuple(f'dim_{next(dim_counter)}' for _ in range(ndim)))
```
mrakitin: I don't think the `dim_counter` is reset to the initial value anywhere, in the `for`-loop here one may get `dims = ('dim_100', 'dim_101', ...)`. Is it expected?

databroker/core.py
```diff
@@ -499,6 +487,15 @@ def canonical(self, *, fill):
...
         """
+        # Special case for 'delayed' since it *is* supported in the local mode
+        # of usage.
+        if fill == 'delayed':
+            raise NotImplementedError(
+                "Delayed access is not yet supported via the client--server "
+                "usage.")
+        FILL_OPTIONS = {'yes', 'no'}
+        if fill not in FILL_OPTIONS:
+            raise ValueError(f"Options for 'fill' are {FILL_OPTIONS}")
```
mrakitin: I think it's worth to also print the value of `fill` as well to understand what we are hitting.

databroker/core.py
```diff
@@ -587,7 +584,12 @@ def __init__(self,
         self._get_resource = get_resource
         self._lookup_resource_for_datum = lookup_resource_for_datum
         self._get_datum_pages = get_datum_pages
-        self.filler = filler
+        filler = get_filler()
+        self.fillers = {}
+        print("HRHR", filler.handler_registry)
```
mrakitin: @gwbischof pointed to this left debug print in person.

databroker/core.py
```diff
@@ -705,12 +707,18 @@ def canonical(self, *, fill):
...
+            arrays, meaning that the I/O can be deferred until the data is
+            actually needed.
             If fill is 'no', the Event documents will contain foreign keys as
             placeholders for the data. This option is useful for exporting
             copies of the documents.
 
         """
+        FILL_OPTIONS = {'yes', 'no', 'delayed'}
+        if fill not in FILL_OPTIONS:
+            raise ValueError(f"Options for 'fill' are {FILL_OPTIONS}")
```
mrakitin: See my comment above.

databroker/core.py
```diff
@@ -1308,54 +1316,34 @@ def dataarray_page_to_dataset_page(dataarray_page):
...
 
-        for key in needs_filling:
-            shape = extract_shape(descriptor, key)
-            dtype = extract_dtype(descriptor, key)
-            filled_doc['data'][key] = array.from_delayed(
-                delayed_fill(filled_doc, key), shape=shape, dtype=dtype)
-        return filled_doc
+# This adds a 'delayed' option to event_model.Filler's `coerce` parameter.
+# By adding it via plugin, we avoid adding a dask.array dependency to
+# event_model and we keep the fiddly hacks into extract_shape here in
```
mrakitin:
```suggestion
# event-model and we keep the fiddly hacks into extract_shape here in
```

databroker/core.py
```diff
@@ -1394,6 +1387,58 @@ def extract_dtype(descriptor, key):
         return reported
 
 
+class NoFiller(event_model.Filler):
```
mrakitin: It feels like this class should be moved to `event-model`. Are there any reasons why not?

databroker/core.py
danielballan: `filler` returns a `name, doc` pair. We want the doc.

mrakitin:
As a side note, should we try to have Python 3.8 in the matrix as well?

databroker/core.py
danielballan: Applied and pushed manually since I don't have permission to accept a revision.

danielballan:
Hahaha, please not in _this_ PR. So complicated already! Can you open an issue on lightsources about doing that for all the repos?

databroker/core.py
danielballan: https://github.com/bluesky/event-model/pull/127

databroker/_drivers/mongo_embedded.py
danielballan: I would have personally preferred to leave them as explicitly declared args, but I think @gwbischof did this on @tacaswell's suggestion and I don't feel strongly.

danielballan: To be clear, they don't technically have to be declared because they are captured by the base class, `Broker`. The reason one might declare them anyway is just to make the signature more obvious / self-describing.

databroker/core.py
mrakitin: Great, thanks for the explanation. Do you think it would be useful to put exactly this comment into the code for the convenience of a random reader not encountered this thread?

databroker/core.py
mrakitin: Thanks!

mrakitin:
Done! https://github.com/lightsources/project-management/issues/75

databroker/_drivers/mongo_embedded.py
mrakitin: It would be cleaner to either (i) declare them explicitly via keyword arguments and keep the docstrings or (ii) assume they are passed via `**kwargs` to the base class, and remove the corresponding docstrings. Right now we ave a mixture of (i) and (ii). I am in favor or either of them, not the mixture.

gwbischof: I went with solution (i)

databroker/_drivers/jsonl.py
```diff
@@ -68,11 +70,36 @@ def __init__(self, paths, *,
         paths : list
             list of filepaths
         handler_registry : dict, optional
-            Maps each asset spec to a handler class or a string specifying the
-            module name and class name, as in (for example)
-            ``{'SOME_SPEC': 'module.submodule.class_name'}``.
-        root_map : dict, optional
-            Maps resource root paths to different paths.
+            This is passed to the Filler or whatever class is given in the
+            filler_class parametr below.
```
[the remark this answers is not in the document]
gwbischof: I don't have permission to commit these changes through Github, I changed them by hand.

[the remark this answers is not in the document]
gwbischof: fixed.

databroker/_drivers/jsonl.py
gwbischof: fixed

databroker/_drivers/mongo_normalized.py
gwbischof: fixed

databroker/_drivers/msgpack.py
gwbischof: fixed

databroker/tests/test_broker.py
gwbischof: fixed

databroker/v2.py
gwbischof: fixed

databroker/v2.py
gwbischof: fixed

databroker/core.py

databroker/core.py
gwbischof: I'm not sure about the answer for this. I am planning to replace this function in a future PR, which will get rid of this code.

databroker/core.py
gwbischof: fixed

databroker/core.py
gwbischof: fixed

databroker/core.py
mrakitin: It used to be a conditional (`if dims is None: ...`) increment of the counter, but now it's unconditional, and the counter is not reset anywhere, which concerns me. I am not sure how serious the consequences of it, but I don't think we should let this potentially buggy code to appear in master even if the future PRs are going to replace the function. It's totally possible I am missing something.

@danielballan, do you have an idea about it?

danielballan: That's a good point. My intention here is that the counter would only be increment is `field_metadata` was missing the key `'dims'`, but as you point out the expression `tuple(f'dim_{next(dim_counter)}' for _ in range(ndim))` will be evaluated every time, whether `'dims'` is missing or not. Fix pushed.

databroker/core.py
```diff
@@ -313,11 +315,14 @@ def documents_to_xarray(*, start_doc, stop_doc, descriptor_docs,
                     for datum_page in get_datum_pages(resource_uid):
                         filler('datum_page', datum_page)
                     # TODO -- When to clear the datum cache in filler?
-                    filler('event', event)
+                    _, filled_event = filler('event', event)
+                    filled_events.append(filled_event)
```
mrakitin, line 319: 👍 I like this more explicit approach.

databroker/core.py
```diff
@@ -366,26 +364,20 @@ def documents_to_xarray(*, start_doc, stop_doc, descriptor_docs,
...
-                    reported_ndim = len(field_metadata['dims'])
-                    if reported_ndim == ndim:
-                        dims = tuple(field_metadata['dims'])
-                    else:
-                        # TODO Warn
-                        ...
-                if dims is None:
-                    # Construct the same default dimension names xarray would.
+                field_metadata = data_keys[key]
+                ndim = len(field_metadata['shape'])
```
mrakitin, line 368: Should this line be moved under the `except KeyError` below as it's done in the code block above?

databroker/core.py
mrakitin: Thanks.

databroker/core.py
danielballan: Good eye, but no, in this case we use it in an additional place below, regardless of whether we hit the `KeyError` codepath or not.

mrakitin: You are right, I didn't check the code below carefully.

PR #468, opened 2019-11-08
Fix uid regex

tacaswell:
This is apparently a long standing bug!  The old rule would match the prefix anywhere, the new rule matches it only at the beginning.


danielballan:
Oops, I noticed after merging that I short-circuited a requested review from @gwbischof. Sorry.

PR #469, opened 2019-11-15
updates to canonical, read_partition

gwbischof:
Adding a chunk_size kwarg to canonical, this will set the size of the event_pages that are returned.
read_partition needs to be changed around to accommodate this.
canonical always returns the datum along with filled events.
canonical is a snapshot and returns only the documents present when it is called. So if canonical is called in the middle of the run, you won't get all of the run, but it can be called again to more documents.
we plan to change read_partition so that the whole unfilled run gets loaded on the first call.
the first partition will be the header(start, descriptor, resources)


databroker/_drivers/mongo_embedded.py
```diff
@@ -225,12 +226,23 @@ def __init__(self, datastore_db, *, handler_registry=None, root_map=None,
...
         super().__init__(handler_registry=handler_registry,
                          root_map=root_map,
                          filler_class=filler_class,
                          **kwargs)
 
+    def _get_filler(self):
+        return self._filler_class(
+                self._handler_registry, root_map=self._root_map, inplace=False)
+
+    def _get_delayed_filler(self):
```
danielballan: This method and at least one other like it in other modules survived our purge during the rebase.

danielballan: Re-upping this

databroker/_drivers/mongo_normalized.py
```diff
@@ -37,24 +39,25 @@ def get_run_start():
...
-            direct_access='forbid',  # ???
-            args=args,
-            cache=None,  # ???
-            parameters=[],
-            metadata=entry_metadata,
-            catalog_dir=None,
-            getenv=True,
-            getshell=True,
-            catalog=self.catalog)
+                name=run_start_doc['uid'],
```
danielballan: The whitespace was right as it was.

databroker/_drivers/mongo_normalized.py
```diff
@@ -112,15 +115,21 @@ def __getitem__(self, name):
                     raise KeyError(f"No run with scan_id={N}")
         if run_start_doc is None:
             raise KeyError(name)
-        entry = self._doc_to_entry(run_start_doc)
+        uid = run_start_doc['uid']
+        try:
+            entry = self._cache[uid]
+            print('cache hit')
```
danielballan: This looks like it's from the commit of mine that you cherry-picked. That work-in-progress should be removed from this PR since it didn't immediately fix the performance issue you were investigating.

databroker/_drivers/mongo_normalized.py
```diff
@@ -210,6 +219,14 @@ def __init__(self, metadatastore_db, asset_registry_db, *,
                          filler_class=filler_class,
                          **kwargs)
 
+    def _get_filler(self):
+        return self._filler_class(
+                self._handler_registry, root_map=self._root_map, inplace=False)
+
+    def _get_delayed_filler(self):
```
danielballan: As above

databroker/core.py
```diff
@@ -476,7 +623,7 @@ def to_dask(self):
     def _close(self):
         self.bag = None
 
-    def canonical(self, *, fill):
+    def canonical(self, *, fill, strict_order=False):
```
danielballan: Did we discuss the best default value for this? My immediate gut feeling is that it should be `True`. Consumers expects documents in the order they were received from the `RunEngine` may be sensitive to this. Advanced users can choose `False` if they are sure it is safe to do so in order to get faster throughput.

databroker/core.py
```diff
@@ -489,7 +636,8 @@ def canonical(self, *, fill):
             If fill is 'no', the Event documents will contain foreign keys as
             placeholders for the data. This option is useful for exporting
             copies of the documents.
-
+        strict_order : bool
```
danielballan:
```suggestion
        strict_order : bool, optional
```

databroker/core.py
```diff
@@ -489,7 +636,8 @@ def canonical(self, *, fill):
...
+            documents are strictly yielded in ascending time order.
```
danielballan, line 627: Note the default value here.

databroker/core.py
```diff
@@ -571,6 +760,7 @@ def __init__(self,
                  get_event_pages,
                  get_event_count,
                  get_resource,
+                 get_resources,
```
danielballan, line 752: This should be added to the docstring.

databroker/core.py
```diff
@@ -657,15 +847,17 @@ def _load(self):
...
+                          'resources': self._resources})
             self._entries[stream_name] = intake.catalog.local.LocalCatalogEntry(
                 name=stream_name,
                 description={},  # TODO
                 driver='databroker.core.BlueskyEventStream',
                 direct_access='forbid',
                 args=args,
-                cache=None,  # ???
-                metadata={'descriptors': descriptors},
+                cache=[self._cache],  # This doesn't seem to do anything
```
danielballan: I think we should revert this change at least until we understand exactly what this does.

databroker/core.py
```diff
@@ -701,7 +893,7 @@ def __getattr__(self, key):
             # before giving up.
             return getattr(self._entry, key)
 
-    def canonical(self, *, fill):
+    def canonical(self, *, fill, strict_order=False):
```
danielballan: As above

databroker/core.py
```diff
@@ -717,14 +909,27 @@ def canonical(self, *, fill):
             If fill is 'no', the Event documents will contain foreign keys as
             placeholders for the data. This option is useful for exporting
             copies of the documents.
+        strict_order : bool
```
danielballan: As above

databroker/core.py
```diff
@@ -911,8 +1027,8 @@ class BlueskyEventStream(DataSourceMixin):
     **kwargs :
         Additional keyword arguments are passed through to the base class.
     """
-    container = 'databroker-xarray'
-    name = 'bluesky-event-stream'
+   # container = 'databroker-xarray'
```
danielballan: No need to keep around the old version of this line as a comment.

databroker/core.py
```diff
@@ -993,6 +1116,93 @@ def to_dask(self):
...
+        self.npartitions += len(partitions)
+        return partitions
+
+    def _get_partition(self, partition):
+        return intake.container.base.get_partition(self.url, self.http_args,
+                                             self._source_id, self.container,
+                                             partition)
+
+class RemoteBlueskyEventStream(RemoteXarray):
+    def read_partition(self, partition):
```
danielballan, line 1197: Can you add a comment on what this does? What would happen if `read_partition` weren't defined and we tried on the base classes instead?

databroker/core.py
```diff
@@ -1219,8 +1433,10 @@ def parse_handler_registry(handler_registry):
     return result
 
 
-# TODO Do we actually need this?
+# This is needed. It determines the type of the class that you get on the
```
danielballan:
```suggestion
# This determines the type of the class that you get on the
```

requirements-test.txt
```diff
@@ -10,7 +10,7 @@ mongobox
 ophyd
 pathlib
 pyqt5
-pytest >=3.9
+pytest >=4.4
```
danielballan, line 13: Just curious---which new feature are we employing that requires the version bump?

databroker/_drivers/mongo_normalized.py
danielballan: Git tip: to remove a commit entirely, do

```
git rebase -i HEAD~SOME_NUMBER_OF_COMMITS_BACK
```

and then, in the list of commits that appears in the interactive view, simply delete the line corresponding to the commit you want to remove.

requirements-test.txt
gwbischof: pytest-rerunfailures 8.0 has requirement pytest>=4.4, but you'll have pytest 4.3.1 which is incompatible.

databroker/core.py

databroker/core.py

databroker/core.py
gwbischof: updated

databroker/core.py

databroker/core.py

databroker/core.py
gwbischof: updated

databroker/core.py
gwbischof: updated

databroker/_drivers/mongo_normalized.py
gwbischof: updated

databroker/_drivers/mongo_normalized.py
gwbischof: updated

databroker/_drivers/mongo_normalized.py
gwbischof: updated

databroker/core.py

databroker/_drivers/mongo_embedded.py
gwbischof: updated

dylanmcreynolds:
I have tested in Xi-Cam with my single-run XPCS example (used during the working session last week), and I have not come across any issues.

danielballan:
Lovely. Merging! Thanks for the quick turn-around @dylanmcreynolds and for the tenacious work on this PR, @gwbischof.

PR #470, opened 2019-11-21
Support export for runs with out run_start in Resource.

danielballan:


danielballan:
Tested interactively when exporting from FXI. Would be tricky to write a unit test for this.

PR #472, opened 2019-12-02
FIX: Only yield non-empty parititions.  Also, fix typo.

danielballan:


tacaswell:
:+1: to the patch, the failure looks real, but unrelated?

databroker/core.py
```diff
@@ -1145,7 +1147,7 @@ def read_partition(self, partition):
             except event_model.UnresolvableForeignKeyError as err:
                 # Slow path: This error should only happen if there is an old style
                 # resource document that doesn't have a run_start key.
-                self._partitions[i:i] = self._missing_datum(err.key, self.PARTITION_SIZE)
+                self._partitions[i:i] = self._missing_datum(err.key, partition_size)
```
mrakitin, line 1150: Would not the `[i:i]` slice from the list always return an empty list? Not sure if `self._partitions[i:i]` makes sense.

tacaswell: Because it on the left hand side, this is inserting the list on the right _into_ this list

```python
In [655]: a = [1, 2, 3]

In [656]: a[2:2] = ['a', 'b', 'c']

In [657]: a
Out[657]: [1, 2, 'a', 'b', 'c', 3]

```

mrakitin: Neat! It's an interesting way of inserting stuff. `.insert()` seems to be harder to use for that purpose as it does not unpack the list.

Here is a link which could help future us: https://stackoverflow.com/a/3748092.

danielballan:
The failure is a commonly-seen intermittent issue from the sqlite backend.

PR #473, opened 2019-12-02
filter out duplicate documents in canonical

gwbischof:


databroker/core.py
```diff
@@ -647,7 +650,28 @@ def stream_gen(entry):
...
-        yield from interlace(*streams, strict_order=strict_order)
+
+        # This following code filters out duplicate documents.
+        # This is needed because we dont know which EventStream, that resource
+        # or datum documents belong to, so each stream has these documents.
+        # Without this filter we would get multiple of the same resource and
+        # datum documents.
+        for name, doc in interlace(*streams, strict_order=strict_order):
+
+            if name == 'datum':
```
danielballan: Could this be moved into `interlace` itself to avoid code duplication?

gwbischof: I think its better if interlace is just for interlacing, and not also filtering. I kind of think it belongs in canonical. All of canonical is duplicated. What if I fix the duplication by getting rid of the duplication of canonical?

danielballan: That's a great suggestion. (Thanks for your patience with my slow reply.)

databroker/core.py
```diff
@@ -535,6 +535,82 @@ def documents_to_xarray(*, start_doc, stop_doc, descriptor_docs,
...
+        If fill is 'no', the Event documents will contain foreign keys as
+        placeholders for the data. This option is useful for exporting
+        copies of the documents.
+    strict_order : bool, optional
+        documents are strictly yielded in ascending time order.
+    """
+    run_start_uid = start['uid']
+    history = set()
+
+    # Special case for 'delayed' since it *is* supported in the local mode
```
danielballan: The value `'delayed'` should be allowed in the local-access case. Note the difference between the `canonical` method on remote and the `canonical` method on local in this respect. We could fix this by moving _just_ the `fill` validation step into the two separate methods.

gwbischof: fixed

databroker/tests/test_v2/test_jsonl.py
```diff
@@ -61,4 +62,5 @@ def bundle(request, intake_server, example_data):  # noqa
         raise ValueError
     return types.SimpleNamespace(cat=cat,
                                  uid=uid,
-                                 docs=docs)
+                                 docs=docs,
+                                 remote=remote)
```
danielballan, line 66: :+1: Great solution

PR #474, opened 2019-12-12
SST fix

gwbischof:
Fix for the SST shape, this makes it work for SST, but may break it for other beamlines

### DIFF
--- a/databroker/core.py
+++ b/databroker/core.py
@@ -474,7 +474,7 @@ def documents_to_xarray(*, start_doc, stop_doc, descriptor_docs,
             try:
                 dims = tuple(field_metadata['dims'])
             except KeyError:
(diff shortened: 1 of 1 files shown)


danielballan:
Agreed that this needs careful testing and would be likely to break other beamlines (e.g. CSX) if merged, but this at least characterizes the issue well. It may be that the solution is to apply updates at SST. I have been throwing around the term "database migration", but we could alternatively use the new update feature in suitcase-mongo, which seems like a lighter touch.

danielballan:
This should now be implemented as a "transform" and applied to SST configuration. Let's leave this PR open until that has been done.

gwbischof:
Closing, moved to here: https://github.com/NSLS-II/nslsii/pull/83

PR #478, opened 2019-12-16
WIP: Allow user to pass options to pymongo find.

danielballan:
One option of interest is `no_cursor_timeout`. I believe the cursor timeouts
were encountered at XPD or PDF (by @tacaswell? or @mrakitin?). I just
encountered them while trying to do a long export task at FXI.

I think we ought to leave timeouts on by default---they are there for a reason---
but we clearly need the ability to request a timeout-free cursor for the
server for some cases.

This should be documented and tested before being merged.

### DIFF
--- a/databroker/_drivers/mongo_normalized.py
+++ b/databroker/_drivers/mongo_normalized.py
@@ -58,8 +58,10 @@ class _Entries(collections.abc.Mapping):
             catalog=self.catalog)
 
     def __iter__(self):
+        find_kwargs = {'sort': [('time', pymongo.DESCENDING)]}
+        find_kwargs.update(self.catalog._find_kwargs)
         cursor = self.catalog._run_start_collection.find(
-            self.catalog._query, sort=[('time', pymongo.DESCENDING)])
+            self.catalog._query, **find_kwargs)
         for run_start_doc in cursor:
             yield run_start_doc['uid']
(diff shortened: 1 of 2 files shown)


databroker/_drivers/mongo_normalized.py
```diff
@@ -204,6 +209,7 @@ def __init__(self, metadatastore_db, asset_registry_db, *,
         self._metadatastore_db = mds_db
         self._asset_registry_db = assets_db
         self._query = query or {}
+        lse.f_find_kwargs = find_kwargs or {}
```
mrakitin:
```suggestion
        self._find_kwargs = find_kwargs or {}
```

databroker/_drivers/mongo_normalized.py
```diff
@@ -300,6 +308,7 @@ def search(self, query):
             metadatastore_db=self._metadatastore_db,
             asset_registry_db=self._asset_registry_db,
             query=query,
+            find_kwargs=find_kwargs,
```
mrakitin: Should it be `kwargs`?

danielballan: Yes, good catch.

danielballan:
Fixed and force-pushed. Also added a test.

PR #479, opened 2019-12-16
add transforms to apply temporary fixes to documents

gwbischof:


databroker/v1.py
```diff
@@ -1640,14 +1640,29 @@ def _from_v0_config(config, auto_register, name):
...
                                root_map=root_map,
-                               name=name)
+                               name=name,
+                               transforms=transforms)
+
 
 _mongo_clients = {}  # cache of pymongo.MongoClient instances
 
 
+def _load_transforms(transforms):
```
danielballan: If we reuse the same pattern that we use for handlers, these would have a form like `package.subpackage.module.class`. See https://github.com/bluesky/databroker/blob/521f87c0087084cf7a23f6a2c24df06b512169cd/databroker/core.py#L1418-L1443 for an example of how to parse that.

databroker/v1.py
```diff
@@ -1568,7 +1568,7 @@ def _pretty_print_time(timestamp):
     return '{ago} ({date})'.format(ago=ago, date=dt)
 
 
-def from_config(config, auto_register=True, name=None):
+def from_config(config, auto_register=True, name=None, transforms=None):
```
danielballan: To be consistent with how `handler_registry` is handled, and to avoid making things too complex, I don't think we should allow the user to specify `transforms` here. They should just come `config`.

databroker/v1.py
```diff
@@ -1640,14 +1640,52 @@ def _from_v0_config(config, auto_register, name):
...
+
+     if 'event' in transforms.keys() or 'datum' in transforms.keys():
+        raise NotImplementedError("Transforms for events and datum are not supported.")
+
+     for name in ('start', 'stop', 'resource', 'descriptor'):
+         transform = transforms.get(name)
+         if isinstance(transform, str):
+            module_name, _, class_name = transform.rpartition('.')
+            function = getattr(importlib.import_module(module_name), class_name)
+         else:
```
danielballan: The `parse_handler_registry` function accepts strings (such as we'd get if this came from a YAML config) or the classes themselves. This is different and I think could be potentially very surprising. It accepts strings, but if the user provides a class (or anything other than a string) it silently ignores the user's input and just assigns a pass-through function.

databroker/core.py
```diff
@@ -820,6 +821,11 @@ def __init__(self,
             self.fillers['yes'].handler_registry, inplace=True)
         self.fillers['delayed'] = get_filler(coerce='delayed')
         self._entry = entry
+
+        transformable = ('start', 'stop', 'resource', 'descriptor')
+        if transforms is None:
+            self._transforms = {key: lambda doc: doc for key in transformable}
```
danielballan: I think this needs an `else`.

databroker/v1.py
gwbischof: Thanks for the feedback! Updated

danielballan: Thanks for the update. Looks closer, but it still differs from `parse_handler_registry` in that `parse_handler_registry` accepts the actual class, as in `{'AD_HDF5': MyHandler}`, which allows users to experiment with one-off classes that aren't importable from a module. I think the same thing could be useful for transformers, so it would be good to support `{'start': my_transformer}` as well as `{'start': 'my.packaged.transfomer'}`.

gwbischof: But load_transforms is for loading from a config_file, and you can't have a class instance in the config file. I don't think calling it directly would be useful, because you would have to pass the result to the BlueskyMongoCatalog init, and you can just pass {'start': my_transformer} directly to the init.

databroker/_drivers/msgpack.py
```diff
@@ -106,11 +110,10 @@ def __init__(self, paths, *,
             paths = [paths]
         self.paths = paths
         self._filename_to_mtime = {}
+        print("MSGPACK TRANSFORMS", transforms)
```
tacaswell: debugging print?

databroker/_drivers/jsonl.py
```diff
@@ -102,6 +102,10 @@ def __init__(self, paths, *, handler_registry=None, root_map=None,
             same methods as ``DocumentRouter``.
         query : dict, optional
             Mongo query that filters entries' RunStart documents
+        transforms : dict
```
danielballan: I think this description could be slightly more clear. Let's workshop it a bit. cc @ronpandolfi @dylanmcreynolds @prjemian @mrakitin 

My pitch:

> A dict that maps any subset of the keys {``start``, ``stop``, ``resource``, ``descriptor``} to a function that accepts a document of the corresponding type and returns it, potentially modified. This feature is for patching up erroneous metadata. It is intended for quick, temporary fixes that may later be applied permanently to the data at rest (e.g via a database migration).

Too verbose? I want to give a clear sense of the intended scope of this feature, which could be at risk of being abused.

As a reminder, the immediate motivation for this feature is that we encountered Event Descriptors that recorded the wrong array shape, breaking code that needs to know the correct shape in advance. Ultimately, we want to fix the code that is reporting the incorrect shape and then go back and do a database migration to update the existing documents. But in the short term, and to test the fix, we want this hook to let us patch up the documents on their way out of databroker.

databroker/core.py
```diff
@@ -1443,6 +1460,50 @@ def parse_handler_registry(handler_registry):
     return result
 
 
+def load_transforms(transforms):
```
danielballan: Minor preference for `parse_transforms` for symmetry with `parse_handler_registry`. We may or may not be actually loading anything, depending on the user input.

databroker/core.py
```diff
@@ -1443,6 +1460,50 @@ def parse_handler_registry(handler_registry):
...
+    >>> load_transforms({'descriptor': 'package.module.ClassName'})
+    {'descriptor': <package.module.ClassName>}
+
+    """
+    transformable = {'start', 'stop', 'resource', 'descriptor'}
+
+    if transforms is None:
+        result = {key: lambda doc: doc for key in transformable}
+        return result
+    elif isinstance(transforms, dict):
```
danielballan: I hesitate to even mention this and wouldn't block merging on it, but in the spirit of embracing duck typing, it would be fine to just assume that, if `transforms` is not None, it is a dict-like and proceed. IMO the cost of restricting users to strict `dict` subclasses outweighs the benefit of trying to give a more helpful error message in the edge case where the user accidentally gives us some object here that we cannot use.

databroker/tests/test_v2/test_jsonl.py
```diff
@@ -7,9 +7,11 @@
 import tempfile
 import time
 import types
+import yaml
```
danielballan: No longer used. (My fault for leaving this behind in my test refactor.)

databroker/v1.py
gwbischof: updated

databroker/v1.py
gwbischof: updated

databroker/_drivers/msgpack.py
gwbischof: fixed

databroker/core.py

databroker/tests/test_v2/test_jsonl.py
gwbischof: updated

databroker/core.py
gwbischof: What type of objects that are not dict or None would also be acceptable?

databroker/_drivers/jsonl.py
gwbischof: updated with your suggestion

databroker/core.py
danielballan: User-defined dict-like objects would be OK, such as might arise from custom config-parsing code. Or these...of which some are `dict` subclasses and some are not.

```
In [11]: import collections                                                                                                                                                                   

In [12]: isinstance(collections.OrderedDict(), dict)                                                                                                                                          
Out[12]: True

In [13]: isinstance(collections.ChainMap(), dict)                                                                                                                                             
Out[13]: False

In [14]: isinstance(collections.defaultdict(lambda: (lambda: x)), dict)                                                                                                                       
Out[14]: True

In [19]: isinstance(collections.UserDict(), dict)                                                                                                                                             
Out[19]: False
```

Rather than simply giving up on validation, as I originally suggested here, it may be better to validate against an abstract base class.

```py
In [15]: isinstance(collections.OrderedDict(), collections.abc.Mapping)                                                                                                                       
Out[15]: True

In [16]: isinstance(collections.ChainMap(), collections.abc.Mapping)                                                                                                                          
Out[16]: True

In [17]: isinstance(collections.defaultdict(lambda: (lambda: x)), collections.abc.Mapping)                                                                                                    
Out[17]: True

In [18]: isinstance(collections.UserDict(), collections.abc.Mapping)                                                                                                                          
Out[18]: True
```

Note that none of these types _actually_ subclass the type `collections.abc.Mapping`, but `collections.abc.Mapping` uses [`__subclasshook__`](https://docs.python.org/3/library/abc.html#abc.ABCMeta.__subclasshook__) to check if an object has the expected interface (`__getitem__`, `__iter__`, etc.) and then tells `isinstance` what to return.

databroker/_drivers/jsonl.py
ronpandolfi: Yes, this will be very useful for testing migrations. The description is clear; better type hinting would help. `Dict[str, Callable]`?
Why preclude patching of event and datums?

danielballan: Good question. We're excluding Event and Datum for now because:
- Adding the overhead of a function call (~200 ms) to every row would be a much more significant performance overhead. As is, this overhead will be imposed ~10 times for a typical Run and no more than ~100 times in extreme cases. If we were to include Event and Datum that would grow to ~10^7 at some beamlines we support.
- Making it possible to use this to edit _data_ (i.e. changing the values in Events) feels like a big step up in scope at best and maybe something we flat-out shouldn't support: one could argue that the possibility of a badly-configured transformer changing your data under you  is would be an anti-feature. An alternative: copy the data into a new Run (with a new unique identifier).
- We have rarely (never?) needed to patch up Event or Datum documents in the wild. The only pressing need is to fix up Event Descriptors.

databroker/in_memory.py
```diff
@@ -58,6 +58,9 @@ def __init__(self, *, handler_registry=None, root_map=None,
             same methods as ``DocumentRouter``.
         query : dict, optional
             Mongo query that filters entries' RunStart documents
+        transforms : dict
+            A dict that maps (``start``, ``stop``, ``resource``, ``descriptor``)                                to a function that accepts a document of the corresponding type. This function
```
danielballan: line break problem

danielballan:
Squashed into a small number of commits and rebased on `master`.

databroker/core.py
```diff
@@ -794,40 +794,18 @@ class BlueskyRun(intake.catalog.Catalog):
...
+        A dict that maps any subset of the keys {start, stop, resource, descriptor}
+        to a function that accepts a document of the corresponding type and
+        returns it, potentially modified. This feature is for patching up
+        erroneous metadata. It is intended for quick, temporary fixes that
+        may later be applied permanently to the data at rest
+        (e.g via a database migration).
     **kwargs :
         Additional keyword arguments are passed through to the base class,
         Catalog.
-    handler_registry : dict, optional
```
danielballan, line 800: Note to reviewers: this isn't related to the code changes in this PR. It should have been done in #465 but we missed and just noticed it while updating this docstring.

databroker/v2.py
```diff
@@ -13,13 +13,58 @@ class Broker(Catalog):
     This is a thin wrapper around intake.Catalog.
 
     It includes an accessor the databroker API version 1.
+
+    Parameters
+    ----------
+    handler_registry : dict, optional
```
mrakitin:
```suggestion
    handler_registry: dict, optional
```

databroker/v2.py
```diff
@@ -13,13 +13,58 @@ class Broker(Catalog):
...
+        parameter below.
+
+        str -> str mapping to account for temporarily moved/copied/remounted
+        files.  Any resources which have a ``root`` in ``root_map`` will be
+        loaded using the mapped ``root``.
+    filler_class: type
+        This is Filler by default. It can be a Filler subclass,
+        ``functools.partial(Filler, ...)``, or any class that provides the same
+        methods as ``DocumentRouter``.
+    transforms : dict
```
mrakitin:
```suggestion
    transforms: dict
```

databroker/v2.py
```diff
@@ -13,13 +13,58 @@ class Broker(Catalog):
...
+        A dict that maps any subset of the keys {start, stop, resource, descriptor}
+        to a function that accepts a document of the corresponding type and
+        returns it, potentially modified. This feature is for patching up
+        erroneous metadata. It is intended for quick, temporary fixes that
+        may later be applied permanently to the data at rest
+        (e.g via a database migration).
```
mrakitin:
```suggestion
        (e.g., via a database migration).
```

databroker/v2.py
```diff
@@ -13,13 +13,58 @@ class Broker(Catalog):
...
+        Catalog.
     """
 
     def __init__(self, *, handler_registry=None, root_map=None,
-                 filler_class=event_model.Filler, **kwargs):
+                 filler_class=event_model.Filler, transforms=None, **kwargs):
 
         self._root_map = root_map
         self._filler_class = filler_class
+        self._transforms = parse_transforms(transforms)
```
mrakitin, line 67: Probably it should be mentioned in the docstring that the transforms are in the string form.

danielballan: To be clear they _may_ be in the string form. Both a callable and a (string) import path to a callable are accepted. I agree we should mention that in the docstring.

danielballan:
Thanks for the detailed review, @mrakitin. @gwbischof has addressed all your comments, so I'm going to pull this in. Nice work, @gwbischof.

PR #481, opened 2019-12-20
FIX: support dask < 2

tacaswell:
Trying to stack a list of scalars can cause dask to fail on old
versions of dask (this was fixed in dask 2.0).

### DIFF
--- a/databroker/core.py
+++ b/databroker/core.py
@@ -1329,7 +1329,11 @@ def _transpose(in_data, keys, field):
         for k in keys:
             out[k][j] = dd[k]
     for k in keys:
-        out[k] = dask.array.stack(out[k])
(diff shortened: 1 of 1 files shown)


databroker/core.py
```diff
@@ -1329,7 +1329,10 @@ def _transpose(in_data, keys, field):
         for k in keys:
             out[k][j] = dd[k]
     for k in keys:
-        out[k] = dask.array.stack(out[k])
```
danielballan, line 1332: It might be good to add a code comment here so this doesn't get "simplified" out later.

tacaswell: done.

tacaswell: on the second try.

PR #483, opened 2019-12-30
register tokenize methods to speed up tokenizing

gwbischof:
Loading runs (BlueskyRun._load()) from a catalog (RSOXS) was taking 200-400ms per run. This was due to the dask.tokenize call in intake: https://github.com/intake/intake/blob/master/intake/utils.py#L86

With these changes BlueskyRuns are loading in about 35ms.

If we completely bypass the tokenize call the BlueskyRuns load in about 10ms.


gwbischof:
test_glue was giving an error and saying that numpy needs to be greater than or equal to 1.16.0

databroker/core.py
```diff
@@ -607,6 +608,31 @@ def stream_gen(entry):
...
+class ArgsDict(dict):
+    """
+    This class is made so that we can define our own tokenize method, which
+    greatly speeds up the time to load a BlueskyRun.
+    """
+    ...
+
+# Register a tokenize method for ArgsDict
+@normalize_token.register(ArgsDict)
+def tokenize_dict(args):
```
tacaswell, line 620: Either make these private or make them unique?

databroker/core.py
```diff
@@ -607,6 +608,31 @@ def stream_gen(entry):
     yield ('stop', stop)
 
 
+class ArgsDict(dict):
+    """
+    This class is made so that we can define our own tokenize method, which
+    greatly speeds up the time to load a BlueskyRun.
+    """
+    ...
```
tacaswell, line 616: These classes should define `__slots__`.

databroker/core.py
```diff
@@ -26,6 +26,7 @@
 from .intake_xarray_core.base import DataSourceMixin
 from .intake_xarray_core.xarray_container import RemoteXarray
 from collections import deque
+from dask.base import normalize_token
```
tacaswell, line 29: How hard is our dask dependency currently? I didn't think we had a hard dep, does intake?

Independent of if intake has a hard dependency, it is probably better to hide this import and the registration of the tokenize methods in a `try: .... except ImportError: ...` block.

databroker/core.py
```diff
@@ -607,6 +608,31 @@ def stream_gen(entry):
...
+    """
+    This class is made so that we can define our own tokenize method, which
+    greatly speeds up the time to load a BlueskyRun.
+    """
+    ...
+
+# Register a tokenize method for MetaDict
+@normalize_token.register(MetaDict)
+def tokenize_dict(meta):
+    return (descriptor['uid'] for descriptor in meta['descriptors'])
```
danielballan, line 633: As written, this returns a generator instance. A tuple is probably intended.
```suggestion
    return tuple(descriptor['uid'] for descriptor in meta['descriptors'])
```

tacaswell:
Do we have no tests that covers this code path?

databroker/core.py
```diff
@@ -607,6 +608,31 @@ def stream_gen(entry):
...
+    """
+    This class is made so that we can define our own tokenize method, which
+    greatly speeds up the time to load a BlueskyRun.
+    """
+    ...
+
+# Register a tokenize method for ArgsDict
+@normalize_token.register(ArgsDict)
+def tokenize_dict(args):
+    return (descriptor['uid'] for descriptor in args['metadata']['descriptors'])
```
danielballan, line 621: This token must uniquely define the `ArgsDict` for the entire scheduler. Any key collisions between different `ArgsDict` instances are likely to manifest as very confusing bugs. I think I can imagine a scenario where there are two `ArgsDict` instances with the same descriptors but, say, different `self.fillers` or something.

From the discussion on Slack, it looks like the costly step is stringify-ing the descriptors. What sort of performance do you see if you make `args` a plain `dict` but change

```diff
-                metadata={'descriptors': descriptors,
-                          'resources': self._resources})
+                metadata=MetaDict({'descriptors': descriptors,
+                          'resources': self._resources}))
```

If that gives us _almost_ the same speedup, it would be significantly safer and a lighter touch.

requirements.txt
```diff
@@ -9,7 +9,7 @@ jsonschema
 mongoquery
 msgpack
 msgpack-numpy  # to be split into extras or a separate package
-numpy
+numpy >=1.16.0
```
danielballan, line 12: Is this still relevant?

databroker/core.py
danielballan: Intake has a [hard dependency on `dask[base]`](https://github.com/intake/intake/blob/34dd38089ccb805759508659b5e8475034df2f0c/requirements.txt#L4), which pulls in a few things but notable does *not* pull in heavy stuff like pandas.

Note that it is possible to load data as documents without _using_ dask, but databroker does now require that the user _install_ it.

requirements.txt
danielballan: I belatedly noticed your [comment](https://github.com/bluesky/databroker/pull/483#issuecomment-569947385) about this. I am skeptical that our very experimental glue integration is a good enough reason to pin numpy. @tacaswell, would you suggest pinning it in the test deps?

danielballan: I belatedly noticed your comment about this. I am skeptical that our very experimental glue integration is a good enough reason to pin numpy. @tacaswell, would you suggest pinning it in the test deps?

danielballan: I belatedly noticed your comment about this. I am skeptical that our very experimental glue integration is a good enough reason to pin numpy. @tacaswell, would you suggest pinning it in the test deps?

danielballan: ...as proposed here https://github.com/bluesky/databroker/pull/484

danielballan: ...as proposed here https://github.com/bluesky/databroker/pull/484

tacaswell: Pinning it as test dep seems right to me.

databroker/core.py
tacaswell: if we are going to directly import dask, we should also have a hard dependency on what we use (or bury the imports, I'm cool either way).

databroker/core.py
tacaswell: We could also wrap the start document in a similar subclass.   At this point we have seen the slow down on big descriptors, but it is only a matter of time until we see the slow down on the (up to 16MB) of random user supplied input in the start document :smiling_imp: 

Similarly, we could wrap just the descriptors to source a token that is `(start_uid, desc_uid)`.

databroker/core.py
danielballan: My wording was poor. Intake has a hard dependency on `dask[base]` but _also_ databroker has a hard dependency on `dask[array,base]` already declared in current `master`.

dylanmcreynolds:
Tracking this over as a Pilot repository issue at https://github.com/lightsources/project-management/issues/87

databroker/core.py
tacaswell: ah, fair enough.

ronpandolfi:
@gwbischof This looked promising. Intake's use of `dask.base.tokenize` seems responsible for a vast portion of the slowness I'm seeing. Why was this closed?

Edit: Nvm, I found https://github.com/bluesky/databroker/pull/489

PR #487, opened 2020-01-09
transforms bug fix

gwbischof:
transform functions were getting called twice. the `deepcopies` were added because we discovered that the transforms were modifying documents cached by the drivers.


danielballan:
I wonder if we can check to be sure we caught them all and defend against accidentally breaking this in the future. In the example transformer used by the test, could we first check that the key to be added does not already exist, and then add it?

PR #489, opened 2020-01-10
BlueskyRun._load() optimization

gwbischof:
This was a joint effort from: @tacaswell, @danielballan, and myself. 

Loading runs (BlueskyRun._load()) from a catalog (RSOXS) was taking 200-400ms per run. This was due to the `dask.tokenize` call in intake: https://github.com/intake/intake/blob/master/intake/utils.py#L86

With these changes `BlueskyRuns` are loading in about 20ms

We got rid of some `deepcopy` calls in load, and decided that transforms will be responsible for not mutating the passed in document.

We converted all of the dictionaries in `BlueskyRun` to `OrderedDicts`, so `tokenize` doesn't sort them. `tokenize` spent the majority of its execution time sorting these dictionaries.

Also we register a tokenize method for all Bluesky docs so that they can quickly be tokenized.

We plan to make a PR into intake that helps speed up `tokenize`, while waiting for that PR we made the changes in databroker by creating the class `StreamEntry`, that inherits from `LocalCatalogEntry`, and overrides `__getstate__` of `DictSerialiseMixin`.

### DIFF
--- a/databroker/core.py
+++ b/databroker/core.py
@@ -962,6 +962,11 @@ class BlueskyRun(intake.catalog.Catalog):
         descriptors_by_name = collections.defaultdict(list)
         for doc in self._descriptors:
             descriptors_by_name[doc.get('name', 'primary')].append(doc)
+        # We employ OrderedDict in several places in this loop. The motivation
+        # is to speed up dask tokenization. When task tokenizes a plain dict,
+        # it sorted the keys, and it turns out that this sort operation
+        # dominates the call time, even for very small dicts. Using an
+        # OrderedDict steers dask toward a different and faster tokenization.
         for stream_name, descriptors in descriptors_by_name.items():
             meta=OrderedDict({'start': self.metadata['start'],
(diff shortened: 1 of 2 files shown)


databroker/core.py
```diff
@@ -26,6 +26,69 @@
 from .intake_xarray_core.base import DataSourceMixin
 from .intake_xarray_core.xarray_container import RemoteXarray
 from collections import deque
+from typing import Iterator, Mapping, TypeVar
+
+K = TypeVar("K")
+V = TypeVar("V")
+
+
+class Frozen(Mapping[K, V]):
```
tacaswell: If we are going to go down the typing route, we can pin this a bit more to `K` always being `str` and `V` might as well be `Any`.

tacaswell:
We will eventually have to think about how to handle the `*Pages`, but not today.

stuartcampbell:
@gwbischof Just wanted to say a thank you for writing a good clear PR - awesome work.

PR #502, opened 2020-01-16
Make BlueskyRun._entries a LazyMap

gwbischof:


danielballan:
@gwbischof opened #505 with a slightly different approach. Having found that the performance is comparable and the readability is better in #505, we are abandoning this one.

PR #494, opened 2020-01-14
Cache Entries and DataSources better.

danielballan:
~This is stale work (from November) which may be the complete wrong approach. I am opening this PR to make the work more easily visible.~

This has evolved and now seems like the correct approach. Example with DEBUG logging turned up:

When a file-based ("in-memory") catalog is instantiated, it creates all of its entries greedily. We see that logged here.
```py
In [1]: catalog = BlueskyMsgpackCatalog(f'{directory}/*.msgpack')                                                                                
Created Entry named '35fb6cca-846b-4bbf-9f68-8b01f9c75503'
Created Entry named '7711ad85-43d1-4777-b1f9-e7d581a1057c'
```

When we access a given run for the first time, BlueskyRun (or a sublcass of it, in this case) is instantiated and, as part of that, the data is loaded.

```py
In [2]: catalog['7711ad'];                                                                                                                                                                    
Loaded BlueskyRunFromGenerator named '7711ad85-43d1-4777-b1f9-e7d581a1057c'
Created BlueskyRunFromGenerator named '7711ad85-43d1-4777-b1f9-e7d581a1057c'
```

That instance is cached on the Entry object and retrieved whenever we access it again.

```py
In [3]: catalog['7711ad'];                                                                                                                                                                    
Entry cache found BlueskyRunFromGenerator named '7711ad85-43d1-4777-b1f9-e7d581a1057c'

In [4]: catalog['7711ad'];                                                                                                                                                                    
Entry cache found BlueskyRunFromGenerator named '7711ad85-43d1-4777-b1f9-e7d581a1057c'
```

Accessing a stream causes `_load` to be called again.

```py
In [5]: catalog['7711ad'].primary;                                                                                                                                                            
Entry cache found BlueskyRunFromGenerator named '7711ad85-43d1-4777-b1f9-e7d581a1057c'
Loaded BlueskyRunFromGenerator named '7711ad85-43d1-4777-b1f9-e7d581a1057c'

In [6]: catalog['7711ad'].primary;                                                                                                                                                            
Entry cache found BlueskyRunFromGenerator named '7711ad85-43d1-4777-b1f9-e7d581a1057c'
Loaded BlueskyRunFromGenerator named '7711ad85-43d1-4777-b1f9-e7d581a1057c'
```

Unless it's called successively less than `catalog.ttl` seconds apart (1 second by default).

```py
In [7]: catalog['7711ad'].primary;                                                                                                                                                            
[...]

PR #505, opened 2020-01-16
Make BlueskyRun._entries a LazyMap with partial

gwbischof:


databroker/core.py
```diff
@@ -1024,18 +1039,12 @@ def _load(self):
...
-                metadata=metadata,
-                catalog_dir=None,
-                getenv=True,
-                getshell=True,
-                catalog=self)
+
+            entries[stream_name] = functools.partial(wrapper, stream_name,
+                                                     descriptors, metadata, args)
+
+        self._entries = LazyMap(entries)
```
danielballan: We can override

```py
def _make_entries_container(self):
    return LazyMap()
```

to make `self._entires` a `LazyMap` to begin with.

Re-setting an attribute like this during the lifecycle of an object can create unexpected side-effects. As discussed with @gwbischof and @tacaswell toward the end of the day yesterday, I think we need to relent on making `LazyMap` mutable. However, we can stay true to the spirit of "make it act like a dict" and "keep cognitive load for the recipient of the dict low" by making a special setter method

```py
lazy_map = LazyMap(a=lambda: 1, b=lambda: 2)
lazy_map.add(c=lambda: 3)
```

The rationale for making a new method is to avoid the ambiguity of "Is it `lazy_map['c'] = lambda: 3` or `lazy_map['c'] = 3`?" If someone is using the special method `.add(...)` they know they are doing something outside of ordinary dict behavior.

We do not need to support updating existing values, and that would add some friction, so I think that trying to `add` a key that exists should blow up.

danielballan:
Remaining to be done here:
- [x] Make `BlueskyEventStream` check for a RunStop doc on `_load` if it doesn't already have one. (Credit to @gwbischof for noticing this.)
- [x] Fix v1, which is expecting to find `run.metadata['descriptors']`. I think removing descriptors from `run.metadata` is unambigiously a good change. In the future we might avoid pulling the descriptors at `_load` time, leaving that up to `BlueskyEventStream`, and gain significant performance wins. But for now, we are pulling them anyway in order to determine the stream names, so we might as well stick them onto `run._descriptors` so that v1 can get them easily to put them in `header.descriptors`.

databroker/core.py
```diff
@@ -1218,12 +1234,19 @@ def __init__(self,
 
         self._run_stop_doc = metadata['stop']
         self._run_start_doc = metadata['start']
-        self._descriptors = [descriptor for descriptor in metadata['descriptors']
-                             if descriptor.get('name') == self._stream_name]
-        # Should figure out a way so that self._resources doesn't have to be
-        # all of the Run's resources.
-        self._resources = metadata['resources']
         self._partitions = None
+        self._load()
```
danielballan: Note to selves: we think this shouldn't be necessary. Make sure it's removed (or understood) before merging.

databroker/utils.py
```diff
@@ -414,3 +414,14 @@ def __iter__(self):
     def __contains__(self, k):
         # make sure checking 'in' does not trigger evaluation
         return k in self.__mapping
+
+    def add(self, *args, **kwargs):
+        dictionary = dict(*args, **kwargs)
+        wrap = self.__Wrapper
+        with self.__lock:
+            for key in dictionary:
```
danielballan: This might be a nice place to use set logic rather than nested for loops.

databroker/core.py
```diff
@@ -161,8 +162,8 @@ def _create_open_args(self, user_parameters):
         open_args['entry'] = self
         return plugin, open_args
 
-    def cache_clear(self):
-        self.__cache.clear()
+    def _get_cached(self, **kwargs):
```
danielballan: I think this change might be accidental. We don't need `_get_cached` (that was from an early draft of #494) and I think we _do_ want `cache_clear`.

databroker/utils.py
```diff
@@ -414,3 +414,14 @@ def __iter__(self):
     def __contains__(self, k):
         # make sure checking 'in' does not trigger evaluation
         return k in self.__mapping
+
+    def add(self, *args, **kwargs):
+        dictionary = dict(*args, **kwargs)
+        wrap = self.__Wrapper
+        with self.__lock:
+            intersection = set(dictionary.keys()).intersection(self.__mapping.keys())
```
danielballan: Iterating over a dict yields its keys, so this may be tightened to save two method calls.
```suggestion
            intersection = set(dictionary).intersection(self.__mapping)
```

databroker/core.py
```diff
@@ -1273,13 +1288,29 @@ def __init__(self,
...
+                             if descriptor.get('name') == self._stream_name]
+        # TODO Should figure out a way so that self._resources doesn't have to
+        # be all of the Run's resources.
+        self._resources = [Resource(self._transforms['resource'](resource))
+                           for resource in self._get_resources()]
+        
+        # get_run_stop() may return None if the document was never created due
+        # to a critical failure or simply not yet emitted during a Run that is
+        # still in progress. If it returns None, pass that through.
+        stop = self._get_run_stop()
```
danielballan: Can we first check if `self._run_stop_doc is None`? Unlike descriptors and resources, once we have a stop document, we don't need to check back to see if any more are coming.

databroker/core.py

databroker/utils.py
gwbischof: updated

databroker/core.py
gwbischof: fixed that one already

databroker/utils.py
gwbischof: updated

databroker/core.py
gwbischof: I copied this from the Run, should I change it in both places?

danielballan: Yes, thanks.

danielballan:
2/3 builds passed. The one failed build has one common intermittent sqlite cursor timeout failure.

PR #511, opened 2020-01-21
Include 'descriptors' in stream metadata.

danielballan:
I think the metadata should be initialized at init time. Is there
any harm in calling `_load` there?

### DIFF
--- a/databroker/core.py
+++ b/databroker/core.py
@@ -1299,11 +1299,15 @@ class BlueskyEventStream(DataSourceMixin):
     def _load(self):
         # TODO Add driver API to fetch only the descriptors of interest instead
         # of fetching all of them and then filtering.
-        self._descriptors = [Descriptor(self._transforms['descriptor'](descriptor))
-                             for descriptor in self._get_event_descriptors()
-                             if descriptor.get('name') == self._stream_name]
+        self._descriptors = d = [Descriptor(self._transforms['descriptor'](descriptor))
+                                 for descriptor in self._get_event_descriptors()
+                                 if descriptor.get('name') == self._stream_name]
+        self.metadata.update({'descriptors': d})
         # TODO Should figure out a way so that self._resources doesn't have to
         # be all of the Run's resources.
+        # TDOO Should we expose this in metadata as well? Since
+        # _get_resources() only discovers new-style Resources that have a
+        # run_start in them, leave it private for now.
         self._resources = [Resource(self._transforms['resource'](resource))
                            for resource in self._get_resources()]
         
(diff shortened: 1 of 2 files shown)


danielballan:
This resolves an issue in XI-cam, which relies on `'descriptors'` being present here: https://github.com/synchrotrons/Xi-cam.CatalogViewer/blob/928888b5cff61482616dfcf3826247fa69e04a85/xicam/catalog_viewer/__init__.py#L43

databroker/core.py
```diff
@@ -1312,10 +1317,9 @@ def _load(self):
         # still in progress. If it returns None, pass that through.
         if self._run_stop_doc is None:
             stop = self._get_run_stop()
-            if stop is None:
-                self._run_stop_doc = stop
```
danielballan, line 1316: If we are here, we have already checked that `stop is None` and `self._run_stop_doc is None`, so setting it here has no effect.

databroker/core.py
```diff
@@ -1295,15 +1295,20 @@ def __init__(self,
             "Created %s for stream name %r",
             self.__class__.__name__,
             self._stream_name)
+        self._load()
```
gwbischof: Does it work without this load?

danielballan: I first wrote the tests without this load (commit 877e63a) and they fail because `_load_partition` hasn't been called yet. Maybe the proper place to put this is in `_load_metadata`.

danielballan: Indeed. It looks like `_load` is a Catalog concept only. DataSources implement `_load_metadata`.

danielballan: ...That's not quite right. But `_load_metadata` _is_ a thing and this seems like the right place to use it.

gwbischof: I think we need to also change the other calls to _load in BlueskyEventStream to _load_metadata.

gwbischof: there is another _load call here: https://github.com/bluesky/databroker/pull/511/files#diff-369696e385332d7139756bd5f292bd03R1337

danielballan: Digging through the base classes and intake-xarray a bit, I think the right thing to do for now is to invent our own method, called once at `__init__` time. Eventually we'll have to deal with reloading to check for new documents, something which is not currently handled, and at that point a separate method may turn out to be useful. It looks like `_load_metadata` is only meant to be called once, and `_load` is not in general a DataSource concept.

danielballan:
After much thrashing, I think this is a reasonable approach for now.

danielballan:
Thanks for the fix-up, @gwbischof. We'll await confirmation from @ihumphrey that this puts Xi-cam in a happy state.

ihumphrey:
@danielballan @gwbischof This change does make Xi-cam happy :smile_cat:

PR #512, opened 2020-01-22
DOC: correct the archiver doc

mrakitin:
During the use of the example for the archiver appliance at ISS, noticed the example is wrong (should be `hdr.table(...)` rather than `db.table(...)`).

Also, <kbd>G</kbd>+<kbd>Q</kbd>'d the blocks of text.


mrakitin:
Thank you, Dan!

PR #513, opened 2020-01-23
Define __slots__ on Document.

danielballan:
See https://docs.python.org/3/reference/datamodel.html#slots

### DIFF
--- a/databroker/core.py
+++ b/databroker/core.py
@@ -64,6 +64,7 @@ class Document(dict):
     consumers that expect an object that satisfies isinstance(obj, dict).
     This implementation detail may change in the future.
     """
+    __slots__ = ()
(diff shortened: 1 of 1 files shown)


danielballan:
The main motivation is to make `Documents` have a cheaper memory footprint. A secondary motivation is to disallow monkey-patching them, just as `dict` disallows monkey-patching:

```py
In [1]: d = {'a': 1}                                                                                                                                                                          

In [2]: d.custom_attribute = 'b'                                                                                                                                                              
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-2-edf89592f27a> in <module>
----> 1 d.custom_attribute = 'b'

AttributeError: 'dict' object has no attribute 'custom_attribute'
```

Issue #516, opened 2020-01-29
hdr.table() returns an empty dataframe for v1


mrakitin:
Observed yesterday at CSX.

Databroker 1.0.0b6:
![20200128_182046](https://user-images.githubusercontent.com/13209176/73358386-809e9880-426c-11ea-8447-d3b71291e6a6.jpg)

Databroker 0.13.0:
![20200128_182152](https://user-images.githubusercontent.com/13209176/73358431-9318d200-426c-11ea-9270-8a043004e44b.jpg)

gwbischof:
I am working on debugging this. When I use databroker 1.0 the run only shows 12 documents. When I use databroker 0.13 it shows 889 documents.

gwbischof:
Also with databroker v1.0.0b3 there are 543 documents for this run

gwbischof:
I checked a different scan and the number of documents for the different versions of databroker matched.

gwbischof:
@mrakitin suggested that we add tests to make sure we get the same data for v0,v1,v2 brokers.

gwbischof:
I think the problem is that each time read_partition is called. self._partitions is None, and the partitions are reloaded each time. When filling there is a UnresolvableForeignKeyError, this inserts a new partitions into self._partitions. But the next call to read_partition self._partitions is None again and the partition that was inserted due to the UnresolvableForeignKeyError is missing.

gwbischof:
This is the problem, I removed the line of code that inserts the missing datum into BlueskyEventStream._partitions, and I got the correct number of events. What is happening is that when we read partition 0, we call load_partitions because self._partitions is None, and then return partition 0. Then we read_partition 1, self._partitions in None again, so we call load_partitions. We try to return partition 1, but we find that there are missing datum, so we insert the missing datum into self._partitions at index 1. This shifts the old partition 1 to partition 2. Then we return the new partition 1. Then we try to get the next partition (partition 2). self._partitions is None again, so we call load_partitions which gives us partition 0 and partition 1.  When we try to return partition 2, we find that there is no partition 2, so we return an empty list. If self._partitions kept its value this would work correctly.

PR #517, opened 2020-01-30
fix LocalCatalogEntry.get() so that a new object isn't created each time it is accessed

gwbischof:
fixes https://github.com/bluesky/databroker/issues/516

### DIFF
--- a/databroker/core.py
+++ b/databroker/core.py
@@ -184,6 +184,9 @@ class StreamEntry(intake.catalog.local.LocalCatalogEntry):
     """
     This is a temporary fix that is being proposed to include in intake.
     """
+    def __init__(self, *args, **kwargs):
+        self._data_source = None
+        super().__init__(*args, **kwargs)
 
     def __getstate__(self):
         args = [arg.__getstate__() if isinstance(arg, DictSerialiseMixin)
@@ -193,6 +196,19 @@ class StreamEntry(intake.catalog.local.LocalCatalogEntry):
                               for k, arg in self._captured_init_kwargs.items()})
         return OrderedDict(cls=self.classname, args=args, kwargs=kwargs)
 
+    def get(self, **user_parameters):
+        if self._data_source is None:
+            """Instantiate the DataSource for the given parameters"""
+            plugin, open_args = self._create_open_args(user_parameters)
+            data_source = plugin(**open_args)
+            data_source.catalog_object = self._catalog
+            data_source.name = self.name
+            data_source.description = self._description
+            data_source.cat = self._catalog
+            self._data_source = data_source
+            return data_source
+        else:
+            return self._data_source
 
 def tail(filename, n=1, bsize=2048):
     """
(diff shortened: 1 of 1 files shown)


gwbischof:
What is happening is that when we read partition 0, we call load_partitions because self._partitions is None, and then return partition 0. Then we read_partition 1, self._partitions in None again, so we call load_partitions. We try to return partition 1, but we find that there are missing datum, so we insert the missing datum into self._partitions at index 1. This shifts the old partition 1 to partition 2. Then we return the new partition 1. Then we try to get the next partition (partition 2). self._partitions is None again, so we call load_partitions which gives us partition 0 and partition 1. When we try to return partition 2, we find that there is no partition 2, so we return an empty list. If self._partitions kept its value this would work correctly.

databroker/core.py
```diff
@@ -193,6 +196,19 @@ def __getstate__(self):
                               for k, arg in self._captured_init_kwargs.items()})
         return OrderedDict(cls=self.classname, args=args, kwargs=kwargs)
 
+    def get(self, **user_parameters):
+        if self._data_source is None:
```
danielballan: The second time `get` is called (and thereafter) the `user_parameters` will be ignored and we will get the result from whatever `user_parameters` were given the first time `get` was called.

The `Entry` class just above attempts to do this caching in a way that accounts for the `user_parameters` a.k.a `kwargs`. Would it work to make `StreamEntry` inherit from `Entry` and reuse its mechanism?

gwbischof: I dont think the LRU cache will work for this.
I think you want to permanently cache the streams or else you can have the same issue again.

databroker/core.py
```diff
@@ -193,6 +196,21 @@ def __getstate__(self):
...
+            data_source = plugin(**open_args)
+            data_source.catalog_object = self._catalog
+            data_source.name = self.name
+            data_source.description = self._description
+            data_source.cat = self._catalog
+            self._data_source = data_source
+            return data_source
+        else:
+            _, open_args = self._create_open_args(user_parameters)
+            self._data_source.__init__(**open_args)
```
tacaswell: If we are going to re-run `__init__` why not just start from scratch?

gwbischof: Updated this, it doesn't work this way any longer.

databroker/core.py
gwbischof: Updated

gwbischof:
I just tested this again at CSX with the latest changes, and it is working correctly.

danielballan:
Rebased and force-pushed to remove presumably accidentally-committed file `bluejeans-v2_2.0.0-189_amd64.deb`.

Notes from call:
* This reuses the same approach that was we use with `Entry` and `BlueskyRun` to avoid instantiating the same thing many times.
* We use a `dict` as the cache instead of `LRUCache` because it is critical that the same instance be reused. It's hard to see how that cache could grow beyond length-1.
* The source of the issue is fragility around how partitions interact with Resources in the case when Resources do not have a `run_start` field. We will revisit this logic once we can assume that _all_ Resources know their RunStart.

PR #519, opened 2020-02-12
Various fixes from acceptance testing

danielballan:
See commit messages for details. Changes relate to:

* Handling more edge cases in the test utility ``compare``
* Compatibility with v0 when the Run Stop document is missing
* Handling ragged arrays

### DIFF
--- a/databroker/core.py
+++ b/databroker/core.py
@@ -1590,6 +1590,23 @@ def _transpose(in_data, keys, field):
             # out as plain numpy arrays. In the future we might make them dask
             # arrays with manual chunks.
             out[k] = numpy.asarray(out[k])
+        except ValueError as err:
+            # TEMPORARY EMERGENCY FALLBACK
+            # If environment variable is set to anything but 0, work around
+            # dask and return a numpy array.
+            switch = int(os.environ.get('DATABROKER_ARRAY_FALLBACK'))
+            if switch:
+                out[k] = numpy.asarray(out[k])
+                warnings.warn(
+                    f"Creating a dask array raised an error. Because the "
+                    f"environment variable DATABROKER_ARRAY_FALLBACK was set"
+                    f"to {switch} we have caught the error and fallen "
+                    f"back to returning a numpy array instead. This may be "
+                    f"very slow. The underlying issue should be resolved. The "
+                    f"error was {err!r}.")
+            else:
+                raise
+
     return out
 
 
--- a/databroker/core.py
+++ b/databroker/core.py
@@ -1599,7 +1599,7 @@ def _transpose(in_data, keys, field):
                 out[k] = numpy.asarray(out[k])
                 warnings.warn(
                     f"Creating a dask array raised an error. Because the "
-                    f"environment variable DATABROKER_ARRAY_FALLBACK was set"
+                    f"environment variable DATABROKER_ARRAY_FALLBACK was set "
                     f"to {switch} we have caught the error and fallen "
                     f"back to returning a numpy array instead. This may be "
                     f"very slow. The underlying issue should be resolved. The "
--- a/databroker/core.py
+++ b/databroker/core.py
@@ -1590,7 +1590,6 @@ def _transpose(in_data, keys, field):
             # out as plain numpy arrays. In the future we might make them dask
             # arrays with manual chunks.
             out[k] = numpy.asarray(out[k])
-            pass
     return out
 
 
--- a/databroker/core.py
+++ b/databroker/core.py
@@ -1578,11 +1578,19 @@ def _transpose(in_data, keys, field):
         for k in keys:
             out[k][j] = dd[k]
     for k in keys:
-        # compatibility with dask < 2
-        if hasattr(out[k][0], 'shape'):
-            out[k] = dask.array.stack(out[k])
-        else:
-            out[k] = dask.array.array(out[k])
+        try:
+            # compatibility with dask < 2
+            if hasattr(out[k][0], 'shape'):
+                out[k] = dask.array.stack(out[k])
+            else:
+                out[k] = dask.array.array(out[k])
+        except NotImplementedError:
+            # There are data structured that dask auto-chunking cannot handle,
+            # such as an list of list of variable length. For now, let these go
+            # out as plain numpy arrays. In the future we might make them dask
+            # arrays with manual chunks.
+            out[k] = numpy.asarray(out[k])
+            pass
     return out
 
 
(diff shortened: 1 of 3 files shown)


danielballan:
CI failures are random failures.

databroker/core.py
```diff
@@ -1578,11 +1578,19 @@ def _transpose(in_data, keys, field):
...
-            out[k] = dask.array.stack(out[k])
-        else:
-            out[k] = dask.array.array(out[k])
+        try:
+            # compatibility with dask < 2
+            if hasattr(out[k][0], 'shape'):
+                out[k] = dask.array.stack(out[k])
+            else:
+                out[k] = dask.array.array(out[k])
+        except NotImplementedError:
```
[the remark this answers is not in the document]
danielballan: Interesting. The case that brought me here was ragged arrays, which numpy/dask have no real solution for and thus raise `NotImplementedError`. If we hit the code path you linked, we probably have a lying descriptor and should fix it with a transformer rather than pulling (potentially large) data into memory as a numpy array.

databroker/core.py
```diff
@@ -1578,11 +1578,19 @@ def _transpose(in_data, keys, field):
...
+            # There are data structured that dask auto-chunking cannot handle,
+            # such as an list of list of variable length. For now, let these go
+            # out as plain numpy arrays. In the future we might make them dask
+            # arrays with manual chunks.
+            out[k] = numpy.asarray(out[k])
+            pass
```
gwbischof: Do we need this pass statement? What is its purpose?

databroker/core.py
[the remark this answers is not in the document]
danielballan: OTOH, it might be clever to add an escape hatch here in case we are overwhelmed by transform-related issues during the ~reply~ deployment. Something like, "If the environment variable `DATABROKER_ARRAY_FALLBACK` is set, catch `ValueError`, warn that things will be slow, and return a numpy array rather than failing." I don't think that should be on by _default_ though.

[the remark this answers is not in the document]
gwbischof: Won't we hit the ValueError when we call dask.array.stack() with a ragged array?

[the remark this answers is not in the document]
danielballan: No, numpy treats a ragged array as an array of "objects" (which happen to be lists or variable length) and thus we hit https://github.com/dask/dask/blob/5f61f7f79b861be5ffdbf59950889166628a8fa6/dask/array/core.py#L2532-L2536

databroker/core.py
danielballan: Oops, left over from iterating on this. Pushed 541c93f7 to remove it.

databroker/core.py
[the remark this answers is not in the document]
danielballan: Added emergency fallback in 6a04e03f

databroker/tests/test_v2/generic.py
```diff
@@ -28,26 +28,46 @@ def compare(a, b):
     a_indexed = {}
     b_indexed = {}
     for name, doc in a:
+        if name == 'resource':
+            # Check for an extraneous duplicate key in old documents.
+            if 'id' in doc:
+                assert doc['id'] == doc['uid']
+                doc = doc.copy()
+                doc.pop('id')
         if name == 'datum':
```
gwbischof: I think this should be an elif

gwbischof, line 37: I think this should be an elif

databroker/tests/test_v2/generic.py
```diff
@@ -28,26 +28,46 @@ def compare(a, b):
...
         else:
             a_indexed[(name, doc['uid'])] = doc
     for name, doc in b:
+        if name == 'resource':
+            # Check for an extraneous duplicate key in old documents.
+            if 'id' in doc:
+                assert doc['id'] == doc['uid']
+                doc = doc.copy()
+                doc.pop('id')
         if name == 'datum':
```
gwbischof, line 51: I think this should be an elif

databroker/tests/test_v2/generic.py
danielballan: Good catch. Pushed.

databroker/tests/test_v2/generic.py
danielballan: Ditto.

databroker/tests/test_v2/generic.py
```diff
@@ -28,26 +28,46 @@ def compare(a, b):
     a_indexed = {}
     b_indexed = {}
     for name, doc in a:
-        if name == 'datum':
+        if name == 'resource':
+            # Check for an extraneous duplicate key in old documents.
+            if 'id' in doc:
+                assert doc['id'] == doc['uid']
+                doc = doc.copy()
+                doc.pop('id')
```
gwbischof, line 36: The resource isn't being added to a_indexed, same below.

databroker/tests/test_v2/generic.py
```diff
@@ -28,26 +28,46 @@ def compare(a, b):
...
+        elif name == 'datum':
             a_indexed[('datum', doc['datum_id'])] = doc
+        # v0 yields {'_name": 'RunStop'} is the stop doc is missing; v2 yields None.
```
gwbischof: typo: is -> if. same below

databroker/tests/test_v2/generic.py
danielballan: Oops, actually it was intentionally an `if`. First we handle the special case of `resource`; then we go on to handle the various ways that documents contain a unique ID (or not). We want `resource` to hit the `else` path of that second part so that it is added to `a_indexed`.

danielballan: Force-pushed to remove commit 305182ab688ecd8ad8e997f2f997a5ba174c0426 and revert to `if`.

databroker/tests/test_v2/generic.py
danielballan: See https://github.com/bluesky/databroker/pull/519#discussion_r380917557

databroker/tests/test_v2/generic.py
danielballan: See https://github.com/bluesky/databroker/pull/519#discussion_r380917557

databroker/tests/test_v2/generic.py
danielballan: Fixed.

Issue #520, opened 2020-02-19
interlace fails when events have the same timestamp + index


gwbischof:
```
v0_cat = <databroker._core.Broker object at 0x7f31fb5f85d0>, v1_cat = <databroker.v1.Broker object at 0x7f31fad23a50>
v2_cat = <Intake catalog: None>, run = -7, beamline = 'QAS'

    def test_compare_canonical(v0_cat, v1_cat, v2_cat, run, beamline):
>       compare(v1_cat[run].documents(), v0_cat[run].documents(), remove_ok=True)

databroker-nsls2-tests/tests/generic.py:89: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
databroker-nsls2-tests/tests/generic.py:56: in compare
    run_a = index(normalize(a))
databroker-nsls2-tests/tests/generic.py:31: in index
    for name, doc in run:
databroker-nsls2-tests/tests/generic.py:14: in normalize
    for name, doc in gen:
../databroker/databroker/v1.py:1176: in documents
    for payload in gen:
../databroker/databroker/v1.py:432: in get_documents
    strict_order=True):
../databroker/databroker/core.py:1151: in canonical
    strict_order=strict_order)
../databroker/databroker/core.py:782: in canonical
    for name, doc in interlace(*streams, strict_order=strict_order):
../databroker/databroker/core.py:450: in interlace
    get_next(i)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

index = 3

    def get_next(index):
        while True:
            try:
                name, doc = next(iters[index])
            except StopIteration:
                return
            if name == 'event':
                heapq.heappush(heap, (doc['time'], index, ('event', doc)))
                return
            if name == 'event_page':
                if strict_order:
                    for event in event_model.unpack_event_page(doc):
>                       heapq.heappush(heap, (event['time'], index, ('event', event)))
[...]

PR #521, opened 2020-02-19
interlace fix

gwbischof:
Fixes https://github.com/bluesky/databroker/issues/520

### DIFF
--- a/databroker/core.py
+++ b/databroker/core.py
@@ -355,11 +355,11 @@ def interlace_event_pages(*gens):
             val = next(iters[indx])
         except StopIteration:
             return
-        heapq.heappush(heap, (val['time'][0], indx, val))
+        heapq.heappush(heap, (val['time'][0], indx, val['uid'][0], val))
     for i in range(len(iters)):
         safe_next(i)
     while heap:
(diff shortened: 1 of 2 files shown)


tacaswell:
I think the `uid` should go before the indx as the `uid` is something from the data where as the index is a internal thing that only needs to be there so we can push the next element from the generator into the heap.

tacaswell:
I don't think `interlace_event_page_chunks` actually works as intended.  It comes out ordered (so the tests pass), but many too few chunks get emitted.  I pushed a test that used to fail and simplified `interlace_event_page_chunks` to use `interlace_event_pages` internally.

gwbischof:
> I don't think `interlace_event_page_chunks` actually works as intended. It comes out ordered (so the tests pass), but many too few chunks get emitted. I pushed a test that used to fail and simplified `interlace_event_page_chunks` to use `interlace_event_pages` internally.

Nice simplification and update to the test! Your updated test is passing locally.

danielballan:
Note this PR targeted the wrong branch (fortunately, a disposable one, not a persistent back-port branch or anything like that). See #521 for the intended PR into `master`.

PR #523, opened 2020-02-20
Document mutability cleanup

danielballan:
The tests that _all_ dict methods that would mutate the documents, including lesser-used ones like `popitem`, raise `NotMutable` as they should.

It also allows `copy.deepcopy(Document(...))` to work, whereas on the `master` branch it is necessary to do `copy.deepcopy(dict(Document(...))`.

@tacaswell At some point early on in defining `Document` I floated the idea of a more aggressive override of `__deepcopy__` and you talked me out of it. I forget exactly what I was proposing then, but I think what I've done here is fairly gentle and beneficial. Agree?

This also prompts the question, "What should `Document(...).copy()` do? "Currently we do not override it, so it falls back to `dict.copy()` returns a plain `dict` with a shallow copy of the original `Document`. Recall that we also support `to_dict()` which returns a `dict` with a deep copy. Options I can think of:
* Leave as is.
* Return `self`, since `Document.copy()` might be expected to return a `Document` a "shallow copy of an immutable type" is meaningless.
* Do the same thing as `to_dict()` does. (Invoking a deep copy here seems like it could be surprising though, so probably not a good option).

### DIFF
--- a/databroker/core.py
+++ b/databroker/core.py
@@ -72,7 +72,7 @@ class Document(dict):
     def __setitem__(self, k, v):
         raise NotMutable(self._NOT_MUTABLE_MSG)
 
-    def __delitem__(self, k, v):
+    def __delitem__(self, k):
         raise NotMutable(self._NOT_MUTABLE_MSG)
 
     def pop(self, k):
--- a/databroker/core.py
+++ b/databroker/core.py
@@ -97,7 +97,7 @@ class Document(dict):
         # Convert to dict and then make a deep copy to ensure that if the user
         # mutates any internally nested dicts there is no spooky action at a
         # distance.
-        return copy.deepcopy(self)
+        return copy.deepcopy(dict(self))
 
     def __deepcopy__(self, memo):
         # Without this, copy.deepcopy(Document(...)) fails because deepcopy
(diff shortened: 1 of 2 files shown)


databroker/core.py
```diff
@@ -72,20 +72,39 @@ class Document(dict):
...
+
     def to_dict(self):
         """
         Create a mutable deep copy.
         """
         # Convert to dict and then make a deep copy to ensure that if the user
         # mutates any internally nested dicts there is no spooky action at a
         # distance.
-        return copy.deepcopy(dict(self))
+        return copy.deepcopy(self)
```
tacaswell: Won't this return a `Document`?

tacaswell:
One question about the implementation of `to_dict`, over all :+1:. 

My memory of the deepcopy discussion was about how far down the chain we should make things immutable and if deep-copy should be required to get a mutable version.

databroker/core.py
danielballan: Oops, yes. Fixed and added test.

danielballan:
Punted the question about `copy()` to a separate issue, https://github.com/bluesky/databroker/issues/524

That may require more consideration, shouldn't block getting these fixes in.

PR #529, opened 2020-02-22
Make Document pickle-able

danielballan:
~This PR is a currently just a failing test. It needs an associated fix.~

Includes fix and test. Also includes unrelated commit that makes the implementation more succinct.

### DIFF
--- a/databroker/core.py
+++ b/databroker/core.py
@@ -64,13 +64,29 @@ class Document(dict):
     consumers that expect an object that satisfies isinstance(obj, dict).
     This implementation detail may change in the future.
     """
-    __slots__ = ()
+    __slots__ = ('__dict__',)
     _NOT_MUTABLE_MSG = (
         "Documents are not mutable. Call the method to_dict() to make a "
         "fully independent and mutable copy.")
 
-    def __setitem__(self, k, v):
-        raise NotMutable(self._NOT_MUTABLE_MSG)
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        # This lets pickle recognize that this is not a literal dict and that
+        # it should respect its custom __setstate__.
+        super().__setattr__('__dict__', self)
+
+    def __getstate__(self):
+        return dict(self)
+
+    def __setstate__(self, state):
+        super().__setattr__('__dict__', self)
+        dict.update(self, state)
+
+    def __setitem__(self, key, value):
+        if isinstance(self.__dict__, self.__class__):
+            raise NotMutable(self._NOT_MUTABLE_MSG)
+        else:
+            return dict.__setitem__(self.__dict__, key, value)
 
     def __delitem__(self, k):
         raise NotMutable(self._NOT_MUTABLE_MSG)
--- a/databroker/core.py
+++ b/databroker/core.py
@@ -65,9 +65,6 @@ class Document(dict):
     This implementation detail may change in the future.
     """
     __slots__ = ('__dict__',)
-    _NOT_MUTABLE_MSG = (
-        "Documents are not mutable. Call the method to_dict() to make a "
-        "fully independent and mutable copy.")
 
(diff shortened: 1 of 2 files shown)


databroker/core.py
```diff
@@ -64,31 +64,39 @@ class Document(dict):
...
+            "fully independent and mutable deep copy.")
 
-    def update(self, d=None, **kwargs):
-        raise NotMutable(self._NOT_MUTABLE_MSG)
+    def __setitem__(self, key, value):
+        if isinstance(self.__dict__, self.__class__):
+            self.__readonly()
+        else:
+            # This path is necessary to support un-pickling.
+            return dict.__setitem__(self.__dict__, key, value)
```
tacaswell: Can we use `setattr` here?

databroker/core.py
```diff
@@ -64,31 +64,41 @@ class Document(dict):
...
-    def __delitem__(self, k):
-        raise NotMutable(self._NOT_MUTABLE_MSG)
-
-    def pop(self, k):
-        raise NotMutable(self._NOT_MUTABLE_MSG)
-
-    def popitem(self):
-        raise NotMutable(self._NOT_MUTABLE_MSG)
+    def __getstate__(self):
+        return dict(self), self.__not_a_real_dict
```
danielballan: If `self.__not_a_real_dict` will always be `True`, I don't think we need to include it in the serialization.

databroker/core.py
```diff
@@ -64,31 +64,41 @@ class Document(dict):
     consumers that expect an object that satisfies isinstance(obj, dict).
     This implementation detail may change in the future.
     """
-    __slots__ = ()
-    _NOT_MUTABLE_MSG = (
-        "Documents are not mutable. Call the method to_dict() to make a "
-        "fully independent and mutable copy.")
+    __slots__ = ('__not_a_real_dict',)
```
danielballan: :+1: Much better to have a dedicated special flag.

databroker/core.py
```diff
@@ -64,31 +64,40 @@ class Document(dict):
...
+        raise NotMutable(
+            "Documents are not mutable. Call the method to_dict() to make a "
+            "fully independent and mutable deep copy.")
 
-    def update(self, d=None, **kwargs):
-        raise NotMutable(self._NOT_MUTABLE_MSG)
+    def __setitem__(self, key, value):
+        try:
+            self.__not_a_real_dict
+            self.__readonly()
```
danielballan: Move this into an `else` block for extra paranoia/clarity?

danielballan:
I think this is in good shape. Soliciting reviewers for fresh eyes...

databroker/tests/test_document.py
```diff
@@ -48,3 +50,24 @@ def test_to_dict():
     b['x']['y']['z'] = 2
     # Verify original is not modified.
     assert a['x']['y']['z'] == 1
+
+
+def test_pickle_round_trip():
+    expected = Document({'x': {'y': {'z': 1}}})
+    serialized = pickle.dumps(expected)
+    actual = pickle.loads(serialized)
```
mrakitin, line 58: Why not to use the same pattern (one call for dumping and loading) as in the 2 tests below?

danielballan: This test, which was the first commit in the PR and precedes the associated fix, is meant to clearly illustrate the _pickling_ succeeds but _unpickling_ raises an error. By separating them into two lines, it's easier to understand where the failure is.

I imagine that Tom decided to do JSON and msgpack in one line because those currently work.

dmgav:
Looks good to me.

danielballan:
Thanks for the review, all.

databroker/tests/test_document.py
tacaswell: Indeed, I developed the later tests by typing into a terminal and then lazily copy-pasted the code into the test.

PR #535, opened 2020-02-26
Raise MismatchDataKeys exception instead of KeyError

gwbischof:


gwbischof:
This needs https://github.com/bluesky/event-model/pull/152

danielballan:
Note to selves: this CI will pass as soon as the relevant PR into event-model is released, but we should also explicitly raise the pin of the minimum required event-model version.

PR #536, opened 2020-03-09
Clean up __repr__s.

### LINKED ISSUE #532: Give BlueskyRun a one-line __repr__ and multi-line _repr_pretty_.
Prompted by discussion in #531

danielballan:
* Use `_repr_pretty_` for any multi-line reprs.
* Consistently use the actual (sub)class name
  (`self.__class__.__name__`) in `__repr__` but use `BlueskyRun` in all
  `_repr_pretty_` to hide the remote vs. local distinction and the
  distinction between `BlueskyRun` and `BlueskyRunFromGenerator`, which
  is not intended to be an end-user-facing distinction.

Closes #532

### DIFF
--- a/databroker/core.py
+++ b/databroker/core.py
@@ -954,8 +954,8 @@ class RemoteBlueskyRun(intake.catalog.base.RemoteCatalog):
             return f"<{self.__class__.__name__} *REPR RENDERING FAILURE* {exc!r}>"
 
     def _repr_pretty_(self, p, cycle):
-        self._load()
         try:
+            self._load()
             start = self.metadata['start']
             stop = self.metadata['stop']
             out = (f"BlueskyRun\n"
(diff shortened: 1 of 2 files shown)


databroker/core.py
```diff
@@ -944,7 +944,16 @@ def read_canonical(self):
...
     def __repr__(self):
+        try:
+            self._load()
+            start = self.metadata['start']
+            return f"<{self.__class__.__name__} uid={start['uid']!r}>"
+        except Exception as exc:
+            return f"<{self.__class__.__name__} *REPR RENDERING FAILURE* {exc!r}>"
+
+    def _repr_pretty_(self, p, cycle):
         self._load()
```
gwbischof: Maybe this load should be in the try block. The same way that it is done in __repr__.

danielballan: Good catch, thanks.

PR #538, opened 2020-03-10
Fix the handlers of the v1 temp config

mrakitin:
This was discovered at the ALS COSMIC beamline during the integration work with @ihumphrey, @dylanmcreynolds and @tangkong.

### DIFF
--- a/databroker/v2.py
+++ b/databroker/v2.py
@@ -123,13 +123,21 @@ def temp():
 
     from databroker.core import discover_handlers
     handler_registry = discover_handlers()
-    # Let ophyd be an optional dependency.  If it is not installed, then we
-    # clearly do not need its handler for this temporary data store.
-    try:
-        import ophyd.sim
-        handler_registry.setdefault('NPY_SEQ', ophyd.sim.NumpySeqHandler)
-    except ImportError:
-        pass
+
+    # The temp databroker is often (but not exclusively) used in the context of
+    # a demo or tutorial with the simulated devices in ophyd.sim, some of which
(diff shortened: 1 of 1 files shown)


databroker/v2.py
```diff
@@ -120,15 +120,18 @@ def temp():
...
-    # temporary data store.
+
+    from databroker.core import discover_handlers
+    handler_registry = discover_handlers()
+    # Let ophyd be an optional dependency.  If it is not installed, then we
+    # clearly do not need its handler for this temporary data store.
     try:
         import ophyd.sim
+        if not handler_registry:
+            handler_registry['NPY_SEQ'] = ophyd.sim.NumpySeqHandler
```
danielballan, line 140: Why only register this if `handler_registry` is completely empty? If the intention is to avoid shadowing the `'NPY_SEQ'` item found by `discover_handlers()` (if any) should we check for that key specifically? Would `handler_registry.setdefault('NPY_SEQ', ophyd.sim.NumpySeqHandler)` be suitable?

mrakitin: Yes, it was exactly the reason. I updated the code based on your suggestion.

danielballan: Great. Looks good. It’s worth noting for anyone following along that this special case is only necessary for ophyd < 1.4.0 because starting 1.4.0 ophyd declares its handler via an entrypoint.

mrakitin: xref: https://github.com/bluesky/ophyd/pull/806

PR #539, opened 2020-03-11
Allow uri's in the config file

### LINKED ISSUE #537: Allow the config file to have a uri instead of host-port
Host port doesn't allow us to configure the authentication or load balancing that is required for connection to the central mongo cluster. Currently databroker only support host-port.

https://api.mongodb.com/python/current/examples/authentication.html#authentication-examples

gwbischof:
Fixes #537

### DIFF
--- a/databroker/assets/mongo.py
+++ b/databroker/assets/mongo.py
@@ -5,6 +5,7 @@ import logging
 import pymongo
 
 from pymongo import MongoClient
+from databroker.v1 import _get_mongo_database
 
 from . import mongo_core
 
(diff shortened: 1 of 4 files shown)


gwbischof:
Should of made this a draft. It is still a work in progress.

danielballan:
Looks on the right track, once you can satisfy the crufty v1 stuff that is failing tests. :-\

gwbischof:
Still need to figure out how to make this work for V0 databroker

gwbischof:
I think it is ready for review now

databroker/assets/mongo.py
```diff
@@ -104,10 +102,9 @@ def _datum_col(self):
 
     @property
     def _connection(self):
-        if self.__conn is None:
-            self.__conn = MongoClient(self.config['host'],
-                                      self.config.get('port', None))
-        return self.__conn
+        if self.__db is None:
+            self.__db = _get_mongo_database(self.config)
+        return self.__db.client
```
danielballan: Can we use `self._db.client` and skip the `None` check above?

databroker/headersource/mongo.py
```diff
@@ -49,24 +45,14 @@ def reconfigure(self, config):
...
-                    self.config['host'],
-                    self.config['port'])
-                self.__conn = MongoClient(uri)
-            else:
-                self.__conn = MongoClient(self.config['host'],
-                                          self.config.get('port', None))
-        return self.__conn
+        if self.__db is None:
+            self.__db = _get_mongo_database(self.config)
+        return self.__db.client
```
danielballan: Same here

databroker/assets/mongo.py
gwbischof: I don't understand the reason for this change, can you explain?

danielballan: If we access `self._db`, it perhaps the `if self.__db is None: ...` check  itself, and we can reduce the repeated code.

gwbischof: updated

databroker/headersource/mongo.py
gwbischof: updated

Issue #510, opened 2020-01-21
Mark documents_to_xarray as experimental *or* refactor it.


danielballan:
We intend to rewrite `documents_to_xarray`, and the changes are expected to break the API. (See #471.) This work may or may not be included in the 1.0 release. If they are not, we should at least make `documents_to_xarray` and any associated functions internal or clearly marked as experimental and subject to non-backward-compatible changes in the future.

danielballan:
Made it internal (`databroker_to-xarray` -> `_databroker_to_xarray`) in #540. The plan is to make this or something with this functionality public in a future release once we have settled on a good API for it.

PR #540, opened 2020-03-11
Reduce public API before release.

danielballan:
There are many utility methods in `core` that were written in prep for a
refactor of `databroker_to_xarray`, some of which are not actually used yet,
and maybe all of which could be moved to `event-model` because they are
generically useful for manipulating Documents. In prepartion for possibly
moving and/or re-tooling them, I have made them private except for those
that are imported to drivers.

I also moved `tail`, which a text-parsing utility only used by the JSONL driver,
into the JSONL driver.

### DIFF
--- a/databroker/_drivers/jsonl.py
+++ b/databroker/_drivers/jsonl.py
@@ -1,3 +1,4 @@
+import errno
 import glob
 import json
 import os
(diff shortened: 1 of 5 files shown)


danielballan:
Closes #510

danielballan:
One build passed, two failed on random failures. Re-running those just to be sure.

gwbischof:
Looks good to me

PR #542, opened 2020-03-12
Always yield event_pages from canonical

### LINKED ISSUE #530: Review whether canonical yield EventPage, Event, or a mix.
I think the intent is to always return EventPage, but in the course of doing something else I think I observed some Events come out of `run.canonical(...)`.

The motivation for always returning EventPages is to make sure that all v2 consumers can handle large numbers of events efficiently.

gwbischof:
Fixes #530


gwbischof:
I tested this PR with 500 runs per beamline, and everything passed.

PR #545, opened 2020-03-20
Flake8 compliance

danielballan:
Inspired by https://github.com/bluesky/ophyd/pull/823 and https://github.com/bluesky/bluesky/pull/1326, this runs `flake8` in the databroker CI and brings (some of) the codebase into compliance.

The flake8 configuration file excludes modules, scripts, and tests related to v0 because the v0 code is expected to be deprecated and removed once v1 (which provides a v0-compatible API shimmed on top of v2) in proven reliable.

Changes include whitespace, removing unused imports, reordering imports, removing some unused local variables, adding some missing f-strings, and fixing some misspelled variables on a new codepath that is apparently untested.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ matrix:
   - python: 3.7  # Separate build to run flake8 check
     env:
       - INTAKE_MASTER=false
-      - FLAKE_8=true
+      - FLAKE8=true
   - python: 3.6
     env:
       - INTAKE_MASTER=false
@@ -54,7 +54,7 @@ env:
     - secure: "spb6HwGVvMPsQw9cEwJscmAlG7seL0svW8ouYlDr4OESJJJqC8bynEpAl7Zm/NuKiZ701IR9gEBFuwCRoEXNJ6AkqgBE0DOMj4Oezg4eUnaZz0aXJ+C6mz31vbcNmU6MrjSElw80cIXBpia3j//VMluMzJMsGx/ELPATu2nwyx1+KXljGpFe1/7HYcq/wFHi+sC02sZRVYJGqrb5Zms35hlk4xasWKlsgLh53kW+ss+4hY67vllL2iRkYiP+8z4e3dTOuDqJWx5lLT16wt0K+wJyP/QTyIP81tmMNoWRbdrfOrzyyVXZDbrlzpZDBRtvlMFMHOeEZOrFPIYuclEjT39Tge+T/OxR6SC8kvzqVItxyxLEKWpQu8DHRVEDAuirteuv4IKEnS80nbWn4v/sBOU+sC0/23dW9zZkMrNmRoSFcjudmy2PGIY8jxzOhGJJ+Rbi0xIaKwdLbDZoPHM3m9m2uWYfGNMBHDgnh/HDIq71zzYWaW4gChYgU0UeDM1RYnr1XRa44znlZ0YLnmEyxn/5LzH/hRDxqc5XH5OAiKLnTddZFYsvdRT1JMNSA5PRSm7E8BCvCOEwovjnvRUc0dktyd8+OC5+Eyf1neJ6M7F2sGxCi3yMKw3VmfCxINu6XO3GsPE7cKACEq0lriCxjNzFWbedOX9dTL1Uqdc3pQA="
 
 before_install:
-  - if [ ! -z "$FLAKE_8" ]; then
+  - if [ ! -z "$FLAKE8" ]; then
         pip install flake8;
         flake8 .;
         let res=$?;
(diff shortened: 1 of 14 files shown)


.travis.yml
```diff
@@ -3,14 +3,27 @@ language: python
...
   include:
+  - python: 3.7  # Separate build to run flake8 check
+    env:
+      - INTAKE_MASTER=false
+      - FLAKE_8=true
   - python: 3.6
-    env: INTAKE_MASTER=false
+    env:
+      - INTAKE_MASTER=false
+      - FLAKE8=
```
mrakitin: This variable name is different from the one which has `true`. I suggest removing the ones which have empty values, as they will default to `false`. I suggested to do the same in https://github.com/bluesky/ophyd/pull/823/files#r394048095. I personally don't like the `_`-separation between `FLAKE` and `8`. `FLAKE8` would match the corresponding program name. But it should be coordinated with @dmgav in his flake8 PR in ophyd.

.travis.yml
```diff
@@ -41,6 +54,13 @@ env:
     - secure: "spb6HwGVvMPsQw9cEwJscmAlG7seL0svW8ouYlDr4OESJJJqC8bynEpAl7Zm/NuKiZ701IR9gEBFuwCRoEXNJ6AkqgBE0DOMj4Oezg4eUnaZz0aXJ+C6mz31vbcNmU6MrjSElw80cIXBpia3j//VMluMzJMsGx/ELPATu2nwyx1+KXljGpFe1/7HYcq/wFHi+sC02sZRVYJGqrb5Zms35hlk4xasWKlsgLh53kW+ss+4hY67vllL2iRkYiP+8z4e3dTOuDqJWx5lLT16wt0K+wJyP/QTyIP81tmMNoWRbdrfOrzyyVXZDbrlzpZDBRtvlMFMHOeEZOrFPIYuclEjT39Tge+T/OxR6SC8kvzqVItxyxLEKWpQu8DHRVEDAuirteuv4IKEnS80nbWn4v/sBOU+sC0/23dW9zZkMrNmRoSFcjudmy2PGIY8jxzOhGJJ+Rbi0xIaKwdLbDZoPHM3m9m2uWYfGNMBHDgnh/HDIq71zzYWaW4gChYgU0UeDM1RYnr1XRa44znlZ0YLnmEyxn/5LzH/hRDxqc5XH5OAiKLnTddZFYsvdRT1JMNSA5PRSm7E8BCvCOEwovjnvRUc0dktyd8+OC5+Eyf1neJ6M7F2sGxCi3yMKw3VmfCxINu6XO3GsPE7cKACEq0lriCxjNzFWbedOX9dTL1Uqdc3pQA="
 
 before_install:
+  - if [ ! -z "$FLAKE_8" ]; then
+        pip install flake8;
+        flake8 .;
```
mrakitin, line 55: I think `set -e` will cover that, and the 3 lines below may not be needed. Not sure about semi-colons.

databroker/v1.py
```diff
@@ -1671,15 +1668,15 @@ def _get_mongo_database(config):
...
     if {'mongo_user', 'mongo_pwd', 'host', 'port'} <= set(config):
         uri = 'mongodb://{0}:{1}@{2}:{3}/'.format(
-                self.config['mongo_user'],
-                self.config['mongo_pwd'],
-                self.config['host'],
-                self.config['port'])
+            config['mongo_user'],
+            config['mongo_pwd'],
+            config['host'],
+            config['port'])
```
mrakitin, line 1679: Why not f-string here?

examples/generate_mongo_data.py
```diff
@@ -22,5 +23,4 @@
 handler.setLevel('DEBUG')
 logger.addHandler(handler)
 
-from databroker._drivers.mongo_normalized import BlueskyMongoCatalog
-catalog = BlueskyMongoCatalog(f'{directory}/*.msgpack')
+catalog = BlueskyMongoCatalog(mds, fs)
```
mrakitin, line 26: Good catch!

.travis.yml
mrakitin: Or it can be called something like `CHECK_STYLE` to open a potential for another tool in the future.

.travis.yml
dmgav: I don't think that using `set -e` will completely replace the functionality. The idea is to add a message to the log after `flake8` exits.

.travis.yml
dmgav: I replaced `FLAKE_8` by `FLAKE8` in the `ophyd` PR.

dmgav:
Looks good to me. I don't see any issues with the changes.

databroker/v1.py
danielballan: I'll make a separate PR to make this the lightest touch possible to satisfy `flake8`.

danielballan:
The `master` branch is broken due to two unrelated issues with upstream changes (ophyd and intake). All errors in CI relate to those. Merging to avoid rebase pain, will address `master` breakages separately.

PR #547, opened 2020-03-26
Make LazyMap pickleable.

danielballan:


gwbischof:
Do you know what the problem is with the CI failures?

danielballan:
Ci fixed in #548

danielballan:
One random failure, all good otherwise.

PR #551, opened 2020-03-31
Fix python2 codepath.

danielballan:
Code was not actually valid Python 2 syntax, so the nice error never
showed. Testd with `python2 setup.py sdist`.


danielballan:
One typical random failure on one CI build. 3/4 passed.

Issue #431, opened 2019-08-29
all-digit uids get treated as scan_id by intake client-server


gwbischof:
This test fails occasionally with:
`KeyError: '86038425'`

The uuid doesn't have any letters in it so databroker thinks its an index and not a partial uuid.

gwbischof:
This only happens if it is run with the server

tacaswell:
Could probably cast to int -> try -> catch failure -> try again as a string?

danielballan:
That's clever, but it could lead to even more confusing failures. What if the user enters

```py
db['1003']
```

and there happens to be a scan with `scan_id=1003`? If the user gives us a string, we have to assume they mean (partial) UUID.

gwbischof:
test_search_for_smoke is the test.

danielballan:
I think the right fix for this is to make sure that the type---`int` scan_id or `str` (partial) uid---survives the client-server interaction upstream in intake. Since intake's payload is msgpack-encoded, the distinction should survive the passage between client and server.

danielballan:
This is resolved on the 2.0.0-dev branch.

PR #552, opened 2020-03-31
Update to support databases with authentication

gwbischof:
If the user account is read only, index creation fails. So in this PR indexes creation is changed so that they are made only when needed.

### DIFF
--- a/databroker/assets/base_registry.py
+++ b/databroker/assets/base_registry.py
@@ -463,8 +463,6 @@ class RegistryTemplate(BaseRegistryRO):
             The uid of the created resource.
 
         '''
-        if callable(getattr(self, '_create_resource_index', None)):
-            self._create_resource_index()
         if root is None:
             root = ''
 
@@ -506,8 +504,6 @@ class RegistryTemplate(BaseRegistryRO):
             Datum uid to put put into the Event data.
 
         '''
-        if callable(getattr(self, '_create_datum_index', None)):
-            self._create_datum_index()
         if validate:
             raise RuntimeError('validate not implemented yet')
         col = self._datum_col
@@ -542,8 +538,6 @@ class RegistryTemplate(BaseRegistryRO):
             entry
 
         '''
-        if callable(getattr(self, '_create_datum_index', None)):
-            self._create_datum_index()
         if validate:
             raise RuntimeError('validate not implemented yet')
         col = self._datum_col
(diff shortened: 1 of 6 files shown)


gwbischof:
Still a draft

databroker/assets/base_registry.py
```diff
@@ -463,6 +463,8 @@ def register_resource(self, spec, root, rpath, rkwargs,
             The uid of the created resource.
 
         '''
+        if callable(getattr(self, '_create_resource_index', None)):
```
danielballan: If this exists but is not a callable, what does that mean? Should it pass silently?

danielballan:
With fresh eyes I now understand what looks off about this to me. I think this is doing:

```py
In [2]: class A: 
   ...:     def some_method(self): 
   ...:         print('some method!') 
   ...:         if hasattr(self, '_extra_step'): 
   ...:             self._extra_step() 
   ...:                                                                                                                                                                                       

In [3]: class B(A): 
   ...:     def _extra_step(self): 
   ...:         print('extra step!') 
   ...:                                                                                                                                                                                       

In [4]: a = A()                                                                                                                                                                               

In [5]: b = B()                                                                                                                                                                               

In [6]: a.some_method()                                                                                                                                                                       
some method!

In [7]: b.some_method()                                                                                                                                                                       
some method!
extra step!
```

where it would be better to rely on `super()` as in

```py
In [14]: class A: 
    ...:     def some_method(self): 
    ...:         print('some method!') 
    ...:                                                                                                                                                                                      

In [15]: class B(A): 
    ...:     def _extra_step(self): 
    ...:         print('extra step!') 
    ...:     def some_method(self): 
    ...:         super().some_method() 
    ...:         self._extra_step() 
    ...:                                                                                                                                                                                      

In [16]: a = A()                                                                                                                                                                              

In [17]: b = B()                                                                                                                                                                              

In [18]: a.some_method()                                                                                                                                                                      
some method!

In [19]: b.some_method()                                                                                                                                                                      
some method!
extra step!
```

or, equivalently

```py
In [20]: class B(A): 
    ...:     def some_method(self): 
    ...:         super().some_method() 
    ...:         print('extra step!') 
    ...:                                                                                                                                                                                      
```

By using `super()`, we can put all the special logic in the subclass, rather than teaching the base class to anticipate and delegate to any subclass-specific methods that may exist.

databroker/assets/base_registry.py
tacaswell: I agree, `hasattr` or `... is not None` is what I would have naively expected to be here.

tacaswell:
I agree with @danielballan about using `super()` and extending the sub-classes.

Setting up the indexes on init is something we put in early because without the indexes some of our standard operations are super slow and we were working around insufficient devops on the floor.

I think it is ok to either silently eat this exception if we can't create the index or just drop it all together and assert that the devops need to be done correctly.

danielballan:
One random failure on one build, due to known issue #431

PR #553, opened 2020-04-03
Support relative root_map.

danielballan:
This is needed for databroker-pack. Needs tests.


tacaswell:
ok, so we now know there _is_ a bug in here, I'm inclined to merge this as is to unblock databroker-pack and then sort out why updating this is changing the return value of `get_config()`.

PR #559, opened 2020-05-07
Try __getitem__ again with a forced reload.

danielballan:
* If a __getitem__ lookup fails, try again after forcing a reload.
* Proactively force a reload before recency-based indexing (e.g.
``catalog[-1]``) because a stale cache is a big problem in that case.

This is dashed off and needs careful thought before merging. ~Also, it should be applied to the mongo drivers if indeed it is to be merged.~

### DIFF
--- a/databroker/_drivers/jsonl.py
+++ b/databroker/_drivers/jsonl.py
@@ -6,7 +6,7 @@ import pathlib
 import event_model
 
 from ..in_memory import BlueskyInMemoryCatalog
-
+from ..core import retry
 
 def gen(filename):
     """
@@ -142,6 +142,7 @@ class BlueskyJSONLCatalog(BlueskyInMemoryCatalog):
                 stop_doc = get_stop(filename)
                 self.upsert(start_doc, stop_doc, gen, (filename,), {})
 
+    @retry
     def search(self, query):
         """
         Return a new Catalog with a subset of the entries in this Catalog.
--- a/databroker/_drivers/jsonl.py
+++ b/databroker/_drivers/jsonl.py
@@ -8,6 +8,7 @@ import event_model
 from ..in_memory import BlueskyInMemoryCatalog
 from ..core import retry
 
+
 def gen(filename):
     """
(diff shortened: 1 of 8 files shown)


gwbischof:
How about a decorator for retrying?

danielballan:
> Can you describe the situation where `__getitem__` had a problem (race condition)?

You can reproduce it like

```py
db[-1]
RE(...)  # assume this is fast
db[-1]
```
The second lookup uses a cached set of entries, and misses the fact that one was just added. For recency-based lookups this is especially series because you _will_ get a result but it will be wrong, which is worse than getting an error.

> How about a decorator for retrying?

I am not sure about that because the behavior is different depending on which branch we are on:

* If we are doing a recency-based lookup, `force_reload()` _first_ to make sure we get an up-to-date result.
* Otherwise, try to save overhead by first attempting the lookup and only force a reload if the lookup fails.

Some of the logic in `__getitem__` would have to be copied into the decorator. Might be worth trying but it's not obviously simpler than a direct approach.

gwbischof:
Earlier you said that this needs careful thought before merging. Was there some specific concern that you had?

danielballan:
No, only that I dashed this off in the state of mind, "AHHHHI  just gotta get this working in time for the presentation!" and wanted to revisit it with a cooler head. I don't see any downsides here.

gwbischof:
The mongo tests pass without the retry, and the force_reload when using index -1. Should we update that anyway?

danielballan:
I have struck out my comment above

> Also, it should be applied to the mongo drivers if indeed it is to be merged.

because it turns out when the mongo drivers process `-N` they always query the database, so there is no chance of a stale cache there. No changes needed.

danielballan:
Travis is green https://travis-ci.org/github/bluesky/databroker/builds/696600089 although the UI here hasn't updated (at least for me).

This is all @gwbischof's party now, my PR in name only, so I'm calling myself a qualified reviewer and merging.

PR #563, opened 2020-05-14
Prevent recursion back to same catalog when looking for attrs

ronpandolfi:
This PR addresses an issue with `BlueskyRun.__getattr__` that causes infinite recursion. When a key is not present on a BR, it's `._entry` is queried as well. Often, `._entry` ends up being the same BR. This checks for that case, and raises before continuing to `._entry`.

### DIFF
--- a/databroker/core.py
+++ b/databroker/core.py
@@ -1144,7 +1144,8 @@ class BlueskyRun(intake.catalog.Catalog):
             if key != "_entry" and self._entry.name != self.name:
                 return getattr(self._entry, key)
             else:
-                raise AttributeError("Aborted before recursing back to self.") from ex
+                # Aborted before recursing back to self.
+                raise AttributeError(self) from ex
 
     def canonical(self, *, fill, strict_order=True):
         yield from _canonical(start=self.metadata['start'],
--- a/databroker/core.py
+++ b/databroker/core.py
@@ -1142,7 +1142,6 @@ class BlueskyRun(intake.catalog.Catalog):
             # The user might be trying to access an Entry method. Try that
             # before giving up.
             if key != "_entry" and self._entry.name != self.name:
-                print(self._entry.name, self.name)
                 return getattr(self._entry, key)
             else:
                 raise AttributeError("Aborted before recursing back to self.") from ex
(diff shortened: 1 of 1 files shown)


databroker/core.py
```diff
@@ -1138,10 +1138,14 @@ def __getattr__(self, key):
...
-        except AttributeError:
+        except AttributeError as ex:
             # The user might be trying to access an Entry method. Try that
             # before giving up.
-            return getattr(self._entry, key)
+            if key != "_entry" and self._entry.name != self.name:
+                print(self._entry.name, self.name)
+                return getattr(self._entry, key)
+            else:
+                raise AttributeError("Aborted before recursing back to self.") from ex
```
danielballan: For users a plain `AttributeError(self)` might be more helpful here, not sure we need to expose the recursion detail in the message---how about sticking it in a comment instead so it will appear in the traceback? Just a thought.

ronpandolfi:
@danielballan Can you confirm that the check I'm doing is correct. I'm not sure what the best strategy is for checking equivalence here. Seems to work for me, but I haven't seen all the edges.
```
self._entry.name != self.name
```

tacaswell:
if the entry _is_ the BR then why not check `self._entry is self`?

ronpandolfi:
> if the entry _is_ the BR then why not check `self._entry is self`?

@tacaswell I had tried that, and don't fully understand what happened, but it doesn't work. It still recurses infinitely. At each depth, I asked
```
id(self), id(self._entry), id(self._entry._entry)
```
and it turns out that the first two are never equal. The last two are always equal. Somehow self._entry is not self, but `getattr(self._entry, key)` goes back to doing the getattr of the parent.

danielballan:
Wait, the `Entry` is never the `BlueskyRun`. They are two separate objects coupled by their `__getattr__`.

danielballan:
Was there decision on how to move forward on this? I think someone mentioned that between when I left and re-joined last Friday's call, some insight was made.

danielballan:
Thanks again for kicking this off, @ronpandolfi. The issue was fixed in #569, and #572 (just merged) adds a test to ensure it doesn't creep back in.

PR #569, opened 2020-05-21
Adjust for Entry unification

danielballan:
This PR adjust for changes in intake that will be introduced in
intake/intake#490. This branch has been tested against that PR branch, as well
as intake `master` and the most recent intake release, `0.5.5`.

This is how it looks interactively against that PR branch:

```py
In [1]: import databroker                                                                                                                                                                     

In [3]: list(databroker.catalog)                                                                                                                                                              
Out[3]: 
['dev1',
 'dev2',
 'fxi_feb',
 'test1',
 'test_unpack1',
 'test_unpack3',
 'dev',
 'csx',
 'example']

In [4]: databroker.catalog.dev1                                                                                                                                                               
Out[4]: <Intake catalog: dev1>  # <--- NOT AN ENTRY ANYMORE

In [5]: type(databroker.catalog.dev1)                                                                                                                                                         
Out[5]: databroker._drivers.msgpack.BlueskyMsgpackCatalog
```

### DIFF
--- a/databroker/core.py
+++ b/databroker/core.py
@@ -29,7 +29,6 @@ from .intake_xarray_core.xarray_container import RemoteXarray
 from .utils import LazyMap
 from collections import deque, OrderedDict
 from dask.base import normalize_token
-from intake.utils import DictSerialiseMixin
 
 
 logger = logging.getLogger(__name__)
@@ -255,7 +254,6 @@ class Entry(intake.catalog.local.LocalCatalogEntry):
     #     print('bob')
     #     metadata = self.describe()['metadata']
     #     return ('Entry', metadata['start']['uid'])
-
 
 
 class StreamEntry(Entry):
@@ -825,7 +823,9 @@ class RemoteBlueskyRun(intake.catalog.base.RemoteCatalog):
     kwargs: ignored
     """
     name = 'bluesky-run'
+
     # opt-out of the persistence features of intake
+
     @property
     def has_been_persisted(self):
         return False
--- a/databroker/core.py
+++ b/databroker/core.py
@@ -251,6 +251,12 @@ class Entry(intake.catalog.local.LocalCatalogEntry):
             self.__cache[token] = datasource
         return datasource
 
+    # def __dask_tokenize__(self):
+    #     print('bob')
+    #     metadata = self.describe()['metadata']
+    #     return ('Entry', metadata['start']['uid'])
+
+
 
 class StreamEntry(Entry):
     """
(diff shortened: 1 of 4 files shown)


danielballan:
Restarted CI to run against current intake `master` just to be sure.

ronpandolfi:
Confirming this works on my side. Excellent work both 👍

databroker/core.py
```diff
@@ -237,6 +250,11 @@ def get(self, **kwargs):
             self.__cache[token] = datasource
         return datasource
 
+    # def __dask_tokenize__(self):
```
gwbischof, line 253: I don't think we need these comments.

databroker/core.py
```diff
@@ -245,13 +263,11 @@ class StreamEntry(Entry):
...
         return dict()
 
-    def __getstate__(self):
-        args = [arg.__getstate__() if isinstance(arg, DictSerialiseMixin)
-                else arg for arg in self._captured_init_args]
-        kwargs = OrderedDict({k: arg.__getstate__()
-                              if isinstance(arg, DictSerialiseMixin) else arg
-                              for k, arg in self._captured_init_kwargs.items()})
-        return OrderedDict(cls=self.classname, args=args, kwargs=kwargs)
+    # def __dask_tokenize__(self):
```
gwbischof, line 266: Same here

gwbischof: And a few other places also

tacaswell:
The `__dask_tokenize__` methods are left there as a long-form TODOs / skeletons.

We played with them while sorting out how to make this fast.  One of the effects of the changes here is to defer actually computing the tokenization until actually needed (which for most cases for use is "never" because we turned off all of the persist functionality), but if we ever do need the tokenization it will be slow again.  Finishing the implementation of `__dask_tokenize__` to not convert the kwargs to a string and then hash it is the way to do that (now that upstream is changed to respect our implementation).

gwbischof:
If we use the intake server do we need to tokenize?

danielballan:
I think so, yes. To summarize the state of things:

Previously --- Tokenizaiton happened "greedily" when an instance was first created, so we always paid the price for it, whether the token was actually accessed or not.
In this PR --- Tokenization happens only if and when the token is first accessed. And it is cached, so any repeated access will be fast. Often, in the local use case, it is never accessed, so now things are a lot faster.
In the future --- When the `__dask_tokenize__` sketches are finished, we can make tokenization faster across the board.

PR #573, opened 2020-06-03
Work around a regression in items() in intake 0.6.0.

danielballan:


danielballan:
Added an empty commit to poke Travis, which somehow didn't trigger on 5bceeb7.

danielballan:
CI failures verified to be common transient/random failures related to sqlite. Self-merging as a hot-fix because databroker is currently broken against the latest intake release.

PR #575, opened 2020-06-04
Allow filler_class as string object path.

danielballan:


databroker/v2.py
```diff
@@ -62,8 +63,13 @@ class Broker(Catalog):
     def __init__(self, *, handler_registry=None, root_map=None,
                  filler_class=event_model.Filler, transforms=None, **kwargs):
 
+
```
gwbischof:
```suggestion

```

PR #581, opened 2020-06-18
FIX: if a handler is not lazy we have unknown shape don't force it

tacaswell:
Passing the -1 in the shape through will result in an eventual shape
miss-match when trying to retrieve the data.  Rather than fail,
aggressively pull the data up.  The assumption is that if there were a
handler that was expected to load large volumes of data the handler
would be natively delayed.

### DIFF
--- a/databroker/core.py
+++ b/databroker/core.py
@@ -2134,9 +2134,14 @@ def coerce_dask(handler_class, filler_state):
             descriptor = filler_state.descriptor
             key = filler_state.key
             shape = extract_shape(descriptor, key)
-            dtype = extract_dtype(descriptor, key)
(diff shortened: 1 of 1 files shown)


tacaswell:
@gwbischof Can you please sort out a test case for this?

danielballan:
Confirmed on Run`'89c1bfac-73eb-407b-9012-d852231c6eda'` from FXI that using databroker 1.0.6 we get

```py
>>> run.primary.read()
ValueError: replacement data must match the Variable's shape. replacement data has shape (172, 10); Variable has shape (172, -1)
```

but with this branch access works.

danielballan:
Punting unit testing to https://github.com/bluesky/databroker/issues/598

PR #585, opened 2020-07-30
Projector implementation

dylanmcreynolds:
Implementation of projector that can be called with a BlueskyRun, reading a projection as specified in https://github.com/bluesky/event-model/pull/179

This is the result of lots of discussion with @ronpandolfi  and @tacaswell. 

The result is an xarray.Dataset. Single value fields (like those that might appear in the start doc) would be projected into the Dataset's attrs field. Multi-value event stream fields appear in the Dataset with the projection's key and a Datarray value.

The projection can also support calculable fields. The projection type "calculation" can specify a callable that can be invoked by the projector to populate a Datarray into the return Dataset.

There was also a suggestion for a static type in projections where a value was pulled directly out of the projection. I'm open to this, but left it out for now. I think a lot of thought needs to go into what types of values are supported.

I had originally defined a class structure for projectors, but was convinced that since I wasn't storing state, that a function was good enough.

### DIFF
--- a/databroker/projector.py
+++ b/databroker/projector.py
@@ -30,7 +30,7 @@ def get_run_projection(run: BlueskyRun, projection_name: str = None):
     Raises
     ------
     KeyError
-        If the a projection_name is specified and there is more than one 
+        If the a projection_name is specified and there is more than one
         projection in the run with that name
     """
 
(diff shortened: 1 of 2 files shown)


danielballan:
Power-cycled to rerun CI with fixes in master related to docs.

danielballan:
One random failure. Merging!

PR #586, opened 2020-08-04
Added some documentation on v2 metadata access

cryos:


doc/source/v2/user/index.rst
```diff
@@ -249,6 +249,32 @@ See the documentation on dask_.
...
+
+Information only knowable at the end, like the exit status (success, abort,
+fail) is stored in ``run.metadata['stop']``.
+
+.. ipython:: python
+
+    run.metadata['stop']
+
+The v1 API stored metadata about devices involved and their configuration,
+accessed using ``descriptors``, this is roughly equivalent to what is available
```
danielballan, line 269: This isn't as precise as I'd prefer. The "baseline" stream contains readings from a different (perhaps overlapping) set of hardware from other streams. From databroker's POV it is "just another stream" and does not necessarily even exist, depending on how the data acquisition system was configured. The "descriptors", which are present in v1 and v2, contain metadata about the columns (dtype, shape, optional units, etc.) and additional one-time only readings from the hardware necessary for interpreting the data in the columns (exposure time, motor velocity setting, etc.). A difference between v1 and v2 is that in v2 descriptor readings are "flattened" into the xarray.Dataset whereas in v1 they had to accessed by a special, separate method. I don't love the flattening, but it serves the purpose of making these readings more discoverable. Perhaps we'll come up with a better way before we declare v2 "stable".

cryos: I think I get the mapping and made a new commit that hopefully maps v1 to v2 more accurately.

doc/source/v2/user/index.rst
```diff
@@ -249,6 +249,32 @@ See the documentation on dask_.
...
 require standing up a server.
 
+Explore Metadata
+----------------
+
+Everything recorded at the start of the run is in ``run.metadata['start']``.
+
+.. ipython:: python
+
+    run.metadata['start']
```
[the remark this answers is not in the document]
cryos: You are welcome, just trying to find my way around.

cryos:
Thanks @danielballan! Happy to give it at least a day before merging.

mrakitin:
The CI failures should be resolved before proceeding:
```
Warning, treated as error:

/home/travis/build/bluesky/databroker/doc/source/generated/databroker.Broker.stats.rst:document isn't included in any toctree

Makefile:61: recipe for target 'html' failed

make: *** [html] Error 2

make: Leaving directory '/home/travis/build/bluesky/databroker/doc'
```

tacaswell:
The master branch is broken with the same issue: https://travis-ci.org/github/bluesky/databroker/builds/714958654  I don't think we should hold this PR hostage to fixing that.

doc/source/v2/user/index.rst
tacaswell: This makes it sound a bit like we have gotten rid of descriptors altogether to me rather than re-arrange.  Maybe something like "In v1 you accessed information about devices + configuration by directly accessing the descriptor documents, in v2 this infromation is flattened into ..." ?

mrakitin:
Sounds good, @tacaswell, agree it's irrelevant to this PR (but the docs won't be built/published!)

danielballan:
> eaving merging to @danielballan

Note: This was accidentally self-merged (it's been a long day for us all) but I am in favor of it having been merged. Fix for pre-existing CI problem is in a separate PR, https://github.com/bluesky/databroker/pull/587.

PR #584, opened 2020-07-29
Expose the mongo db.collection.stats() report

prjemian:
Expose access to the [mongo collection statistics](https://docs.mongodb.com/manual/reference/method/db.collection.stats/).

PR #587, opened 2020-08-06
Add test for 'stats' method and further document it.

danielballan:
The new `stats()` method introduced in #584 was (automatically) documented in
the v2 reference documentation, but because of the way v1 documentation is
organized it had to be manually added there.

Also, while I was here I added a unit test for `stats()`.


prjemian:
Just a bit slow to the party here.  Had 3-hour call today training someone on the hklpy code.  Very successful, most questions answered.

PR #590, opened 2020-08-25
Expose mongo client and database

danielballan:
I'm not sure this is a good idea.


tacaswell:
bring back `DangerBroker`!

Issue #591, opened 2020-08-27
Old resource documents


gwbischof:
Some old resource documents don't have a uid, instead they use '_id'.

The old version of databroker was able to find resources if the resource didn't have a uid key:
https://github.com/bluesky/databroker/blob/master/databroker/assets/mongo_core.py#L48

For backward compatibility I think we should do the same for the mongo_normalized and embedded drivers.

With databroker.v1 the resource document could not be found. But it is found with the old databroker.
The resource document doesn't have a uid key.
![image](https://user-images.githubusercontent.com/11858848/91497785-9b7dcd80-e88c-11ea-82c9-69ae26eb3eb2.png)

PR #592, opened 2020-08-27
also search for resources by _id

gwbischof:
Fixes https://github.com/bluesky/databroker/issues/591

### DIFF
--- a/databroker/_drivers/mongo_normalized.py
+++ b/databroker/_drivers/mongo_normalized.py
@@ -285,7 +285,7 @@ class BlueskyMongoCatalog(Broker):
         # referenced by '_id'.
         if doc is None:
             try:
-                uid = ObjectId(uid)
+                _id = ObjectId(uid)
             except InvalidId:
                 pass
             else:
--- a/databroker/_drivers/mongo_normalized.py
+++ b/databroker/_drivers/mongo_normalized.py
@@ -290,6 +290,7 @@ class BlueskyMongoCatalog(Broker):
             doc = self._resource_collection.find_one({'uid': uid})
         else:
             doc = self._resource_collection.find_one({'_id': uid})
+            doc['uid'] = uid
 
         if doc is None:
             raise ValueError(f"Could not find Resource with uid={uid}")
--- a/databroker/_drivers/mongo_normalized.py
+++ b/databroker/_drivers/mongo_normalized.py
@@ -304,8 +304,8 @@ class BlueskyMongoCatalog(Broker):
         return doc['resource']
 
     def _get_datum_cursor(self, resource_uid):
-        cursor = self._datum_collection.find({'resource': resource_uid}, {'_id': False})
         self._schema = {}  # TODO This is cheating, I think.
+        return self._datum_collection.find({'resource': resource_uid}, {'_id': False})
 
     def _make_entries_container(self):
         return _Entries(self)
--- a/databroker/_drivers/mongo_normalized.py
+++ b/databroker/_drivers/mongo_normalized.py
@@ -284,7 +284,7 @@ class BlueskyMongoCatalog(Broker):
         # Some old resource documents don't have a 'uid' and they are
         # referenced by '_id'.
         if doc is None:
-            doc = self._resource_collection.find_one({'_id': uid}, {'_id': False}))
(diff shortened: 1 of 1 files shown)


danielballan:
This logic goes, "First, check if the string in a Datum's `resource` field can be interpreted as a valid `ObjectId`. If it can be, assume it's an `_id` and look it up in that fashion. If it can't be, fall back to treating it as a `uid`." I think we should do the opposite: first try to treat it as a `uid`, and if that lookup fails then try interpreting it as an `ObjectId`. If that is valid, try looking it up as an `_id`.

The reason is that, [according to the schema for Resource documents, the `uid` can be any string](https://github.com/bluesky/event-model/blob/91d0b1c1a23d1d6fe473fa3782fd7522a1785af9/event_model/schemas/resource.json#L24-L27). Therefore, it's possible that some document producer could generate Resource documents with a `uid` that just _happens_ to be a valid `ObjectId`. That would put this logic on the wrong path. It is safer to assume the documents are schema-compliant and try that first, only trying the `_id` path if we know that the `uid` path fails. (In all the document-writing code that I _know of_ the Resource `uid` is a UUID4 and not a valid `ObjectId`, but there is ophyd code at other facilities that we don't follow closely and perhaps some that isn't public.)

There are also performance implications, which are in a distant second place to correctness implications but maybe still worth mentioning. In this implementation, we spend time attempting to cast every Datum's `resource` as an `ObjectId`. For all documents with a UUID4, this test will fail, and we will land in `except InvalidId:`. Exception-catching is cheap compared to a database lookup, but expensive compared to (for example) a dict lookup or a function call. If we expect to find an `_id` in the minority of cases---for "old" document only---we might  prioritize the performance of "new", schema-compliant documents by first trying a `uid`-based lookup and falling back to `ObjectId` only if that database lookup fails. That would make the "old" case slower (we waste a database lookup) but it would make the "new", schema-compliant case faster (we save a stack thrash / exception catch).

gwbischof:
Good point.
I think that it may be best to use a transform here, I don't think we need code to handle invalid documents here.
Using transforms will help us track, and implement the migration.
What do you think?

danielballan:
Oh, even better! Yes, I think that makes a lot of sense.

gwbischof:
Actually, on second thought, I don't think a transform will work, because without the change to the code the resource document won't be found...

databroker/_drivers/mongo_normalized.py
```diff
@@ -283,11 +279,21 @@ def _get_resources(self, run_start_uid):
...
     def _get_resource(self, uid):
-        doc = self._resource_collection.find_one(
-            {'uid': uid})
+        doc = self._resource_collection.find_one({'uid': uid}, {'_id': False})
+
+        # Some old resource documents don't have a 'uid' and they are
+        # referenced by '_id'.
+        if doc is None:
+            try:
+                uid = ObjectId(uid)
```
danielballan: Redefining local variables is something I try to avoid. It can make code harder to read, and it can reduce the usefulness of interactive debugging  (i.e. `%debug`) because you can't access the original value.
```suggestion
                _id = ObjectId(uid)
```

databroker/_drivers/mongo_normalized.py
```diff
@@ -283,11 +279,21 @@ def _get_resources(self, run_start_uid):
...
+            except InvalidId:
+                pass
+            else:
+                doc = self._resource_collection.find_one({'_id': uid}, {'_id': False})
```
danielballan: Follows from the change suggested above
```suggestion
                doc = self._resource_collection.find_one({'_id': _id}, {'_id': False})
```

databroker/_drivers/mongo_normalized.py
```diff
@@ -283,11 +279,21 @@ def _get_resources(self, run_start_uid):
...
+                doc['uid'] = uid
```
danielballan, line 293: This needs to be a string, not an `ObjectId`. So, I guess that's a more concrete reason to use a separate variable name for the `_id` above, so we'll still have access to the original (string) `uid` here.

danielballan:
Oh, indeed, the transforms are applied too late to help here.

danielballan:
CI failure is due to #597. The other builds pass.

PR #611, opened 2020-10-06
add EventDatumNotFound exception

jklynch:
Debugging problems resulting in DatumNotFound exceptions is made easier if the corresponding event can be quickly found.

This PR adds EventDatumNotFound to carry the event uid and corresponding datum id.

### DIFF
--- a/databroker/assets/core.py
+++ b/databroker/assets/core.py
@@ -15,7 +15,7 @@ class DatumNotFound(Exception):
     Raised if a Datum id is not found.
     """
     def __init__(self, datum_id, *args, **kwargs):
-        self.super(*args, **kwargs)
+        super(*args, **kwargs)
         self.datum_id = datum_id
 
     def __str__(self):
@@ -27,7 +27,7 @@ class EventDatumNotFound(Exception):
     Raised if an Event document is found to have an unknown Datum id.
     """
     def __init__(self, event_uid, datum_id, *args, **kwargs):
-        self.super(*args, **kwargs)
+        super(*args, **kwargs)
         self.event_uid = event_uid
         self.datum_id = datum_id
 
--- a/databroker/assets/core.py
+++ b/databroker/assets/core.py
@@ -9,8 +9,33 @@ import time as ttime
 import pandas as pd
 from ..utils import sanitize_np, apply_to_dict_recursively
 
+
 class DatumNotFound(Exception):
-    pass
+    """
+    Raised if a Datum id is not found.
+    """
+    def __init__(self, datum_id, *args, **kwargs):
+        self.super(*args, **kwargs)
+        self.datum_id = datum_id
+
+    def __str__(self):
+        return f"No datum found with datum id {self.datum_id}"
+
+
+class EventDatumNotFound(Exception):
+    """
+    Raised if an Event document is found to have an unknown Datum id.
+    """
+    def __init__(self, event_uid, datum_id, *args, **kwargs):
+        self.super(*args, **kwargs)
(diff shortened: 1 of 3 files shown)


danielballan:
As an aside, note that in v2---which relies on `event_model.Filler` to handle these activities---the exceptions are already improved to be exactly as you wish.

https://github.com/bluesky/event-model/blob/14b082682a428d54e92676368777285b17815183/event_model/__init__.py#L932-L935

Here we are looking at v0 codepaths. It makes sense to apply the same improvements there, since we are still relying on v0 at a couple beamlines.

<hr />

Yes, I think it would be better to include it in the exception rather than catch/log/re-raise. It is useful to have the information "in band", and of course at NSLS-II we log all exceptions at the application layer (in `blueksky_ipython.log`) so it will be logged along with everything else.

The function that raises this exception does not have access to the event uid, so we can't just add it there.

https://github.com/bluesky/databroker/blob/82063143b1faa81126b7a18806341b10a28934d7/databroker/assets/core.py#L36-L60

Fortunately, Python gives us `raise from`, and this situation is exactly what it was designed for.

```py
class SomeNewExceptionType(Exception):
    def __init__(self, event_uid, *args, **kwargs):
        self.event_uid = event_uid
        super().__init__(*arsg, **kwargs)

...

try:
    ...
except DatumNotFound as err:
    raise SomeNewExceptionType(event_uid) from err
```

Your new exception will be shown at the bottom of the traceback, but the original exception will be above it, and `%debug` will enable a motivated debugger to access the `event_uid` programmatically. You might also extend `DatumNotFound` to capture the `datum_id` in the same way, enabling something like

```py
class SomeNewExceptionType(Exception):
    def __init__(self, event_uid, datum_id,  *args, **kwargs):
        self.event_uid = event_uid
        super().__init__(*arsg, **kwargs)

...

try:
    ...
except DatumNotFound as err:
    raise SomeNewExceptionType(event_uid, err.datum_id) from err
```

danielballan (changes requested):
Seems help. Left a suggestion about the implementation.

databroker/assets/core.py
```diff
@@ -9,8 +9,33 @@
...
 class DatumNotFound(Exception):
-    pass
+    """
+    Raised if a Datum id is not found.
+    """
+    def __init__(self, datum_id, *args):
+        super().__init__(*args)
+        self.datum_id = datum_id
+
+    def __str__(self):
```
danielballan: Overriding `__str__` seems more aggressive than necessary. If one prints this object in a debugging session, it will appear to be a string rather than an Exception object, which would be confusing. Maybe providing a default value for the Exception's `msg` would serve the need here.

```py
In [1]: class Thing(Exception): 
   ...:     def __init__(self, msg=None, *args): 
   ...:         if msg is None: 
   ...:             msg = "My default msg" 
   ...:         super().__init__(msg, *args) 
   ...:                                                                                                                                                                                                                                                                                                                                

In [2]: raise Thing()                                                                                                                                                                         
---------------------------------------------------------------------------
Thing                                     Traceback (most recent call last)
<ipython-input-3-8bca5b9ebda8> in <module>
----> 1 raise Thing()

Thing: My default msg

In [3]: raise Thing("A different msg")                                                                                                                                                        
---------------------------------------------------------------------------
Thing                                     Traceback (most recent call last)
<ipython-input-4-25c6e6d5f226> in <module>
----> 1 raise Thing("A different msg")

Thing: A different msg
```

danielballan (approved):

PR #615, opened 2020-10-09
Project Event Model Configurations

dylanmcreynolds:
This PR primarily enhances projections to add a projection type for capturing Event Descriptor Configuration field. It is tied to a PR in event model: https://github.com/bluesky/event-model/pull/192

In addition to adding this field type, I refactored the project_xarray function, creating a Projector class that handles the mechanics of parsing a projection. It uses callbacks to communicate to a caller. The project_xarray function now uses the Projector class, and is now handling only the construction of a return xarray.Dataset.


dylanmcreynolds:
This PR is very dependent on event model work. I have a little more work to do here to update changes that were recently merged into event model, so let's not merge this until I've completed that.

databroker/projector.py
```diff
@@ -137,54 +259,100 @@ def project_xarray(run: BlueskyRun, *args, projection=None, projection_name=None
...
-            projection_location = mapping.get('location')
-            projection_data = None
-            projection_linked_field = mapping.get('field')
-
-            # single value data that will go in the top
-            # dataset's attributes
-            if projection_location == 'configuration':
-                attrs[field_key] = run.metadata['start'][projection_linked_field]
-                continue
+        projector = Projector(
```
danielballan, line 328: Is this `try` block intended to be part of the `event_field_cb` method? As written it runs once at class definition time. If this "works" it might indicate a need for another unit test.

danielballan (commented):
See some notes in line. Should we mark these as experimental in the docstrings?

dylanmcreynolds:
Do you have a formal format for marking "experimental" or is it as simple as saying it in the docstring?

danielballan:
We just put it in the docstring. There has been a move toward more formality around this in some projects, but it's not yet widespread. Could consider it in the future.

PR #616, opened 2020-10-22
Expose timestamps, configuration, and configuration timestamps in v2.

danielballan:
Recall that devices implement `read()` and `read_configuration()`. Respectively, these produce readings that go into Events and configurational readings that go into Event Descriptors. Each of these readings include both values and associated hardware timestamps.

```py
In [1]: from ophyd import Device, Component

In [2]: from ophyd.sim import Signal, SynSignal

In [3]: import random

In [4]: class Thing(Device):
   ...:     intensity = Component(SynSignal, func=random.random)
   ...:     exposure_time = Component(Signal, value=1, kind='config')
   ...: 

In [5]: thing = Thing(name='thing')

In [6]: thing.read()  # These will go into the Event.
Out[6]: 
OrderedDict([('thing_intensity',
              {'value': 0.008608399314166015, 'timestamp': 1603384540.002283})])

In [7]: thing.read_configuration()  # These will go into the Event Descriptor.
Out[7]: 
OrderedDict([('thing_exposure_time',
              {'value': 1, 'timestamp': 1603384540.0025058})])
```

Let's take data with this and then access it with the v2 API. (Off-screen we created `RE` and hooked it up to a `serializer` connected to Mongo, with a corresponding `catalog`.) 

```py
In [7]: RE(count([thing], 10), serializer)
Out[7]: ('102541b3-bc23-42b8-83d6-a9994178edf5',)

In [8]: run = catalog[-1]

In [11]: run
Out[11]: 
BlueskyRun
  uid='102541b3-bc23-42b8-83d6-a9994178edf5'
  exit_status='success'
  2020-10-22 12:37:20.926 -- 2020-10-22 12:37:21.013
  Streams:
    * primary

In [12]: run.primary
Out[12]: <BlueskyEventStream 'primary' from Run 102541b3...>

In [13]: run.primary.read()
Out[13]: 
<xarray.Dataset>
Dimensions:                    (time: 10)
Coordinates:
  * time                       (time) float64 1.603e+09 1.603e+09 ... 1.603e+09
Data variables:
    thing_intensity            (time) float64 0.3347 0.3347 ... 0.3347 0.3347
    thing:thing_exposure_time  (time) int64 1 1 1 1 1 1 1 1 1 1
    seq_num                    (time) int64 1 2 3 4 5 6 7 8 9 10
    uid                        (time) <U36 '47f9c410-1c0f-478e-be40-550ccb63a...
```

This gives us the values from the Events and the configuration values from the Event Descriptor. It automatically numpy-broadcasts the configuration values so that it is easy to do column arithmetic, such as dividing intensity by exposure. This result also includes the special columns `seq_num` and `uid`, which are also from the Events.

There are major usability problems and technical with this:

* The mixture of Event-wise readings and configuration is potentially confusing, and the mapping between which configuration belongs to which columns is not machine-readable nor necessarily human-readable.
* In real use cases, it quickly becomes very long because some devices record a lot of configuration.
* There is no way to access the hardware timestamps from this API. These are rarely useful for analysis -- more for hardware debugging -- but they should still be accessible.
* There is a potential for name collisions!

The structure of this output was posed early in the development of v2, and it needs to be revisited to address these issues. Because v2 is marked as "experimental" we can make the necessary changes in a minor release, e.g. databroker v1.2.0. This PR:

* Slims down the output of `run.<stream name>.read()` to return only the data from `Event['data']` with no possibility of name collisions. (This is a backward-incompatible change because it removes columns.)
* Adds the accessors `run.<stream name>.config.<object name>`, `run.<stream name>.timestamps`, and `run.<stream name>.config_timestamps.<object name>` to access these other values from the documents.
* Uses xarray's `attrs` dict to provide a machine- and human-readable mapping between each column in the output of `run.<stream name>.read()` and the associated table of configuration readings. For example, if you have multiple detectors, you can take which intensities go with which exposure times.

```py
In [12]: run.primary.read()  # Same usage as above, but a leaner result
Out[12]: 
<xarray.Dataset>
Dimensions:          (time: 10)
Coordinates:
  * time             (time) float64 1.603e+09 1.603e+09 ... 1.603e+09 1.603e+09
Data variables:
    thing_intensity  (time) float64 0.1325 0.1325 0.1325 ... 0.1325 0.1325

In [13]: run.primary().read()['thing_intensity']  # Each column (DataArray) has "Attributes" showing the object (i.e. device) it came from
Out[13]: 
<xarray.DataArray 'thing_intensity' (time: 10)>
array([0.13245665, 0.13245665, 0.13245665, 0.13245665, 0.13245665,
       0.13245665, 0.13245665, 0.13245665, 0.13245665, 0.13245665])
Coordinates:
  * time     (time) float64 1.603e+09 1.603e+09 ... 1.603e+09 1.603e+09
Attributes:
    object:   thing

In [14]: list(run.primary.config)  # one entry for each of EventDescriptor['object_keys'], i.e. ophyd device names
Out[14]: ['thing']

In [15]: run.primary.config.thing.read()  # from EventDescriptor['configuration']['thing']['data']
Out[15]: 
<xarray.Dataset>
Dimensions:              (time: 10)
Coordinates:
  * time                 (time) float64 1.603e+09 1.603e+09 ... 1.603e+09
Data variables:
    thing_exposure_time  (time) int64 1 1 1 1 1 1 1 1 1 1

In [16]: run.primary.config_timestamps.thing.read()  # from EventDescriptor['configuration']['thing']['timestamps']
Out[16]: 
<xarray.Dataset>
Dimensions:              (time: 10)
Coordinates:
  * time                 (time) float64 1.603e+09 1.603e+09 ... 1.603e+09
Data variables:
    thing_exposure_time  (time) float64 1.603e+09 1.603e+09 ... 1.603e+09

In [17]: run.primary.timestamps.read()  # from Event['timestamps']
Out[17]: 
<xarray.Dataset>
Dimensions:          (time: 10)
Coordinates:
  * time             (time) float64 1.603e+09 1.603e+09 ... 1.603e+09 1.603e+09
Data variables:
    thing_intensity  (time) float64 1.603e+09 1.603e+09 ... 1.603e+09 1.603e+09
```

Needs more unit tests

### DIFF
--- a/databroker/core.py
+++ b/databroker/core.py
@@ -1172,6 +1172,11 @@ class BlueskyEventStream(DataSourceMixin):
     exclude : list, optional
         Fields ('data keys') to exclude. By default none are excluded. This
         parameter is mutually exclusive with ``include``.
+    sub_dict : {"data", "timestamps"}, optional
+        Which sub-dict in the EventPage to use
+    configuration_for : str
+        The name of an object (e.g. device) whose configuration we want to
+        read.
     **kwargs :
         Additional keyword arguments are passed through to the base class.
     """
--- a/databroker/core.py
+++ b/databroker/core.py
@@ -27,7 +27,7 @@ import xarray
 from .intake_xarray_core.base import DataSourceMixin
 from .intake_xarray_core.xarray_container import RemoteXarray
 from .utils import LazyMap
-from bluesky_live.bluesky_run import documents_to_xarray, documents_to_xarray_config
+from bluesky_live.conversion import documents_to_xarray, documents_to_xarray_config
 from collections import deque, OrderedDict
 from dask.base import normalize_token
 
--- a/databroker/core.py
+++ b/databroker/core.py
@@ -27,6 +27,7 @@ import xarray
 from .intake_xarray_core.base import DataSourceMixin
 from .intake_xarray_core.xarray_container import RemoteXarray
 from .utils import LazyMap
+from bluesky_live.bluesky_run import documents_to_xarray, documents_to_xarray_config
 from collections import deque, OrderedDict
 from dask.base import normalize_token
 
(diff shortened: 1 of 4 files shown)


danielballan:
A design point that needs more consideration:

Should we add a `data` accessor, as in `run.primary.data.read()` for symmetry? We have received strong feedback that `run.primary.data.read()` is too verbose to be the _only_ way to access data. Compare it v0 and you can imagine the resistance we would have to adoption of v2:

```py
run.table()  # assumes 'primary' which is great unless you don't know to look for the other streams...
```

Nonetheless we could add it as a synonym. Would that be more or less confusing?

dylanmcreynolds:
I was initially a fan of the data accessor for consistency, but have softened on that. While I like the consistency, it is certainly easy to say "hey, the event is the data, and oh hey there's this config thing, too, that you might be interested in."

dylanmcreynolds (approved):

cryos:
I think there is a lot to be said for making the most common/simple use case as simple as possible. Most data I encounter has most interesting stuff in primary, and if there is more you are likely to go looking. If people are already used to `run.table()` it is likely going to be an easier sell if they can simply type `run.read()` to get something very similar. I really like the changes, and the crowding in the read bothered me when I saw it the first few times.

danielballan:
I am sold on `run.primary.read()` but less sure if we should ever implement `run.read()`. To take an example cherished by @cryos, if `run.read()` were an alias for `run.primary.read()` it would return only FXI's images+timestamps but _not_ the angles (which are recorded asynchronously in a separate stream). Experience so far suggests that users are at risk of concluding that the angles were not recorded and writing bluesky hate mail. By forcing them to go a layer down, `run.primary.read()`, we are forcing them to become aware that that layer exists.

I would like to keep thinking about how to give the user's the data in a structure that is immediately useful to them, rather than in the manner that is native to bluesky. This could include returning xarrays with custom coordinates defined (a use case of interest to @ronpandolfi) or even completely different data structures with additional nesting, more akin to an HDF5 group (a use case mentioned by @cryos recently). I have in mind that this would be implemented on top of the "raw" databroker API in some fashion.

cryos:
I think they already assume that the angles were not recorded, and then have to go hunting down a non-obvious path in the case of FXI made no less bad by this change than before. I take your point, but I think the value to the simpler data is relatively high. I can live without it, but would argue that primary is primary because it is first, and returning it by default would not be a bad idea and is already my go to once I learn that primary is always there ;-)

danielballan:
That's fair. It's easier to add such a method convenience method than to take it back once added, and there are other meanings we might assign to it, such as concatenating all the streams into one `xarray.Dataset`. I move to consider this in a separate issue/PR.

cryos:
I agree, and I think ultimately having a default projector or similar that offers an analysis view is likely to be the most useful default to assign to a base `read()` method that gives a simple view of the most complete set of data we can offer for a given run.

danielballan:
Great, a "default projector" is exactly what I have in mind. I suggested this to @ronpandolfi as a possible solution to the requirement of using xarray coordinates in a call, and a message from @dylanmcreynolds today suggests he is thinking along similar lines.

tacaswell:
I am :-1: on `run.read()` because "primary" is just a convention and I do not think we should bake that choice in any more than we already have.  'primary' is not always there and we have at least one plan at SRX that produces data without a primary stream.

50/50 on the data accessor.  :+1: on symmetry :-1: on their being two ways of doing things.

The ergonomics of the "right" data layout is going to be very field and analysis dependent and frequently different than the natural organization for data collection (or output from the previous analysis step) so I think we should worry about getting the projections down right than trying to make this API feel ideal for interactive use (where the brevity matters over clarity).

:+1: overall.

databroker/core.py
```diff
@@ -1379,6 +1215,8 @@ def __init__(self,
                  entry,
                  include=None,
                  exclude=None,
+                 sub_dict="data",
+                 configuration_for=None,
```
gwbischof, line 1232: Add sub_dict, and configuration_for in doc string?

databroker/tests/test_v2/test_core.py
```diff
@@ -3,7 +3,7 @@
 import tempfile
 import xarray
 from ... import core
-from ...core import _documents_to_xarray
+from ...core import documents_to_xarray
```
gwbischof: Maybe import this from bluesky-live?

gwbischof (commented):
Maybe add a test for documents_to_xarray_config?

danielballan:
> Maybe add a test for `documents_to_xarray_config`?

I think it would be easier and better to test `document_to_xarray` and `document_to_xarray_config` in bluesky-live, where they now live. But I will leave the `databroker_to_xarray` tests that we already _have_ here for now to avoid needless thrashing.

PR #623, opened 2020-11-02
Rename 'canonical' to 'documents'.

### LINKED ISSUE #618: Accessing documents in v2
Following up on #616 and #617, another possible API refinement. This one is backward-compatible.

In v0/v1, we have:

```py
run.documents()  # unfilled by default
```

In v2, we have:

```py
run.canonical(fill='no')  # fill is a required keyword argument accepting {'yes', 'no', 'delayed'}
```

**Where did the name 'canonical' come from?** At the beginning, I was trying very hard to play within intake's framework and avoid adding API specific to the Document Model. I had multiple motivations:
* Make our stuff work seamlessly within the large ecosystem of intake drivers.
* Intentionally question the ways in which we were making ourselves "special" compared to standard SciPy/PyData stuff.
* "Learn the rules before you break them."

The name `canonical` was suggested by Martin (of intake) as a way to mean, "Give me this data in its 'canonical' original representation." This could be a notion that, in time, other Catalogs implement too.

However I have come to see some downsides with this:

* It's a somewhat needless name change from v0/v1.
* It's less self-describing.
* If other Catalogs ever implement a `canonical` method, it would be type-unstable and thus of limited use as an interface.
* It's become clear that although `BlueskyRun` is _implemented_ as a `Catalog` subclass, it's not "just another Catalog" in way that `Broker` is. Yes, it's technically a catalog-of-`BlueskyEventStream` objects, but it is in another sense the "leaf" of the "tree", the unit you want to pass around in your program. A `BlueskyEventStream` is rarely useful on its own. Thus, giving `BlueskyRun` special methods that other `Catalog` subclasses do not have seems more acceptable.

**I propose to (re-)introduce `documents` as an alias to `canonical` and to eventually deprecate and remove `canonical`.**

danielballan:
The old method is still supported and issues a warning.

Closes #618. See that issue for context and motivation.

### DIFF
--- a/doc/source/v2/user/index.rst
+++ b/doc/source/v2/user/index.rst
@@ -285,6 +285,11 @@ metadata. (See event-model_.) To access the run---effectively replaying the
 chronological stream of documents that were emitted during data
 acquisition---use the ``documents()`` method.
 
+.. versionchanged:: 1.2.0
+
+   The ``documents`` method was formerly named ``canonical``. The old name is
+   still supported but deprecated.
+
 .. ipython:: python
 
    run.documents(fill='yes')
--- a/doc/source/v2/user/index.rst
+++ b/doc/source/v2/user/index.rst
@@ -283,11 +283,11 @@ Replay Document Stream
 Bluesky is built around a streaming-friendly representation of data and
 metadata. (See event-model_.) To access the run---effectively replaying the
 chronological stream of documents that were emitted during data
-acquisition---use the ``canonical()`` method.
+acquisition---use the ``documents()`` method.
 
 .. ipython:: python
 
-   run.canonical(fill='yes')
+   run.documents(fill='yes')
 
 This generator yields ``(name, doc)`` pairs and can be fed into streaming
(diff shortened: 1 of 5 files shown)


danielballan:
Includes new test that checks that the other methods work and issue a wanring.

doc/source/v2/user/index.rst
```diff
@@ -283,11 +283,11 @@ Replay Document Stream
 Bluesky is built around a streaming-friendly representation of data and
 metadata. (See event-model_.) To access the run---effectively replaying the
 chronological stream of documents that were emitted during data
-acquisition---use the ``canonical()`` method.
+acquisition---use the ``documents()`` method.
```
dylanmcreynolds, line 286: Should we have a note about the rename here?

danielballan: Good suggestion. Added a note in `versionchanged` directive just below this in a new commit.

dylanmcreynolds (approved):

Issue #612, opened 2020-10-09
Local Test Failures


dylanmcreynolds:
I just merge master into a local branch and I'm seeing a couple of test failures: 

```
___________________________________________________________________________ test_time_range ___________________________________________________________________________

    def test_time_range():
        empty = TimeRange()
        assert empty == dict(empty) == empty.query == {}
        assert TimeRange(**empty.kwargs) == empty
    
        since = TimeRange(since='2020')
        expected = {'time': {'$gte': 1577854800.0}}
>       assert since == dict(since) == since.query == expected
E       AssertionError: assert {'time': {'$g...1577865600.0}} == {'time': {'$g...1577854800.0}}
E         Differing items:
E         {'time': {'$gte': 1577865600.0}} != {'time': {'$gte': 1577854800.0}}
E         Use -v to get the full diff

databroker/tests/test_queries.py:15: AssertionError




____________________________________________________________________ test_partial_uid_lookup[hdf5] ____________________________________________________________________

db = <databroker._core.BrokerES object at 0x7efd344bb850>, RE = <bluesky.run_engine.RunEngine object at 0x7efd344dc670>
hw = namespace(ab_det=ABDetector(prefix='', name='det', read_attrs=['a', 'b'], configuration_attrs=[]), bool_sig=Signal(nam...(name='signal', value=0, timestamp=1602261944.045641), trivial_flyer=<ophyd.sim.TrivialFlyer object at 0x7efd343dac70>)

    def test_partial_uid_lookup(db, RE, hw):
        RE.subscribe(db.insert)
    
        # Create enough runs that there are two that begin with the same char.
        for _ in range(50):
            RE(count([hw.det]))
    
        with pytest.raises(ValueError):
            # Some letter will happen to be the first letter of more than one uid.
            for first_letter in string.ascii_lowercase:
>               db[first_letter]

databroker/tests/test_broker.py:246: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
databroker/_core.py:1083: in __getitem__
    for start, stop in search(key, self)]
[...]

PR #624, opened 2020-11-02
Remove ET bias from TimeRange tests.

danielballan:


danielballan:
Verified locally by @dylanmcreynolds in Pacific Time. Addresses part of #612.

PR #626, opened 2020-11-09
Revert workarounds to intake issues fixed upstream.

danielballan:
Specifically, intake#543 and intake#545 were both addressed upstream. Because
these issues were fixed before they were released, there is no reason to
retain workarounds in databroker.

### DIFF
--- a/databroker/core.py
+++ b/databroker/core.py
@@ -874,10 +874,6 @@ class BlueskyRun(intake.catalog.Catalog):
         Additional keyword arguments are passed through to the base class,
         Catalog.
     """
-    # Work around
-    # https://github.com/intake/intake/issues/545
-    _container = None
-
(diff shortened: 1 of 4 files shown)


danielballan:
Note to selves: let's pay careful attention to all the CI builds here.

* Builds against intake 0.6.0 should pass if it's accurate that 0.6.0 was released _before_ they issues.
* Builds against intake master should pass if it's accurate that all of the issues have been fixed upstream.

danielballan:
We have hit Travis' limit on databroker. https://travis-ci.com/github/bluesky/databroker

danielballan:
Intake didn't quite fix intake#543 yet. The rest of the fix is proposed here: https://github.com/intake/intake/pull/549

PR #627, opened 2020-11-12
Handle no stop doc in repr.

danielballan:
Needs tests.

### DIFF
--- a/databroker/core.py
+++ b/databroker/core.py
@@ -819,7 +819,7 @@ class RemoteBlueskyRun(intake_RemoteCatalog):
         try:
             self._load()
             start = self.metadata['start']
-            stop = self.metadata['stop']
+            stop = self.metadata['stop'] or {}
             out = (f"BlueskyRun\n"
                    f"  uid={start['uid']!r}\n"
                    f"  exit_status={stop.get('exit_status')!r}\n"
@@ -969,7 +969,7 @@ class BlueskyRun(intake.catalog.Catalog):
         try:
             self._load()
             start = self.metadata['start']
-            stop = self.metadata['stop']
+            stop = self.metadata['stop'] or {}
             out = (f"BlueskyRun\n"
                    f"  uid={start['uid']!r}\n"
                    f"  exit_status={stop.get('exit_status')!r}\n"


danielballan:
Power-cycled to run against new GitHub Actions CI.

danielballan:
It turns out our test harness makes this really challenging to test. I'd like to punt on a unit test until we sort out a better test harness (leveraging bluesky-live). Here's an interactive test to prove that this works.

```py
# generate_data.py
import logging
from suitcase.mongo_normalized import Serializer
from bluesky import RunEngine
from bluesky.plans import count
from ophyd.sim import det
import uuid

from databroker._drivers.mongo_normalized import BlueskyMongoCatalog


RE = RunEngine()

mds = f'mongodb://localhost:27017/databroker-test-{uuid.uuid4()}'
fs = f'mongodb://localhost:27017/databroker-test-{uuid.uuid4()}'

serializer = Serializer(mds, fs)

def serializer_that_fails_to_save_stop_doc(name, doc):
    if name == "stop":
        # Drop it on the floor.
        return
    serializer(name, doc)

RE(count([det]), serializer_that_fails_to_save_stop_doc)

catalog = BlueskyMongoCatalog(mds, fs)

from IPython.display import display

from IPython.core.formatters import PlainTextFormatter
f = PlainTextFormatter()
print(f(catalog[-1]))
```

Output before fix (current `master`):

```
<BlueskyRun *REPR_RENDERING_FAILURE* AttributeError("'NoneType' object has no attribute 'get'")>
```

Output with fix (this PR):

```
BlueskyRun
  uid='44a9be60-de22-41da-9ca7-079bd08e0399'
  exit_status=None
  2020-12-17 15:52:08.440 -- ?
  Streams:
    * primary
```

databroker/core.py
```diff
@@ -819,7 +819,7 @@ def _repr_pretty_(self, p, cycle):
         try:
             self._load()
             start = self.metadata['start']
-            stop = self.metadata['stop']
+            stop = self.metadata['stop'] or {}
```
dylanmcreynolds, line 822: If there's really no stop document, won't this raise a key error instead of returning {}?

danielballan: The `metadata` dict is initialized with `'stop'` set to `None` in `_load()` (which is called at `__init__` time) if the stop doc does not exist yet. It's arguable that leaving it empty would be better, but that would be a backward-incompatible change.


https://github.com/bluesky/databroker/blob/6b7e31b6cc399686df926fec51e43193d488791f/databroker/core.py#L987-L1001

dylanmcreynolds: I see, cool.

PR #628, opened 2020-11-16
Summary projector

dylanmcreynolds:
This PR primarily creates a simplified projector. Before this, the only projector available was the project_xarray function which is comprehensive and builds up an xarray.Dataset with data from the full run.

That is pretty heavy weight, and potentially overkill for applications when want two things:

- a quick summary of a few metadata fields when presenting a list of runs
- projected fields

So, this uses the same rules for selecting a projection that the xarray projector uses, but only projects metadata from the start document. It also has a parameter that allows callers to send a list of just a few fields that they want. I had support of GraphQL in mind there.

Maybe we want to add stop document info too? A summary might want to know whether runs completed and how many events they have.

### DIFF
--- a/databroker/projector.py
+++ b/databroker/projector.py
@@ -208,7 +208,7 @@ class Projector():
                 if self._metadata_cb:
                     value = run.metadata['start'].get(projection_linked_field)
                     if value is None:
-                        self.issues.append(f"{run['start']['uid']} Start key misising in run {field_key}: {projection_linked_field}")
+                        self.issues.append(f"{run.metadata['start']['uid']} Start key misising in run {field_key}: {projection_linked_field}")
                         continue
                     self._metadata_cb(field_key, value)
                 continue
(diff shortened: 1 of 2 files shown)


dylanmcreynolds:
I also snuck some changes for consideration into the projectors related to error handling. In previous projectors, errors found during the course of projection raised issues immediately, suspending processing and raising to the caller. This proved really difficult to deal with in a server environment. Now, the projector returns an "issues" list along with it's payload which can be interrogated by the caller. I'm a little concerned about the API breakage, but it works really well over in splash land.

databroker/projector.py
```diff
@@ -137,6 +140,11 @@ def __init__(self, metadata_cb=None, event_configuration_cb=None, event_field_cb
         self._metadata_cb = metadata_cb
         self._event_configuration_cb = event_configuration_cb
         self._event_field_cb = event_field_cb
+        self._issues = []
+
+    @property
+    def issues(self):
```
danielballan, line 146: Per discussion, maybe we want an array of `("msg", exception)` pairs here.

PR #629, opened 2020-12-02
FIX: do not error our when checking comparison with non-Header

tacaswell:


danielballan:
Power-cycled to run against new GitHub Actions CI.

danielballan (changes requested):
Good idea. Test failures look real.

danielballan:
Found the issue, fixed it for you.

danielballan (approved):

tacaswell:
Thanks!

PR #642, opened 2021-01-22
Rework docs

danielballan:
I restructured and mostly rewrote the docs. I structured it according to [this recommendation](https://documentation.divio.com/), inspired by how well it works for [scanspec](https://dls-controls.github.io/scanspec/master/index.html).

This draft is published as HTML at https://blueskyproject.io/databroker/docs-rewrite-draft/index.html for easy review. (I pushed it manually.)

TO DO

- Squash a couple remaining sphinx warnings.
- Take a pass at the "Relationship to Intake" page, which is a bit stale.
- Take a pass at the v2 API docs, which leak some intake methods like `persist` that we don't want to support at this time.


thomascobb:
Did you spot the corresponding CSS for the column layout?
https://github.com/dls-controls/scanspec/blob/8aed8ebdfeb75062ff8cb7d5e36dd98e1e3c3da1/docs/_static/theme_overrides.css#L20-L33

danielballan:
I did not. Thanks!

docs/Makefile
```diff
@@ -2,7 +2,7 @@
 #
 
 # You can set these variables from the command line.
-SPHINXOPTS    = -W
+# SPHINXOPTS    = -W
```
tacaswell:
```suggestion
SPHINXOPTS    = -W
```

docs/source/explanations/v2-transition.rst
```diff
@@ -0,0 +1,91 @@
...
+some time after 2015. Databroker 1.x provides three public interfaces on top of
+these two implementations.
+
+=========== ========= ============== ==========================
+API version Interface Implementation Who should use it?
+=========== ========= ============== ==========================
+v2          New       New            All new users
+v1          Original  New            Users with old scripts that use original interface
+v0          Original  Original       Users who hit bugs in v1/v2 and need a fallback
+=========== ========= ============== ==========================
```
tacaswell:
```suggestion
=========== ========= ============== ====================================================
API version Interface Implementation Who should use it?
=========== ========= ============== ====================================================
v2          New       New            All new users
v1          Original  New            Users with old scripts that use original interface
v0          Original  Original       Users who hit bugs in v1/v2 and need a fallback
=========== ========= ============== ====================================================
```

docs/Makefile
danielballan: Don't worry, that's in a `TMP: ...` commit. Just turned off while I applied the wrecking ball..

docs/source/tutorials/get-metadata.rst
```diff
@@ -0,0 +1,190 @@
...
+
+TO DO: Obtain an example catalog that has some failures in it so that this
+example is not so trivial.
+
+Low-level Hardware Timestamps
+-----------------------------
+
+.. note::
+
+   We mention this for completeness, but you will probably never need it.
```
tacaswell: This seems too strong, I think using these for (~.1s precision) alignment should be fair-game.

tacaswell:
Nice work!

My strongest issue is with the wording around timestamps.

untzag:
Thanks so much @danielballan this is wonderful to see :heart: my humble comments follow

Suggestion: glossary of terms. Run, catalog, document... what EXACTLY does databroker mean by these? You have some content in [databroker-concepts](https://blueskyproject.io/databroker/docs-rewrite-draft/explanations/relationship-to-intake.html#databroker-concepts) that would be useful for this.

Comment: the docs are written very-much with bluesky in mind (understandably). As a challenge question, how could databroker be useful for non-bluesky users? What kinds of information would be useful to developers using databroker without bluesky?

Suggestion: List some more context about how data was generated for each of the tutorials. I'd like to know what hardware participated and what bluesky plan was used in each case. This context would help me understand the stored data.

Suggestion: Recommend mongo via docker.

Question: Do the v0/v1/v2 differences only appear in the data access interface? It seems like there hasn't actually been a change to the write interface. If this is true, it would be a helpful clarifying detail under "What are the API versions v0, v1, v2?"

Comment: you do not document the low-level data model at all. Consider making a page that talks about the data model and relates it explicitly to the event model.

dylanmcreynolds:
@untzag has an interesting point databroker without runengine. That's seems uncharted territory that we're figuring out as we go along. It took me many months to guess at what a strategy might be, but it certainly involves ingestors of various sorts. The details maybe belong more in event model, but something high level here would be good.

untzag:
On slack, @danielballan helped me understand some more of the context. I suggest putting the following context right up front (in the index page, for example)

Databroker is a data ACCESS tool for BLUESKY DOCUMENTS. Consider linking to the event model to explain exactly what the scope of bluesky documents is.

Databroker is developed in concert with SUITCASE. SUITCASE does data writing, and databroker is an access layer (mongo -> databroker -> intake).

This context would really help the clueless such as myself :smile:

danielballan:
Thank you all, these are extremely useful comments. I will address the quick ones now and open issues to address the more labor-intensive ones in follow-up PRs.

I have no experience deploying MongoDB in Docker. Is that how you deploy it, @untzag? A quick Google search didn't turn up any _official_ documentation on deploying MongoDB **Community Edition** in Docker, but there are plenty on blog posts on it. Is there any particular resource that you'd recommend?

untzag:
@danielballan I've just been using the official docker image: https://hub.docker.com/_/mongo and so far "it just works"

to be 100% honest I don't run databroker in production... anywhere, so don't trust my advice too much

ksunden:
https://github.com/docker-library/mongo/blob/master/4.4/Dockerfile#L73-L84

Just a note that the "official" here does not refer to "by mongo", but rather "by docker"

It is indeed the community edition by default (though build time option for enterprise, pretty sure the one on dockerhub is community edition)

danielballan:
Perfect, thanks. I'll add a mention of and link to that.

tacaswell:
If you are going to run mongo in a container, make sure you are bind mounting the volumes where the mongo files live in rather than letting docker own those volumes.

tacaswell:
That said, for bluesky-pods we do just blindly start mongo containers https://github.com/bluesky/bluesky-pods/blob/1cc5e700f2aa1cef24b00bf15781cbe652c0a7c3/start_databroker_pod.sh#L11 and https://github.com/bluesky/bluesky-pods/blob/1cc5e700f2aa1cef24b00bf15781cbe652c0a7c3/start_acqusition_pod.sh#L25

dylanmcreynolds:
If you were deploying mongo in a production databroker environment, you will also want to require authentication in mongo and "hide" those passwords in the intake configuration: https://intake.readthedocs.io/en/latest/catalog.html?highlight=environment%20variable#templating , which works pretty well for containers/mongo/databroker.

danielballan:
If you have a chance to flip through the latest commits, @untzag and @dylanmcreynolds I would welcome any word-smithing suggestions.

docs/source/explanations/v2-transition.rst
danielballan: Applied manually

gwbischof (approved):

PR #646, opened 2021-02-18
Pin intake below 0.6.1.

danielballan:
We see at least one critical bug against intake 0.6.1. This is a short term hotfix.


tacaswell:
Would it be better to pin `!=0.6.1` ?  That way we don't have to un-do it later.

danielballan:
We have a workaround in #647 so this is not needed. We may have to revisit it if we find more regressions, but I'll close for now.

PR #647, opened 2021-02-18
MergedCatalog is no longer working due to intake change

gwbischof:
With the latest release of intake 0.6.1 we see this problem:

image

It should behave like this:

image

This PR fixes the problem.


gwbischof:
This PR also changes the sqlite connection timeout to fix a problem with tests failing intermittently

PR #648, opened 2021-02-18
v1.2.1 release notes

gwbischof:
v1.2.1 release notes


danielballan:
Two of the Unit Tests builds failed, but all the failures were the usual sqlite TimeoutError that is now addressed in `master`.

PR #666, opened 2021-06-28
WIP: Get columns

tacaswell:
Work to address extreme

### DIFF
--- a/databroker/mongo_normalized.py
+++ b/databroker/mongo_normalized.py
@@ -327,6 +327,16 @@ class DatasetFromDocuments:
         descriptor, *_ = self.metadata["descriptors"]
         data_vars = {}
         dim_counter = itertools.count()
+        # Collect the keys (column names) that are of unicode data type.
+        unicode_keys = []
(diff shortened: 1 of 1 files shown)


danielballan:
Tested interactively on filled and unfilled data at CSX.

Timing for reading a table with dimensions 2 rows x 269 columns is greatly improved by about 5X. The absolute time (~400ms) is acceptable.

Before (at commit b13dd52e390a837c57bc149e38d6eabc6da4adf5):

```py
In [4]: %timeit h.table("baseline")
2.22 s ± 31.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

After (at commit cfeedbc5488ca2355d3e7183cebe3c860bfdc525):

```py
In [4]: %timeit h.table("baseline")
406 ms ± 2.31 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

danielballan:
Note: Unit tests are not even _close_ to passing on `v2.0.0-dev`. We're focusing on integrations tests with real beamline data during the design process.

PR #674, opened 2021-09-29
Add replay script

AbbyGi:
## Description

Adds a script that replays Bluesky runs with the correct timing.

## Motivation and Context

This simulates live streaming data so we can develop and test code for streaming data without needing to be at a beamline looking at new data.

## How Has This Been Tested?

This has been tested interactively while developing GUI applications.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -17,7 +17,6 @@ jobs:
     strategy:
       matrix:
         python-version: [3.7, 3.8, 3.9]
-        intake-version: ["master", "latest"]
       fail-fast: false
     steps:
 
@@ -37,14 +36,6 @@ jobs:
       run: |
         set -vxeuo pipefail
         python -m pip install -r requirements-test.txt
-        python -m pip list
-
-    - name: Install intake mater for some builds.
-      if: ${{ matrix.intake-version == 'master' }}
-      shell: bash -l {0}
-      run: |
-        set -vxeuo pipefail
-        python -m pip install git+https://github.com/intake/intake.git@master
         python -m pip list
 
     - name: Test with pytest
--- a/continuous_integration/scripts/install.sh
+++ b/continuous_integration/scripts/install.sh
@@ -9,5 +9,5 @@ python -m pip install --upgrade pip setuptools wheel numpy
 # Versioneer uses the most recent git tag to generate __version__, which appears
 # in the published documentation.
 git fetch --tags
-python -m pip install .
+python -m pip install .[all]
 python -m pip list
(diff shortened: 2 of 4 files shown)


gwbischof:
How are Kafka and ZMQ used here?

gwbischof:
Could `replay` take a list of generators? So I can simulate multiple runs from different beamlines at the same time? Or maybe we should make a new function that replays multiple generators?

tacaswell:
You probably want to lift the time sorting logic than merges documents from multiple collections into a time ordered stream in databroker and write a function like

```python
def merge_runs(*runs):
    for name, doc in sort_and_collate(runs):
        yield name, doc
```

gwbischof:
There is this function here that does something like that: https://github.com/bluesky/databroker/blob/970e9148dfab5e77101d40f059ecb30d064eac81/databroker/core.py#L436

Which reminds me, that we may need to interlace the event/datum_pages for a single run if we want them to be in the right order. But if the generator is canonical() then that should already be done.

danielballan:
> How are Kafka and ZMQ used here?

The CLI portion of this (in `main()`) is designed to replay the documents over the network, via 0MQ or Kafka, so that can be consumed by some other process. The underlying function with the "brains" (`replay`) can also be used directly in process, without a CLI or network, if that's better for your use case.

> Could replay take a list of generators? So I can simulate multiple runs from different beamlines at the same time? 

I see a way to do this that is simpler and also more realistic: a separate process per beamline, each running `replay` on one run at a time, and each (naturally) making reference to the same clock. With a Run, there are strict ordering guarantees, but across beamlines, there is a little more slip in the order of things, and the relative timing does not have to be exactly perfect. Assuming your goal is load-testing Mongo, hitting it with ~28 processes all writing concurrently is more realistic that funneling everything through one process.

> Which reminds me, that we may need to interlace the event/datum_pages for a single run if we want them to be in the right order. But if the generator is canonical() then that should already be done.

This now happens here. It turns out it's a lot simpler to do this at the single-document level --- straight from Mongo before anythign could have been grouped into pages.

https://github.com/bluesky/databroker/blob/c1f67cfd2a6e882ad0879c97a14a3dca6b76185b/databroker/mongo_normalized.py#L327-L381

Replay can count on the documents coming out of databroker in strict time order.

mrakitin:
Is this PR relevant to these items?
- https://github.com/bluesky/event-model/issues/150
- https://github.com/bluesky/event-model/pull/155
- https://github.com/bluesky/event-model/issues/156

danielballan:
Certainly relevant to https://github.com/bluesky/event-model/issues/150 and https://github.com/bluesky/event-model/pull/155. We may need to think about how something in event-model itself (or bluesky-live?) would be useful, in addition to this tool in databroker. Probably not? Needs though.

PR #677, opened 2021-10-08
Get the CI passing

gwbischof:


danielballan:
This seems to pass, but I'm confused why. The new RunEngine test fixture, as of bluesky v1.8.0 is parameters to return `uids` (the current default in bluesky) _or_ `RunEngineResult`. This PR seems to switch to expect `RunEngineResult`, but I would think you'd get that from only half the tests.

https://github.com/bluesky/bluesky/pull/1464/files#diff-a60ccb9dc99370e861dfdbbe7ac82aad946ee3a2db31c38debecdcb9a3284f7eR9

gwbischof:
We are only getting `RunEngineResult`.   I agree from the code you shared it looks like we should be getting both.

danielballan:
This is due to a bug in `RE` fixture. It _intends_ to parameterize on `[True, False]` but it in fact parameterizes on the strings `["True", "False"]` which is, of course, always `True`.

https://github.com/bluesky/bluesky/pull/1464/files#diff-a60ccb9dc99370e861dfdbbe7ac82aad946ee3a2db31c38debecdcb9a3284f7eR9

When the bug is fixed, this PR will break. Can you check `RE._call_returns_result` and adjust accordingly? We should add a _public_ property, but there doesn't seem to be one yet.

tacaswell:
Thank you! Merging over the broken docs get v2 work going.

tacaswell:
This got cherry-picked to the v2.0.0-dev branch as 9b1351dcc7fca992e75c18080ded2b9fe0329b8c

PR #682, opened 2021-12-30
DOC: corrected the same typo a few times.

clintonroy:


mrakitin (approved):
Thanks for the fixes, @clintonroy! The CI failures may be irrelevant.

tacaswell:
Agree the failures look un-related, but they are very worrying...

PR #685, opened 2022-01-20
don't use intake v0.6.5, and get the CI passing

gwbischof:
Intake 0.6.5 seems to have a bug in it causing errors in databroker.
So the CI run against intake master is expected to fail.
See issue https://github.com/intake/intake/issues/643

Pymongo 4 dropped `Collection.insert()` so we have to use `Collection.insert_one()` or `Collection.insert_many()`

### DIFF
--- a/databroker/assets/core.py
+++ b/databroker/assets/core.py
@@ -116,7 +116,7 @@ def bulk_insert_datum(col, resource, datum_ids,
             apply_to_dict_recursively(datum, sanitize_np)
             yield datum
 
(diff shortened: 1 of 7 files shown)


tacaswell:
We should either drop the intake master branch tests or sort out how to provide that missing attribute.

mrakitin (approved):

mrakitin:
> We should either drop the intake master branch tests or sort out how to provide that missing attribute.

+1 for dropping the tests against intake master.

PR #693, opened 2022-03-08
Hacks to get databroker github actions working

clintonroy:
I've removed things that I just couldn't get working straight away:
 * python3.9, seems to have problems with sqlite, i don't understand the code at all
 * no longer installing intake master; i don't understand why intake tests would be run here, or why we wouldn't run the pinned version, rather than master.

Still, maybe this would be a useful place to start?

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -16,7 +16,7 @@ jobs:
         - 27017:27017
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9]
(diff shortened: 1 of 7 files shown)


tacaswell (approved):

danielballan:
The intake `master` tests were designed as an early warning system for incoming changes to intake that could break us on the next intake release. As we're moving off of intake and onto tiled in 2.0.0, this is as important.

The rest we can resolve in the future---it's good to have un-broken CI!

PR #696, opened 2022-04-06
Add new storage backend that supports writing via client

jmaruland:


danielballan:
Suggestions for testing...

Add a new method to `MongoAdapter`:

```py
@classmethod
def from_mongomock(cls, **kwargs)
    # This is similar to the classmethod from_uri but instead of connecting to
    # a real database it creates a mock, temporary one to use.
    client = mongomock.MongoClient()
    database = client["mock_database"]
    return cls(database, **kwargs)
```

Make a test file:

```py
# databroker/tests/test_experimental.py

from tiled.client import from_tree

def test_saving_and_retreiving(tmpdir):
    tree = MongoAdapter.from_mongomock(directory=tmpdir)
    client = from_tree(tree)
    # Do things similar to the example from the recon gist.
```

databroker/experimental/apischema_schemas.py
```diff
@@ -0,0 +1,146 @@
...
+            r"[?:@&=])*)(?:/(?:(?:(?:[a-zA-Z\d$\-_.+!*'(),]|(?:%[a-fA-F\d]{2}))|[?:@&=])*))*)(?:;type=[AIDaid])?)?)",
+            re.IGNORECASE,
+        )
+        if not regex.search(self.data_url):
+            raise ValidationError(
+                "the link provided in data_url is not a compatible url"
+            )
+
+
+if __name__ == "__main__":
```
danielballan: This block was probably used for _ad hoc_ testing. I think it should be removed before merging.

databroker/experimental/client_ext.py
```diff
@@ -0,0 +1,67 @@
+import base64
```
danielballan: We don't need this module anymore; it will be handled by tiled itself.

databroker/experimental/config.yml
```diff
@@ -0,0 +1,8 @@
+authentication:
+  allow_anonymous_access: true
+trees:
+  - path: /
+    tree: "server_ext:MongoAdapter.from_uri" # "server_ext:MongoAdapter.from_uri"
```
danielballan: This will only work if the current working directory is `databroker/experimental`. Use the full module path `databroker.experimental.server_ext` so that it works from anywhere.

databroker/experimental/config.yml
```diff
@@ -0,0 +1,8 @@
...
+    args:
+      uri: mongodb://localhost:27017/fxi-recon-tiled-dev-5
```
danielballan: I wonder, instead of hard-coding example database here, we should use `MongoAdapter.from_mongomock` so that it constructs a temporary database and does not require MongoDB to be running unsecured on `localhost`.

databroker/experimental/config.yml
```diff
@@ -0,0 +1,8 @@
...
+      directory: C:/Users/jmaruland/Documents/MongoDB/tmp/fxi-recon-tiled-dev  # writeable directory
```
danielballan: Definitely don't want to commit this. :-)

databroker/experimental/server_ext.py
```diff
@@ -0,0 +1,503 @@
...
+                dask.dataframe.from_pandas(
+                    pandas.read_parquet(path),
+                    npartitions=self.doc.structure.macro.npartitions,
+                )
+            )
+
+        elif self.doc.data_blob is not None:
+            self.dataframe_adapter = DataFrameAdapter(
+                dask.dataframe.from_pandas(
+                    deserialize_arrow(base64.b64decode(self.doc.data_blob)),
```
danielballan: Reminder to remove base64 here (and from the tiled PR).

danielballan (changes requested):

databroker/experimental/server_ext.py
```diff
@@ -0,0 +1,499 @@
...
+
+from sys import platform
+
+
+class PostMetadataRequest(pydantic.BaseModel):
+    structure_family: StructureFamily
+    structure: Union[ArrayStructure, DataFrameStructure]
+    metadata: Dict
+    specs: List[str]
+    mimetype: str
```
danielballan: After my recent revisions, the client does not include a `mimetype` in the body of its request. The `mimetype` that the data is stored in is up to the server, not the client.

databroker/experimental/server_ext.py
```diff
@@ -0,0 +1,499 @@
...
+
+@router.post("/node/metadata/{path:path}", response_model=PostMetadataResponse)
+def post_metadata(
+    request: Request,
+    body: PostMetadataRequest,
+    entry=Security(entry, scopes=["write:data", "write:metadata"]),
+):
+    if not isinstance(entry, MongoAdapter):
+        raise HTTPException(
+            status_code=404, detail="This path cannot accept reconstruction metadata."
```
danielballan: Let's change this error message to "This node is not writable."

databroker/experimental/server_ext.py
```diff
@@ -0,0 +1,499 @@
...
+        body.structure.micro.meta = deserialize_arrow(
+            base64.b64decode(body.structure.micro.meta)
+        )
+
+    uid = entry.post_metadata(
+        metadata=body.metadata,
+        structure_family=body.structure_family,
+        structure=body.structure,
+        specs=body.specs,
+        mimetype=body.mimetype,
```
danielballan: The `mimetype` parameter should be removed here too. It's up to `entry` what MIME type to use.

databroker/experimental/server_ext.py
```diff
@@ -0,0 +1,499 @@
...
+            directory=self.directory,
+            metadata=metadata,
+            queries=queries,
+            sorting=sorting,
+            access_policy=self.access_policy,
+            principal=principal,
+            **kwargs,
+        )
+
+    def post_metadata(self, metadata, structure_family, structure, specs, mimetype):
```
danielballan: Remove `mimetype`.

databroker/experimental/server_ext.py
```diff
@@ -0,0 +1,499 @@
...
+        uid = str(uuid.uuid4())
+
+        validated_document = Document(
+            uid=uid,
+            structure_family=structure_family,
+            structure=structure,
+            metadata=metadata,
+            specs=specs,
+            mimetype=mimetype,
```
danielballan: `mimetype` should be defined above based on the structure. We always save array as `application/octet-stream` and we always save dataframe as `APACHE_ARROW_FILE_MIME_TYPE`.

databroker/experimental/server_ext.py
```diff
@@ -339,7 +335,13 @@ def new_variation(
             **kwargs,
         )
 
-    def post_metadata(self, metadata, structure_family, structure, specs, mimetype):
+    def post_metadata(self, metadata, structure_family, structure, specs):
+
+        mime_structure_association = {
```
danielballan, line 278: I think this could be at global scope. No reason to re-define it on every call of `post_metadata`.

databroker/experimental/server_ext.py
```diff
@@ -0,0 +1,501 @@
...
+        database,
+        directory,
+        queries=None,
+        sorting=None,
+        metadata=None,
+        principal=None,
+        access_policy=None,
+    ):
+        self.database = database
+        self.collection = database["reconstructions"]
```
kleinhenz: maybe just take the collection object in the constructor instead of the database client that way you don't have to decide on a fixed collection name.

jmaruland: @kleinhenz Thank you for catching this. That was a total miss on my part.
@danielballan, do you thin it a good idea if we add this option to the config file?

databroker/experimental/server_ext.py
```diff
@@ -0,0 +1,501 @@
...
+        elif doc["structure_family"] == StructureFamily.dataframe:
+            return WritingDataFrameAdapter(self.collection, self.directory, doc)
+        else:
+            raise ValueError("Unsupported Structure Family value in the databse")
+
+    def __iter__(self):
+        # TODO Apply pagination, as we do in Databroker.
+        for doc in list(
+            self.collection.find(
+                # self._build_mongo_query({"active": True}), {"uid": True}
```
kleinhenz: why is this commented out? The below solution relies on `data_url` being defined which is not as generic as the document model which allows either `data_url` or `data_blob`.

kleinhenz: should `data_blob` be removed from `Document`?

jmaruland: That was dragged from the initial draft. Now, that I made enough test with the new document for the pydantic model, we are not using the active parameter anymore. I will remove this comment and other like these because they are outdated and not useful for this new approach

databroker/experimental/server_ext.py
kleinhenz: why is this resolved? I don't think it makes sense to base64 encode in mongo.

kleinhenz: Also if the document model supports `data_blob` should there be a way to make the writing adapters use this?

databroker/experimental/server_ext.py
```diff
@@ -0,0 +1,501 @@
...
+            self.dataframe_adapter = DataFrameAdapter(
+                dask.dataframe.from_pandas(
+                    pandas.read_parquet(path),
+                    npartitions=self.doc.structure.macro.npartitions,
+                )
+            )
+
+        elif self.doc.data_blob is not None:
+            self.dataframe_adapter = DataFrameAdapter(
+                dask.dataframe.from_pandas(
```
kleinhenz, line 145: making a new comment since the previous conversation is hidden since it's marked as resolved.

Why do you assume the data blob is base64 encoded?
Also if the document model supports `data_blob` should there be a way to make the writing adapters use this for writing.

databroker/experimental/server_ext.py
jmaruland: > why is this resolved? I don't think it makes sense to base64 encode in mongo.

I will remove this part. It was not checked by the tests because `data_blob` is not being implemented yet. We left a space for it in the Document for our next steps of development. Dan was thinking on letting the server decide whether writing the data in the document in `data_blob` or save it in disc and link it with `data_url`, depending on the size of the data

databroker/experimental/server_ext.py
jmaruland: We prioritized `data_url` for now because we can say that the aimm project uses datasets with small files. We left the space to extend the use of `data_blob` in the next deployment

databroker/experimental/apischema_schemas.py
```diff
@@ -0,0 +1,117 @@
+from apischema import ValidationError, validator
```
kleinhenz: this file is now unused?

databroker/experimental/example_array.py
```diff
@@ -0,0 +1,25 @@
+import numpy
```
kleinhenz: this file should probably be either deleted or moved into tests

databroker/experimental/example_dataframe.py
```diff
@@ -0,0 +1,34 @@
+import numpy
```
kleinhenz: this file should probably be either deleted or moved into tests

databroker/experimental/queries.py
```diff
@@ -0,0 +1,24 @@
...
+
+    query: str  # We cannot put a dict in a URL, so this a JSON str.
+
+    def __init__(self, query):
+        if not isinstance(query, str):
+            query = json.dumps(query)
+        self.query = query
+
+
+def scan_id(number):
```
kleinhenz: maybe this belongs somewhere else? It looks like it is only used in tests.

databroker/experimental/schemas.py
```diff
@@ -0,0 +1,181 @@
...
+                "video",
+            ]
+        )
+
+        if m_type not in mime_type_list:
+            raise ValueError(f"{m_type} is not a valid mime type")
+        return v
+
+
+def try_array_schema():
```
kleinhenz: delete or move to tests

databroker/experimental/schemas.py
```diff
@@ -0,0 +1,181 @@
...
+        uid=name,
+        structure_family=structure_family,
+        structure=structure,
+        metadata=metadata,
+        specs=specs,
+        mimetype=mimetype,
+    )
+
+
+def try_dataframe_schema():
```
kleinhenz: delete or move to tests

databroker/experimental/server_ext.py
```diff
@@ -0,0 +1,501 @@
...
+    def inner(self, *args, **kwargs):
+        if self.dataframe_adapter is None:
+            raise ValueError("Not active")
+        else:
+            return method(self, *args, **kwargs)
+
+    return inner
+
+
+class WritingArrayAdapter:
```
kleinhenz, line 59: Why not just subclass ArrayAdapter?

databroker/experimental/server_ext.py
```diff
@@ -0,0 +1,501 @@
...
+            {"uid": self.doc.uid},
+            {
+                "$set": {
+                    "data_url": f"file://localhost/{str(path).replace(os.sep, '/')}"
+                }
+            },
+        )
+
+
+class WritingDataFrameAdapter:
```
kleinhenz, line 122: Same question as above. Why not just subclass?

kleinhenz:
running this locally the openapi documentation appears broken.

The error looks a bit obscure.

```
  File "/home/joseph/.local/opt/miniconda3/envs/aimm/lib/python3.9/site-packages/fastapi/utils.py", line 28, in get_model_definitions
    m_schema, m_definitions, m_nested_models = model_process_schema(
  File "pydantic/schema.py", line 617, in pydantic.schema.model_process_schema
  File "pydantic/schema.py", line 658, in pydantic.schema.model_type_schema
  File "pydantic/schema.py", line 258, in pydantic.schema.field_schema
  File "pydantic/schema.py", line 498, in pydantic.schema.field_type_schema
  File "pydantic/schema.py", line 925, in pydantic.schema.field_singleton_schema
  File "pydantic/schema.py", line 617, in pydantic.schema.model_process_schema
  File "pydantic/schema.py", line 658, in pydantic.schema.model_type_schema
  File "pydantic/schema.py", line 258, in pydantic.schema.field_schema
  File "pydantic/schema.py", line 563, in pydantic.schema.field_type_schema
  File "pydantic/schema.py", line 848, in pydantic.schema.field_singleton_schema
  File "pydantic/schema.py", line 748, in pydantic.schema.field_singleton_sub_fields_schema
  File "pydantic/schema.py", line 563, in pydantic.schema.field_type_schema
  File "pydantic/schema.py", line 922, in pydantic.schema.field_singleton_schema
  File "/home/joseph/.local/opt/miniconda3/envs/aimm/lib/python3.9/abc.py", line 123, in __subclasscheck__
    return _abc_subclasscheck(cls, subclass)
TypeError: issubclass() arg 1 must be a class
```

I checked that I don't get this for aimmdb or the basic tiled example tree so something is going on. Can anyone else reproduce?

databroker/experimental/server_ext.py
```diff
@@ -0,0 +1,501 @@
...
+    def microstructure(self):
+        return self.array_adapter.microstructure()
+
+    def macrostructure(self):
+        return self.array_adapter.macrostructure()
+
+    def put_data(self, body):
+        # Organize files into subdirectories with the first two
+        # charcters of the uid to avoid one giant directory.
+        path = self.directory / self.doc.uid[:2] / self.doc.uid
```
kleinhenz: maybe add the `.hdf5` extension to the path so that the data store is more easily readable to external programs

databroker/experimental/server_ext.py
```diff
@@ -0,0 +1,501 @@
...
+    if not isinstance(entry, WritingArrayAdapter):
+        raise HTTPException(
+            status_code=404, detail="This path cannot accept this array."
+        )
+    data = await request.body()
+    entry.put_data(data)
+    return json_or_msgpack(request, None)
+
+
+@router.put("/dataframe/full/{path:path}")
```
kleinhenz: should this be changed to `/node/full` to match the corresponding `GET` method? This change would also have to be made in the tiled client.

kleinhenz, line 91: should this be changed to `/node/full` to match the corresponding `GET` method? This change would also have to be made in the tiled client.

databroker/experimental/server_ext.py
```diff
@@ -0,0 +1,501 @@
...
+
+class PostMetadataRequest(pydantic.BaseModel):
+    structure_family: StructureFamily
+    structure: Union[ArrayStructure, DataFrameStructure]
+    metadata: Dict
+    specs: List[str]
+
+
+class PostMetadataResponse(pydantic.BaseModel):
+    uid: str
```
kleinhenz: maybe `loc` would be a better name since in the generic case this could be a path rather than a flat uid.

kleinhenz, line 47: maybe `loc` would be a better name since in the generic case this could be a path rather than a flat uid.

databroker/experimental/server_ext.py
```diff
@@ -0,0 +1,501 @@
...
+                serialize_arrow(validated_document.structure.micro.meta, {})
+            )
+
+        self.collection.insert_one(validated_document.dict())
+        return uid
+
+    def authenticated_as(self, identity):
+        if self.principal is not None:
+            raise RuntimeError(f"Already authenticated as {self.principal}")
+        if self.access_policy is not None:
```
kleinhenz, line 300: this probably doesn't need to be a blocker but it is probably worth thinking about how access policies should be integrated with the writing methods

databroker/experimental/server_ext.py
```diff
@@ -0,0 +1,501 @@
...
+            sorting=sorting,
+            access_policy=self.access_policy,
+            principal=principal,
+            **kwargs,
+        )
+
+    def post_metadata(self, metadata, structure_family, structure, specs):
+
+        mime_structure_association = {
+            StructureFamily.array: "application/octet-stream",
```
kleinhenz: I'm pretty sure this should be "application/x-hdf5" since it refers to the storage type and you are writing your arrays as hdf5

databroker/experimental/apischema_schemas.py
```diff
@@ -0,0 +1,117 @@
...
+}
+
+# @dataclass
+# class Box(Generic[T]):
+# content: T
+
+
+@dataclass
+class Document(Generic[StrucT]):
+    id: str
```
kleinhenz: I think this will still leave the mongo documents with an `_id` as the primary index. Dealing with this in pydantic is a bit of a pain since it disallows fields with leading underscores. Not sure what the best solution is.

databroker/experimental/server_ext.py
```diff
@@ -0,0 +1,501 @@
...
+                )
+            elif doc["structure_family"] == StructureFamily.dataframe:
+                yield (
+                    doc["uid"],
+                    WritingDataFrameAdapter(self.database, self.directory, doc),
+                )
+            else:
+                raise ValueError("Unsupported Structure Family value in the databse")
+
+    def _item_by_index(self, index, direction):
```
kleinhenz: I think this can be implemented in terms of `_items_slice`

databroker/experimental/server_ext.py
```diff
@@ -0,0 +1,501 @@
...
+router = APIRouter()
+
+
+@router.post("/node/metadata/{path:path}", response_model=PostMetadataResponse)
+def post_metadata(
+    request: Request,
+    body: PostMetadataRequest,
+    entry=Security(entry, scopes=["write:data", "write:metadata"]),
+):
+    if not isinstance(entry, MongoAdapter):
```
kleinhenz: this is too restrictive for the generic case. I think the pythonic thing to do would be to just call `entry.post_metadata` and catch the error if the method does not exist.

kleinhenz, line 59: this is too restrictive for the generic case. I think the pythonic thing to do would be to just call `entry.post_metadata` and catch the error if the method does not exist.

databroker/experimental/server_ext.py
```diff
@@ -0,0 +1,501 @@
...
+    )
+    return json_or_msgpack(request, {"uid": uid})
+
+
+@router.put("/array/full/{path:path}")
+async def put_array_full(
+    request: Request,
+    entry=Security(entry, scopes=["write:data", "write:metadata"]),
+):
+    if not isinstance(entry, WritingArrayAdapter):
```
kleinhenz: same comment as above. this is too restrictive for the generic case. I think the pythonic thing to do would be to just call `entry.post_metadata` and catch the error if the method does not exist.

kleinhenz, line 82: same comment as above. this is too restrictive for the generic case. I think the pythonic thing to do would be to just call `entry.post_metadata` and catch the error if the method does not exist.

databroker/experimental/server_ext.py
```diff
@@ -0,0 +1,501 @@
...
+    entry.put_data(data)
+    return json_or_msgpack(request, None)
+
+
+@router.put("/dataframe/full/{path:path}")
+async def put_dataframe_full(
+    request: Request,
+    entry=Security(entry, scopes=["write:data", "write:metadata"]),
+):
+    if not isinstance(entry, WritingDataFrameAdapter):
```
kleinhenz: same comment as above. this is too restrictive for the generic case. I think the pythonic thing to do would be to just call `entry.post_metadata` and catch the error if the method does not exist.

kleinhenz, line 96: same comment as above. this is too restrictive for the generic case. I think the pythonic thing to do would be to just call `entry.post_metadata` and catch the error if the method does not exist.

databroker/experimental/server_ext.py
```diff
@@ -0,0 +1,501 @@
...
+        # Organize files into subdirectories with the first two
+        # charcters of the uid to avoid one giant directory.
+        path = self.directory / self.doc.uid[:2] / self.doc.uid
+        path.parent.mkdir(parents=True, exist_ok=True)
+        array = numpy.frombuffer(
+            body, dtype=self.doc.structure.micro.to_numpy_dtype()
+        ).reshape(self.doc.structure.macro.shape)
+        with h5py.File(path, "w") as file:
+            file.create_dataset("data", data=array)
+        self.collection.update_one(
```
kleinhenz: I think it's probably worth doing a check here that the update succeeded (one document matched, one document updated).

kleinhenz: I think it's probably worth doing a check here that the update succeeded (one document matched, one document updated).

databroker/experimental/server_ext.py
```diff
@@ -0,0 +1,501 @@
...
+
+from .schemas import Document
+from .queries import RawMongo
+
+from sys import platform
+
+
+class PostMetadataRequest(pydantic.BaseModel):
+    structure_family: StructureFamily
+    structure: Union[ArrayStructure, DataFrameStructure]
```
kleinhenz: my suggestion for fixing [this](https://github.com/bluesky/databroker/pull/696#issuecomment-1124181963) is to replace this type annotation with `Structure` from `tiled.server.schemas` which treats `micro` and `macro` as opaque dictionaries. This is the approach used in e.g. the `/node/metadata` route.

databroker/experimental/apischema_schemas.py
danielballan: In bluesky/databroker, we have done well by having an application that means nothing to MongoDB in addition to the ObjectId `_id` that is under Mongo's control. The application id survives translation into other storage formats (like archival in JSON or msgpack) and avoids the ObjectId-specific semantics.

It does make sense to _index_ the `id` column, though.

databroker/experimental/example_array.py
danielballan: Agreed. I think this effectively has been translated into a unit test (elsewhere in this PR) and a notebook (which should go into public-demos) so this copy can be deleted.

databroker/experimental/example_dataframe.py
danielballan: Ditto above

databroker/experimental/server_ext.py
danielballan: The opaque dictionaries are a bit of cop-out / short-term corner-cutting on my part. The OpenAPI schema will be a lot more useful if we fill these out.

The specific issue here (in the linked traceback) is that OpenAPI has no idea what to do with a `pandas.DataFrame`. Seems likely we should change that to use Arrow-encoded bytes, as discussed on Slack.

databroker/experimental/server_ext.py
danielballan: I like that!

danielballan: Or `key`?

databroker/experimental/server_ext.py
danielballan: Yes. In my "recon" example from which this is derived, I forgot that `/dataframe/full` changed to `/node/full`.

databroker/experimental/server_ext.py
danielballan: Yeah, this occurred to me recently. I'm not sure what I was thinking here.

databroker/experimental/server_ext.py
danielballan: We subsequently discussed this at length on Slack, but just for the record I'll put the conclusion here: I think this should get a generic name like `nodes`. I don't think there is any reason to make it configurable. It is internal to the adapter; it would be like a SQL application making the names of the tables it uses configurable. Not necessary.

databroker/experimental/server_ext.py
danielballan: The half-implementation here is a little confusing. It might be best to comment this out or remove it until there is a way to actually use `data_blob`.

Yes, the base64-encoding doesn't belong here. The only time we should be base64-encoding Arrow is when we are forced to, when it is being transmitted in JSON. MongoDB supports binary blobs, and when we implement `data_blob` we should use that.

databroker/experimental/server_ext.py
danielballan: They have very different internals. The `ArrayAdapter` hold an array in memory as its state.

databroker/experimental/server_ext.py
danielballan: Same as with ArrayAdapter.

databroker/experimental/server_ext.py
danielballan: Yep, let's handle this in a follow-up PR.

databroker/experimental/server_ext.py
kleinhenz: is there a reason to require this to be defined separately for the IndexersMixin interface in tiled?

databroker/experimental/server_ext.py
jmaruland: I updated this line and now, it uses a collections name as `nodes`

databroker/experimental/server_ext.py
jmaruland: Fixed it

databroker/experimental/apischema_schemas.py
jmaruland: Good catch. That was the first draft of the Document when we were trying to circumvent the infinite recursive loop that was happening with pydantic. The process is fixed now and this files is irrelevant. Deleted now

databroker/experimental/example_array.py
jmaruland: Deleted

databroker/experimental/example_dataframe.py
jmaruland: Deleted

databroker/experimental/queries.py
jmaruland: The `RawMongo` class is used and registered separately in `server_ext.py` because databroker uses another `RawMongo` class somewhere else. the method `scan_id()` was moved to `test_experimental.py`

databroker/experimental/schemas.py
jmaruland: Deleted

databroker/experimental/schemas.py
jmaruland: Deleted

databroker/experimental/server_ext.py
jmaruland: Fixed. But if we are using this approach, I will also add the file extension for the parquet files that are generated to save the dataframes on disk. I will add this to the dataframe writer, as well

databroker/experimental/server_ext.py
jmaruland: Yes, I was following the idea of the recon example. I applied this changes and writer is using `node/full` for both array and dataframe now

databroker/experimental/server_ext.py
jmaruland: Fixed

databroker/experimental/server_ext.py
jmaruland: I like the idea of having a method with this name. The implementation of this would make things easier for someone using this class. I think that calling `_item_slice` from within `_item_by_index` is a good approach.

databroker/experimental/server_ext.py
jmaruland: I can see a few way of doing this:

- First, the method update_one() returns an UpdateResult object where we can get this confirmation from by calling UpdateResult.acknowledged (https://pymongo.readthedocs.io/en/stable/api/pymongo/collection.html#pymongo.collection.Collection.update_one).
- The other one would be to wrap this command with a try/except statement and let the Collection class handle the rest

I don't think we should run a validation between the original document and the document updated in mongodb; at least not at this level

danielballan: I think UpdateResult also tells you how many documents were “matched” (that is, satisfied the query) and how many were updated. See Joe’s comment in that light.

databroker/experimental/server_ext.py
danielballan: I think this needs to return a single item, not a list or iterable of items. That is, we want `thing`, not `[thing]`. I think this currently returns `[thing]`.

jmaruland: Fixed it. I see how the list was being generated for this case

databroker/experimental/server_ext.py
danielballan: In https://github.com/bluesky/tiled/pull/213 the client has been changed to use the term `key`. The `PostMetadataResponse` must be updated. The `Document` may (and, as you have argued, should) be updated for consistency.

danielballan:
Next step is to support the new queries added in tiled v0.1.0a63.

1. Add `apply_mongo_query` method to `MongoAdapter`, copied from `databroker/mongo_normalized.py`.
2. At the bottom of `server_ext.py`:

```py
from tiled.queries import Eq


def eq(query, catalog):
    return catalog.apply_mongo_query({f"metadata.{query.key}": query.value})


MongoAdapter.register_query(Eq, eq)
```

Do this for `Eq`, `Comparison`, `Contains`, `FullText`, etc. Only do this for the queries in `tiled/quieries.py`. Do not worry about the qureies in `databroker/queries.py` because they are Bluesky-specific.

danielballan (approved):

Issue #699, opened 2022-05-18
Handle case where there is a 'time' data key.


danielballan:
In databroker, we make a best effort to fit the streaming "document model" into standard PyData types.

In v1, we use DataFrames (`.table(...)`) or generators (`.data(...)`). In v2, we use xarray Datasets. They have different trade-offs. DataFrames do not handle image (or in general N-dimensional) data well. Datasets do.

Both provide a concept of special columns that are used for alignment. DataFrames call it "index"; Datasets call it "coordinates". They are related conceptually and in terms of their implementation (xarray uses pandas.Index internally). But they have some semantic differences. DataFrames' indexes live in a separate namespace from columns, so there is no worry about name collisions. Datasets put their columns ("variables") and their indexes ("coordinates") in one namespace, and so the same name cannot be used for both a variable and a coordinate.

For databroker, this means that if the user uses `"time"` as a data key in an Event document, it will collide with the Event time like this:

```py
{
    "time": ...,  # all Events have an overall time
    "data": {
        "time": ...,  # ophyd-defined name for a data key might also be "time"
    },
    ...
}
```

And in v2 we currently do not handle that gracefully. @tacaswell has proposed that we leave the user-defined name alone and add underscopes to the front of our coordinate `time`, trying first `_time` and then `__time` and then `___time` (etc.) until we find a name that doesn't collide. Hopefully `_time` will usually be enough, but this retry-based approach is guaranteed to avoid a collision even if some user decides to use both `time` and `_time` for example.

danielballan:
[Copied from Slack] This is the place to start: https://github.com/bluesky/databroker/blob/a2a5d6600ecf75efdd1c70cb67a0ebe2a1584bbd/databroker/mongo_normalized.py#L87-L113

gwbischof:
In databroker-intake the event.data.time overrides event.time

PR #700, opened 2022-05-18
Time data key conflict

gwbischof:


danielballan:
This is different from the proposal in #699 in two ways. It changes the default name instead of trying the normal key “time” and only resorting to a weird key with underscores in the event of a conflict. And it has no retry pathway: it will fail if anyone every makes an ophyd device with the new key in it.

gwbischof:
We plan to fix this by reserving the `time` and `seq_num` keys.  See https://github.com/bluesky/event-model/pull/223
So this PR is no longer needed

PR #705, opened 2022-05-27
Deprecate RawMongo

### LINKED ISSUE #701: Drop RawMongo query type?
There are multiple signals that we need to treat `RawMongo` as a rare escape hatch and maybe even drop support for it entirely.

1. **It exposes the database schema.** One goal of Databroker (and Tiled) is to de-couple data storage from data APIs such that data providers can evolve and improve data storage without breaking user code. As written, `RawMongo` is tightly coupled to the "mongo normalized" database layout. If we add other layouts, such as the ongoing work in #696, or evolve the existing one, such as the work (on hiatus) on "mongo embedded", the differences between the layouts will leak into user code. This is already prompting confusing design questions in #696.
2. **It exposes the database technology.** I think scaled deployments of Databroker will be built on MongoDB for at least the next several years, so I am not urgently concerned by this point. But, MongoDB's open source commitment has deteriorated, and it would be useful to at least have a _path_ to Databroker backed by PostreSQL or another technology—perhaps one not yet invented or prominent. If we have a lot of user code relying on the full MongoDB query API, moving will be a challenge.
3. **It enables arbitrary filter execution.** I am not completely sure this is _wrong_ but it is at least _exotic_. It is likely down to the project's roots in replicating the single-user workflows that a couple post-docs (hello!) used in graduate school to manage their own data. It has issues in a multi-tenant context. There are, at least, some performance concerns, and I cannot rule out security concerns.

I think the costs of this change will be fairly low if we execute it soon. Few (if any) users have taken up the power of complex MongoDB queries, in any version of Databroker. Most are doing key-value matching. This could easily be addressed by a more targeted `KeyValue` query type. As other use cases come to light, they can be addressed with other targeted queries. They would restore a clear separation between client and server, enabling us to really separate storage from access and also provide stronger constraints on load from queries. As a side effect, I think the usability for most users would also be improved.

The change would not be cost-free. There is some potential benefit to enable power-users to access the full expressive capabilities of MongoDB queries. But I think we should view that as, at most, a temporary bridge to let us understand what queries they need so that we can "pave the cow paths" and provide purpose-built queries that are more constrained. I am advocating to at _least_ demote `RawMongo` into an experimental status and perhaps put it on the path to removal in the near future.

danielballan:
Closes #701 

- Updates to changes in tiled slicing
- Updates to changes in tiled query API

### DIFF
--- a/databroker/client.py
+++ b/databroker/client.py
@@ -279,7 +279,7 @@ class CatalogOfBlueskyRuns(CatalogOfBlueskyRunsMixin, Node):
             raise KeyError(f"No match for partial_uid {partial_uid}")
         else:
             # By construction there must be only one result. Return it.
-            return results.values.first()
+            return results.values().first()
 
     def get_serializer(self):
         from tiled.server.app import get_root_tree
--- a/databroker/client.py
+++ b/databroker/client.py
@@ -218,7 +218,7 @@ class CatalogOfBlueskyRuns(CatalogOfBlueskyRunsMixin, Node):
     >>> catalog.uid["9acjef"]  # (partial) uid lookup
     >>> catalog[1234]  # automatically do scan_id lookup for positive integer
     >>> catalog["9acjef"]  # automatically do (partial) uid lookup for string
-    >>> catalog[-5]  # automatically do catalog.values_indexer[-N] for negative integer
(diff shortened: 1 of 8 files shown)


tacaswell (approved):
Mostly a conceptual review.

PR #707, opened 2022-06-14
Send xarrays with tiled

jmaruland:
Edit from @danielballan: This PR prompted a re-think in how we handle xarrays in Tiled in general. We'll leave it open for now as there may be something useful to mine from it later, but it is not actively being developed.

<hr />

Fixing some bugs in the schema for `DataArrays` (in progress).
The structure for the macros that is being generated by `DataArrayAdapter` is not consistent with the current schema for `DataArrayMacroStructure` in server/pydantic_xarray.py.

### DIFF
--- a/databroker/experimental/schemas.py
+++ b/databroker/experimental/schemas.py
@@ -24,7 +24,9 @@ structure_association = {
 class Document(pydantic.BaseModel):
     key: str
     structure_family: StructureFamily
-    structure: Union[ArrayStructure, DataFrameStructure, DataArrayStructure]
+    structure: Union[
+        ArrayStructure, DataFrameStructure, DataArrayStructure, DatasetStructure
+    ]
     metadata: Dict
     specs: List[str]
     mimetype: str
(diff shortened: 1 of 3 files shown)


danielballan:
Addressing these errors in the test example:

```
ValidationError: 3 validation errors for DataArrayMacroStructure
coords -> x
  str type expected (type=type_error.str)
coords -> y
  str type expected (type=type_error.str)
name
  none is not an allowed value (type=type_error.none.not_allowed)
```

Two unrelated things are going wrong.

## Imprecise definitions in the original dataclass

Dataclasses are annotated, but they don't actually verify/validate that the annotations are correct. (An external library like apischema may be used to do the validation. Unlike Pydantic, Dataclasses doesn't build that in.) This means that there could be errors in the dataclass-based definitions (my fault!) that haven't been caught up to this point. You have hit such an error:

```py
name: str
```

should be

```py
name: Optional[str]
```
as name is allowed to be `None` for xarrays:

```py
In [5]: xarray.DataArray([1,2]).name is None
Out[5]: True
```

Both the original dataclass definition and the pydantic definition (both in tiled) should be fixed.

## Dropped self-reference

DataArrayMacroStructure is self-referential. It contains `coords`, a dictionary that maps `str` to another `DataArrayMacroStructure`.

How do you refer to something before you have defined it?

One way is what I chose in `tiled.structures.xarray`. I define it one way temporarily

https://github.com/bluesky/tiled/blob/d5aad00a092c9bef4d50b0137b0b83eadd13f356/tiled/structures/xarray.py#L24-L26

and then I override the definition below that, once I have something to refer to.

https://github.com/bluesky/tiled/blob/d5aad00a092c9bef4d50b0137b0b83eadd13f356/tiled/structures/xarray.py#L60-L65

This is actually _not_ a good way to do it; it's just what I stumbled on being somewhat inexperienced with type annotations in Python.

I have since learn there are better ways. Here are two: https://stackoverflow.com/a/69802572/1221924. The second one is preferred but only works on Python 3.7+. Since we only support Python 3.7+ anyway, this is fine for us.

In the pydantic xarray definition in https://github.com/bluesky/tiled/pull/237, I think the second part of my (bad) approach got dropped entirely, and so `coords` was left as `Dict[str, str]` instead of `Dict[str, DataArrayStructure]`. That explains the errors.

danielballan:
This was superseded by work upstream in Tiled core.

Issue #712, opened 2022-06-30
Add support for DELETE 


danielballan:
Similar to https://github.com/AI-multimodal/aimmdb/pull/13/files#diff-2573ab5805e99bbae8e14de22a207ad6da2ae65126d0280bc0d3b3aaa48bc163R101

danielballan:
The `ArrayAdapter` and `DataFrameAdapter` should support a `delete()` method that deletes the document from MongoDB (see Joe's example) and removes the file on disk.

danielballan:
Closed by #715

PR #715, opened 2022-07-06
Added delete() method for arrays and dataframes

jmaruland:


jmaruland:
Closes https://github.com/bluesky/databroker/issues/712

databroker/experimental/server_ext.py
```diff
@@ -121,6 +121,17 @@ def put_data(self, body):
...
             raise ValueError("Error while writing to database")
 
+    def delete(self):
+        if self.doc.data_url is not None:
+            path = self.doc.data_url.path
+            if platform == "win32" and path[0] == "/":
+                path = path[1:]
+
+            self.file.close()
+            os.remove(path)
```
danielballan: I think `path.unlink()` would be a safer (future-proof) choice here.

databroker/experimental/server_ext.py
```diff
@@ -194,6 +205,16 @@ def put_data(self, body):
         if result.matched_count != result.modified_count:
             raise ValueError("Error while writing to database")
 
+    def delete(self):
+        if self.doc.data_url is not None:
+            path = self.doc.data_url.path
+            if platform == "win32" and path[0] == "/":
+                path = path[1:]
+
+            os.remove(path)
```
danielballan: Same as above.

databroker/tests/test_experimental.py
```diff
@@ -123,3 +123,40 @@ def test_queries(tmpdir):
...
+    )
+
+    del client[key]
+
+    # For arrays
+    test_array = numpy.ones((5, 5))
+
+    key = client.write_array(test_array, {"scan_id": 1, "method": "A"}, ["BlueskyNode"])
+
+    del client[key]
```
danielballan: We should check that the item is really gone.

```py
assert key not in client
```

It would also be good to confirm that the file has been deleted, if we can locate it.

danielballan (changes requested):

databroker/experimental/server_ext.py
jmaruland: Fixed in https://github.com/bluesky/databroker/pull/715/files#diff-d4869bded9d2c76766f74b02b105f0f5e2825963da33e28a1a566e5a8af6f6b0R131

databroker/experimental/server_ext.py
jmaruland: Also fixed
https://github.com/bluesky/databroker/pull/715/files#diff-d4869bded9d2c76766f74b02b105f0f5e2825963da33e28a1a566e5a8af6f6b0R214

databroker/tests/test_experimental.py
jmaruland: Added the assert lines for the array and dataframe cases
https://github.com/bluesky/databroker/pull/715/files#diff-d4869bded9d2c76766f74b02b105f0f5e2825963da33e28a1a566e5a8af6f6b0R214

PR #716, opened 2022-07-08
Added update_metadata() to write_ adapters

jmaruland:
This PR started as an approach to add more information to the document schema that will help to keep track of all the updates made to the metadata and specs of the samples.
In addition, we implemented a revisions system with the mongo database that will keep track of old versions of documents. Every time that `update_metadata()` is run, the active document that is saved in collections is copied to revisions where every entry is protected by the same `key` id plus a `revision` number. There two parameters is used as a unique identifier for every entry.


databroker/experimental/server_ext.py
```diff
@@ -305,6 +306,7 @@ def post_metadata(self, metadata, structure_family, structure, specs):
         }
 
         key = str(uuid.uuid4())
+        created_date = datetime.today()
```
danielballan: Typically the full datetime is recorded, not just the date. Any particular reason for choosing date?

danielballan: Prompted by yesterday's conversation I think we should expand the scope of this PR to include:

* A "Revisions" collection
* Tracking creation time and update time --- perhaps, as you suggested, only creation time on Document and only update time on DocumentRevision
* Metadata updates

I think each aspect may inform the details of how we do the others, so it is sensible to do them as a single PR.

jmaruland: > Typically the full datetime is recorded, not just the date. Any particular reason for choosing date?

I thought it would be a good idea to record both date and time. This would give us a more exact information about the creation or update of a particular entry. Keep in mind that I'm importing `datetime` from the `datetime` package and not `date` which can only generate a date with no time

danielballan: I see. I have never seen `today()` used. I wrongly assumed it gave the date, like `datetime.now().date()`. I see now that it gives a datetime, similar to `datetime.now()`. According to the documentation [`now()` is preferred](https://docs.python.org/3/library/datetime.html#datetime.datetime.now). It does not state why it is preferred; perhaps it's just that the name is clearer.

Also, we need to consider timezones. For this kind of application, I think using UTC time is best. See that same documentation link above for useful information on how to get that.

databroker/experimental/schemas.py
```diff
@@ -74,3 +65,90 @@ def is_mime_type(cls, v):
...
+        if values.get("data_blob") is not None and values.get("data_url") is not None:
+            raise ValueError(
+                "Not Valid: data_blob and data_url contain values. Use just one"
+            )
+        return values
+
+
+class DocumentRevision(BaseDocument):
+    revision: int
+
```
danielballan, line 89: It may be useful to add a classmethod constructor here to do what you were trying to do in `__init__`.

```py
@classmethod
def from_document(cls, document)
    return cls(key=document.key, ...)
```

databroker/experimental/schemas.py
```diff
@@ -18,15 +19,15 @@
...
     key: str
     structure_family: StructureFamily
     structure: Union[ArrayStructure, DataFrameStructure]
     metadata: Dict
     specs: List[str]
     mimetype: str
-    data_blob: Optional[bytes]
-    data_url: Optional[pydantic.AnyUrl]
+    created_at: datetime
+    updated_at: datetime
```
danielballan: Do you want to move this to just `Document`? You argued that we should only write it down in one place, and I'm convinced.

databroker/experimental/server_ext.py
```diff
@@ -62,8 +63,9 @@ def inner(self, *args, **kwargs):
 class WritingArrayAdapter:
     structure_family = "array"
 
-    def __init__(self, collection, directory, doc):
+    def __init__(self, collection, revisions, directory, doc):
```
danielballan: I think it would make sense to pass in the `database` here rather than separately passing in each of its collections.

databroker/experimental/server_ext.py
```diff
@@ -125,6 +127,54 @@ def put_data(self, body):
         if result.matched_count != result.modified_count:
             raise ValueError("Error while writing to database")
 
+    def put_metadata(self, metadata, specs):
+
+        revisions_doc = (
+            self.revisions.find({"key": self.doc.key}).sort({"revision": -1}).limit(1)
+        )
+        if revisions_doc is not None:
```
danielballan: We should look at ways to do an atomic increment in MongoDB. There is a race condition here if two clients do updates concurrently. We need to ensure the `revision` is unique.

databroker/experimental/server_ext.py
```diff
@@ -125,6 +127,54 @@ def put_data(self, body):
...
+            updated_at=self.doc.updated_at,
+            revision=revision,
+        )
+
+        self.revisions.insert_one(validated_revision.dict())
+
+        updated_at = datetime.now(tz=timezone.utc)
+        self.doc.updated_at = updated_at
+
+        if len(metadata) > 0:
```
danielballan: If I want to update `metadata` to be empty `{}` or `specs` to be empty `[]`, shouldn't that update be processed?

databroker/experimental/server_ext.py
```diff
@@ -125,6 +127,54 @@ def put_data(self, body):
...
+                "$set": {
+                    "metadata": self.doc.metadata,
+                    "specs": self.doc.specs,
+                    "updated_at": self.doc.updated_at,
+                }
+            },
+        )
+
+        if result.matched_count != result.modified_count:
+            raise ValueError("Error while writing to database")
```
danielballan: I suggest classifying this as a `RuntimeError`.

databroker/experimental/server_ext.py
jmaruland: `datetime.today()` was changed to `datetime.now()`. Also, use of UTC was added

databroker/experimental/schemas.py
jmaruland: For this particular case, I think that we could move `created_at` to `Document` and leave `updated_at` in `BaseDocument`. This way, we can keep track of when a document version was created before we move it to revisions

databroker/experimental/server_ext.py
danielballan: Note: we're rely on an index. There is a chance of a race, but the result will be failure (and client can try again...) rather than duplication.

danielballan:
Now that we are adding indexes, I think we should also add an index to the `nodes` collection. This will make lookup by `key` faster. We might as well also enforce it to be unique. Using UUID4 _should_ achieve that result anyway, but it doesn't hurt to claim uniqueness via an index as well.

databroker/experimental/schemas.py
danielballan: Oops, yes, I agree. I meant to aim this comment at `created_at` but I hit the wrong line. The current version and the past revisions should each have `updated_at` but only the current version needs `created_at`.

databroker/experimental/server_ext.py
jmaruland: Fixed in the latest commit

databroker/experimental/server_ext.py
jmaruland: Fixed in the latest commit

databroker/experimental/server_ext.py
```diff
@@ -41,22 +43,63 @@
...
 
+class Revisions:
+    def __init__(self, collection, key):
+        self._collection = collection
+        self._key = key
+
+    def __len__(self):
+        return self._collection.count_documents(
+            {"key": self._key}
+        )  # maybe wrong MongoDB usage here...
```
danielballan: Delete comment (assuming this usage is now correct).

databroker/experimental/server_ext.py
```diff
@@ -41,22 +43,63 @@
...
+
+    def __getitem__(self, item_):
```
danielballan, line 61: This looks mixed up and likely needs testing.

The usage `r[i:j]` should lead to `skip(offset).limit(j - i)`. The usage `r[i:]` or `r[i:None]` or `r[:-1]` (all equivalent) should lead to `skip(offset)` with no limit. Pymongo also accept `skip(offset).limit(0)` where `0` means "no limit", which is an option if you find it leads to cleaner code.

databroker/experimental/server_ext.py
```diff
@@ -68,7 +111,18 @@ def new(cls, collection, doc):
...
-        return cls(collection, doc)
+        return cls(database, doc.key)
+
+    @property
+    def doc(self):
+        now = time.monotonic()
+        if now > self.deadline:
+            self._doc = Document(
+                **self.collection.find_one({"key": self.key})
+            )  # run query
```
danielballan: This comment seems superfluous. :-)

databroker/experimental/server_ext.py
```diff
@@ -369,15 +546,22 @@ def post_metadata(self, metadata, structure_family, structure, specs):
...
         _adapter_class_by_family[structure_family].new(
-            self.collection, validated_document
+            self.database, validated_document
         )
         return key
 
+    def create_indexes(self):
+        self.revision_coll.create_index(
+            [("key", pymongo.ASCENDING), ("revision", pymongo.DESCENDING)], unique=True
+        )
```
danielballan, line 563: While we're creating indexes, we should also create an index on the `nodes` collection to ensure that `key` is unique.

danielballan (approved):

PR #719, opened 2022-07-19
Add ScanIDRange

### LINKED ISSUE #711: Request for ScanID query that uses start_id and end_id
This requested query would replace a MongoDB query I was using.

## Expected Behavior
Create a ScanIDRange query to specify the start_id and end_id.  This would be similar to running a MongoDB using scan_id >= start_id and scan_id <= end_id.  Or specifically the following in JSON:
{"scan_id": {
  "$gte": start_id,
  "$lte": end_id
}}
I'd prefer the start_id to default to 0 and the end_id to default to returning everything after the start_id.

## Current Behavior
ScanID must have every scan_id specified explicitly.

## Possible Solution
Create new query object that support start_id and end_id.

## Context
This supports creating a GUI for accessing run data through Tiled/Databroker.

## Your Environment
Using Tiled to remotely access blueSkyEvents.

gwbischof:
Fixes #711

### DIFF
--- a/databroker/queries.py
+++ b/databroker/queries.py
@@ -98,6 +98,47 @@ def ScanID(*scan_ids, duplicates="latest"):
     return _ScanID(scan_ids=scan_ids, duplicates=duplicates)
 
 
+@register(name="scan_id_range")
+@dataclass`
+class _ScanIDRange:
+    """
+    Find scans in the range.
+    """
+
+    start_id: int
+    end_id: int
+    duplicates: Duplicates
+
+    def __init__(self, *, start_id, end_id, duplicates):
+        self.start_id = start_id
+        self.end_id = end_id
+        self.duplicates = Duplicates(duplicates)
+
+    def encode(self):
+        return {
+            "start_id": self.start_id,
+            "end_if": self.end_id,
+            "duplicates": self.duplicates.value,
+        }
+
+    @classmethod
+    def decode(cls, *, start_id, end_id, duplicates):
+        return cls(
+            start_id=int(start_id),
+            end_id=int(end_id),
+            duplicates=Duplicates(duplicates),
+        )
+
+
+def ScanIDRange(start_id, end_id, duplicates="latest"):
+    # Wrap _ScanIDRange to provide a nice usage for *one or more scan_ids*:
+    # >>> ScanIDRange(5, 100)
+    # Placing a varargs parameter (*scan_ids) in the dataclass constructor
+    # would cause trouble on the server side and generally feels "wrong"
+    # so we have this wrapper function instead.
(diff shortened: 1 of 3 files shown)


databroker/queries.py
```diff
@@ -98,6 +98,47 @@ def ScanID(*scan_ids, duplicates="latest"):
...
+    def decode(cls, *, start_id, end_id, duplicates):
+        return cls(
+            start_id=int(start_id),
+            end_id=int(end_id),
+            duplicates=Duplicates(duplicates),
+        )
+
+
+def ScanIDRange(start_id, end_id, duplicates="latest"):
+    # Wrap _ScanIDRange to provide a nice usage for *one or more scan_ids*:
```
danielballan: `ScanID` accepts variadic arguments, i.e. `ScanID(*scan_ids)`, so that we can one `ScanID(5)` or several `ScanID(5, 13, 21)` without doing this: `ScanID([5])`.

Variadic arguments don't play well with schemas, so I made

```py
_ScanID(scan_ids)  # must be a list
```

wrapped by

```py
ScanID(*scan_ids)  # variadic
```

That pattern is not necessary for `ScanIDRange` because there is no use case for variadic arguments here. You can just make one class, `ScanIDRange`, without the wrapping.

databroker/queries.py
```diff
@@ -341,6 +382,41 @@ def scan_id(query, catalog):
     return results
 
 
+def scan_id_range(query, catalog):
+    mongo_results = catalog.apply_mongo_query({"scan_id": {"$gt": query.start_id, "$lt": query.end_id}})
```
danielballan: I think users will expect an inclusive lower bound. `ScanIDRange(5, 10)` should include `5` for sure. What to do with the upper bound is less clear. We could follow the behavior of `range` user a half-open interval, or we could make it fully closed.

@michael-sutherland What would you expect here?

michael-sutherland: I'd expect the same behavior as range, half-open interval.

danielballan:
A lot of the requirements added are not directly dependencies of databroker but secondary dependencies through tiled. It would be easier to maintain going forward to just add `tiled[all]`.

The `ScanIDRange` query needs to be registered here too: https://github.com/bluesky/databroker/blob/bef2c2d97792d5ea45bac7156e28ea14f27df4d9/databroker/queries.py#L417

danielballan (approved):

danielballan:
Works for me locally on test data:

```py
In [5]: c
Out[5]: <Catalog {22521, 22524, 22525, 22526, 22528, 22529, 22534, ...} ~123 entries>

In [6]: c.search(ScanIDRange(22520, 22530))
Out[6]: <Catalog {22521, 22524, 22525, 22526, 22528, 22529}>
```

PR #722, opened 2022-08-04
[Experimental backend] Support chunked upload and sparse arrays

danielballan:
This PR changes only the alpha experimental backend.

This PR switches from HDF5 for Zarr for array storage because the latter has first-class support for concurrent writes of non-overlapping chunks. For DataFrame storage, it continues to rely on parquet, but it switches from a single parquet file to a directory of parquet file(s), supporting partitioned data.

It also adds support for Tiled's new `sparse` structure family, backed also be a directory of parquet files.


danielballan:
Self-merging this to support work on the experimental floor. This does not touch any production databroker files, only `databroker.experimental.`

PR #728, opened 2022-08-18
Header.data() fix

### LINKED ISSUE #727: 'Array' object has no attribute 'data'
This code:
```
db = databroker.from_profile("nsls2", username=None)['chx']['raw'].v1
header2 = db['e909f4a2-12e3-4521-a7a6-be2b728a826b']
header.data("eiger4m_single_image")
```
Produces this error:
```
File /nsls2/data/chx/shared/config/prefect_overlay/2022-2.2-py39-tiled/lib/python3.9/site-packages/databroker/v1.py:1047, in Header.data(self, field, stream_name, fill)
   1045 if not fill:
   1046     raise ValueError("Only fill=True is now supported by the data(...) method.")
-> 1047 for item in self._run[stream_name]["data"][field][:].data:
   1048     yield item

AttributeError: 'Array' object has no attribute 'data'
```

gwbischof:
Fixes #727

### DIFF
--- a/databroker/v1.py
+++ b/databroker/v1.py
@@ -1044,7 +1044,7 @@ class Header:
         """
         if not fill:
             raise ValueError("Only fill=True is now supported by the data(...) method.")
-        for item in self._run[stream_name]["data"][field][:].data:
+        for item in self._run[stream_name]["data"][field][:]:
             yield item
 
     def stream(self, *args, **kwargs):


danielballan:
Thanks.

I wonder why the tests didn't catch this crucial bug. Can you look into why the existing tests didn't catch it and extend it such that it would fail without the fix in this PR?

https://github.com/bluesky/databroker/blob/c41645019ac63fb2237beef9a69b628516c2fca5/databroker/tests/test_broker.py#L910-L922

danielballan:
I still think it's worth adding a unit test here but I'm going to merge to get this fix out.

PR #731, opened 2022-09-29
Added and implemented new references schema

jmaruland:
This PR implements a new approach in the use of references as an additional field in the document. It's use is similar to what we are doing with specs as a list of strings but references will consist in a list of `ReferenceDocument` object that can store a label and a url. The `references` object does not sit inside `metadata`; it seems more appropriate to leave it at the same level of `metadata` and `specs`


danielballan:
Failures on the 3.9 build were all from the very-deprecated databroker v0 sqlite backend, `Test_AD_hdf5_files`.

PR #732, opened 2022-09-30
Add CLI with shape-fixer.

danielballan:
## Description

This adds a CLI to databroker. In this PR it only has one command, but it is expected to grow more.

## Motivation and Context

This captures [the "shape-fixer"](https://github.com/danielballan/migration/blob/main/shape_fixer.py) which we use to retroactively edit the `shape` in the EventDescriptor to match the actual shape. This is intended to make it easily usable by people other than me. :-)

## How Has This Been Tested?

The shape-fixer has been used on about ten beamlines. This exact PR branch was run at ISS today.

### DIFF
--- a/databroker/cli.py
+++ b/databroker/cli.py
@@ -85,5 +85,8 @@ def shape_fix(
             progress.update(task, advance=1)
 
 
+main = cli_app
+
+
 if __name__ == "__main__":
-    cli_app()
+    main()
--- a/databroker/cli.py
+++ b/databroker/cli.py
@@ -54,14 +54,14 @@ def shape_fix(
 
     items = adapter.items()
     if limit:
-        typer.echo(f"Limited to first {limit} only")
+        typer.echo(f"Limited to first {limit} BlueskyRuns only")
         items = items[:limit]
 
     with Progress() as progress:
         task = progress.add_task("Migrating...", total=len(items))
         for uid, run in items:
             try:
-                for stream in run.values():
+                for stream_name, stream in run.items():
                     descriptor = stream.metadata["descriptors"][0]
                     recorded_shapes, measured_shapes = measure(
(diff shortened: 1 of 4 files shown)


danielballan:
Example usage, which we should put into docs once we add documentation about the migration path:

```
PYTHONPATH=~/.config/tiled/ databroker admin shape-fixer  mongodb://... --asset-registry-uri mongodb://... --query "TimeRange(since='2022-08')" --limit 10 --dry-run --strict --handler 'APB = iss_patches:APBBinFileHandler' --handler 'PIZZABOX_ENC_FILE_TXT_PD = iss_patches:PizzaBoxEncHandlerTxtPD' --handler 'XSP3 = iss_patches:ISSXspress3HDF5Handler' --handler 'APB_TRIGGER = iss_patches:APBTriggerFileHandler' --handler 'PIL100k_HDF5 = iss_patches:ISSPilatusHDF5Handler' --handler 'AD_HDF5 = area_detector_handlers.handlers:AreaDetectorHDF5Handler'
```

danielballan:
Example output:

image

danielballan:
Before merging we should get this working via

```
databroker ...
```

not just

```
python -m databroker.cli ...
```

We rely on scripts like `iss_patches.py` being importable from the CWD. Annoyingly, it seems like that using the proper CLI (`databroker ...`) breaks that somehow. I haven't run into this before.

jwlodek (approved):

jklynch:
This would benefit from automated tests, if not to ensure the code works now then to ensure it doesn't break in the future. I imagine this is difficult because of the dependencies on external services. The solution is to make it easy to generate simulated runs of bluesky documents.

danielballan:
Agreed.

Issue #740, opened 2022-10-19
fix memory issue with mongo_normalized when called by Tiled


JunAishima:
Memory usage never decreases when Tiled is used to generating datasets created from documents on beamlines configured to use databroker.mongo_normalized. On a test example with 47 datasets from rsoxs, memory usage increased by 10GB each time the test datasets were retrieved by a client, and never decreased until the Tiled server was shut down.

We would expect that at some point the memory usage by Tiled should stabilize.

This looks to be an issue with aggressive caching (lru_cache) in `mongo_normalized.DatasetFromDocuments._inner_get_columns()`. @danielballan has also suggested removing the decorator from `mongo_normalized.DatasetFromDocuments._get_time_coord()` as well.

## Steps to Reproduce (for bugs)
1. NSLS-II tiled config for rsoxs alone created, `tiled serve config config.yml` for that particular configuration.
2. Client script developed to retrieve datasets and read() them. A set of 47 datasets is available upon request
3. Run the client script repeatedly using the uuids, monitor with `top`. traces with `memray` also available with the current code.
Script fragment:
```
from tiled.client import from_uri
c=from_uri('http://127.0.0.1:8000/api')
dataset_names # list of uuids compiled from a text file 
for dataset in dataset_names:
    c['rsoxs']['raw'][dataset]['primary']['data']['Small Angle CCD Detector_image'].read()

```

This problem has caused Tiled instances to crash in production, which is obviously not ideal for beamlines that rely on this for their processing and analysis procedures.

Tested in the following environment: Conda environment on RHEL 8.6 containing tiled-server 0.1.0a74, databroker 2.0.0b10, with python 3.9.13.

PR #741, opened 2022-10-19
remove caching

JunAishima:
* caching was keeping hold of documents that are never cleared within normal usage within the production system, resulting in overrunning memory and crashed tiled instances

## Description
Remove @lru_cache decorators from functions in DatasetFromDocuments() that led to tiled crashes.

## Motivation and Context
#740 

## How Has This Been Tested?
Tested locally with the same server environment described in the issue. After changes, memory usage never went above 1.0% with the same dataset that was causing the server process to increase at 10GB per client run.

### DIFF
--- a/databroker/mongo_normalized.py
+++ b/databroker/mongo_normalized.py
@@ -724,7 +724,6 @@ class DatasetFromDocuments:
             array = array[slice]
         return array
 
-    @functools.lru_cache(maxsize=1024)
     def _get_time_coord(self, slice_params):
         if slice_params is None:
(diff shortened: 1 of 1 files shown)


danielballan:
Some notes:
* The caches in question were bounded on number of items rather than byte size, and that is an important lesson learned. The byte size could grow very large because the items could include "filled" events with image data.
* We could refine the caches, with lower bounds, bounds that account for byte size. But this kind of caching, designed to expedite repeated requests, is better handled generically closer to the outer edge of the application, in a response cache.

PR #747, opened 2022-12-05
suggested additions to documentation of API versions

dylanmcreynolds:
From a conversation on nakea slack, I would love to highlight the differences between the databroker API version and the databroker software version. Furthermore, the documentation on API versions does not quite capture the differences between the API v2 and the tiled-supported version of databroker, which I think is like v2 but with different initialization.

## Description
I added to the main table of API version the version of databroker software supported. I also tried to highlight the subtlety of API `v2` and `v2 with tiled`. If this approach to the documentation is deemed worthy, we should also an initialization example, for which this PR currently only adds a placeholder.


danielballan:
This is great. I think it should go in coordinated with other changes to document v2. So far, there has been zero documentation of the databroker 2.x pre-release, only the tiled documentation.

PR #748, opened 2022-12-08
Maint: Grant access to replay function from databroker.replay

maffettone:
Previously this function was not importable.

If this was hidden on purpose, please feel free to close this request. 

## Description

Made it importable and removed one layer of indirection in the `__init__.py`

## Motivation and Context

Seems like a lovely function to not be able to import.

## How Has This Been Tested?

`from databroker.replay import replay`


databroker/replay/__init__.py
```diff
@@ -0,0 +1 @@
+from .replay import replay
```
mrakitin:
```suggestion
from .replay import replay  # noqa F401
```

mrakitin (commented):
Looks good to me. Let's make the linter happy with my suggestion.

mrakitin (approved):

databroker/replay/__init__.py
```diff
@@ -0,0 +1 @@
+from .replay import replay  # noqa F401
```
tacaswell:
```suggestion

```

Doing this makes the underlying module inaccessible due to the shadowing.  Can we accept the extra layer or rename `replay.py` -> `__init__.py` ?

maffettone: Somewhat on purpose (based on my interpretation of the design). Nothing else in that module is meant to be imported (private and main functions), and the main function can still be accessed by running the file or adding an entry point if needed. 

I have no qualms about changing it though if recommended style.

mrakitin (approved):
Let's merge it! Any objections, @tacaswell, @maffettone?

danielballan (changes requested):
Making a module in accessible is a blocker from my POV. Even if it's internal, it still should be importable for things like testing.

Could rename `.replay` to `._replay` or something else.

maffettone:
Good point. I followed the initial suggestion to place the importable function into the `__init__`, then renamed the main script as a script to remove confusion.

danielballan (approved):

PR #750, opened 2022-12-09
Fix bug in config access

### LINKED ISSUE #745: Reading configuration of devices in a run doesn't work with tiled server
## Expected Behavior
With databroker backed by intake I am able to access the configuration parameters of devices like this:

```python
db = databroker.catalog['xyz']
run = db[-1]
run.primary.config['motor'].read()
```
This gives an xarray of device configuration parameters. I would like the same functionality using a databroker with a tiled server like this:

```python
db = from_uri("http://localhost:8000/api")
run = db[-1]
run.primary.config['motor'].read()
```

## Current Behavior
At the moment I get an error something like:
```
ClientError: 404: No such entry: ['a6c2c56d-cbea-40a0-a83f-bea66cc41552', 'baseline', 'config', 'ue48_m3'] http://172.17.9.22:8001/api/node/metadata/a6c2c56d-cbea-40a0-a83f-bea66cc41552/baseline/config/ue48_m3
```

## Steps to repeat

```python
from bluesky import RunEngine
from databroker.v2 import temp

## Set up env


RE = RunEngine({})
db = temp()
RE.subscribe(db.v1.insert)

RE(scan([det],motor,1,2,10))
db[-1].primary.config['det'].read()
```

## Context
I use the configuration parameters of the devices involved in a previous plan to restore the configuration if I want to repeat the plan. The ability to get to this through databroker is really useful.

## Your Environment
I am using:

```
tiled==0.1.0a77
databroker==2.0.0b11
```

danielballan:
Closes #745

This adds a test that exercises the bug and then a fix.

### DIFF
--- a/databroker/mongo_normalized.py
+++ b/databroker/mongo_normalized.py
@@ -1023,7 +1023,7 @@ def build_config_xarray(
         data_array = xarray.DataArray(columns[key], dims=dims, attrs=attrs)
         data_arrays[key] = data_array
     ds = xarray.Dataset(data_arrays)
-    return DatasetAdapter(ds)
+    return DatasetAdapter.from_dataset(ds)
 
 
 class MongoAdapter(collections.abc.Mapping, CatalogOfBlueskyRunsMixin, IndexersMixin):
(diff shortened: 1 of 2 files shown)


mrakitin (approved):
Tested locally from the PR branch, worked as expected with the same `det` object as in the tests, however, it failed when using the `ophyd.sim.det` (looks like a bug there, will be reported separately).

PR #751, opened 2023-01-04
Adjust for API change in tiled.

danielballan:
Breakage was in https://github.com/bluesky/tiled/commit/eb4700bbf3c30bfcde4088bf499c3901524099ad, first released in v0.1.0a76. Detected by CHX beamline staff.


tacaswell (approved):

mrakitin (approved):
I guess it will break for all beamlines, right? We can produce a new conda env pretty quickly with this fix released.

Issue #735, opened 2022-10-05
Add writable POST /documents path and reinstate db.insert


danielballan:

danielballan:
This is done.

PR #752, opened 2023-01-25
make a databroker podman image

gwbischof:


requirements-server.txt
```diff
@@ -1,4 +1,5 @@
 area-detector-handlers
+bson
```
danielballan: I think this is probably not what is wanted.

The `pymongo` pip package provides a library named `bson` but it's not the same as what `pip install bson` provides. If this was placed here in response to some `ImportError` it probably indicates that `pymongo` was not installed properly.

danielballan:
This works:

```
podman run -it -p 8000:8000 -e TILED_SINGLE_USER_API_KEY=secret -v ./example_config.yml:/deploy/config.yml databroker
Using configuration from /deploy/config.yml
INFO:     Started server process [1]
INFO:     Waiting for application startup.
OBJECT CACHE: Will use up to 10_095_076_147 bytes (15% of total physical RAM)
INFO:     Application startup complete.
INFO:     Uvicorn running on http://0.0.0.0:8000 (Press CTRL+C to quit)
```

danielballan:
With `docker-compose` (or `podman-compose`) we can start MongoDB in a container alongside Tiled-with-Databroker:

```
$ TILED_SINGLE_USER_API_KEY=secret docker-compose up
```

From a Python script we can use ophyd and bluesky to generate documents, inserts them directly into MongoDB using suitcase-mongo, and then get them out via tiled+databroker over HTTP.

```
$ python examples/write_and_read_back_data.py 
<Catalog {1, 2, 3, 1, 2, 3, 1, 2, 3}>
<BlueskyRun {'primary'} scan_id=3 uid='05719f46' 2023-02-01 12:10>
```

We can also access the data over HTTP from another (non-Python) client. Here I'm extracting the RunStop document from one of the Runs:

```
$ http :8000/api/v1/node/search/?api_key=secret | jq .data[].id  # list keys (RunStart uids)
"342dd68d-ee44-471f-b899-498ece232552"
"3ddedb92-3275-4813-b378-c8b5a11f5253"
"68c41576-9e10-433d-ad5f-99c32847838f"
"e5ec1df9-ec45-456f-930b-728cab846d4e"
"cb1dd8c9-3e92-4d91-934e-539bcc2928d4"
"f7f24dc4-5ff1-4187-a3e2-518dbef1d75a"
"cd0a360b-f604-44e0-afa1-71e3bc568382"
"cbb38927-80cb-4f3b-969e-6fc748754574"
"05719f46-296b-4b6d-9e26-c4b570addded"

$ http :8000/api/v1/node/metadata/05719f46-296b-4b6d-9e26-c4b570addded/?api_key=secret | jq .data.attributes.metadata.stop
{
  "run_start": "05719f46-296b-4b6d-9e26-c4b570addded",
  "time": 1675271424.9367063,
  "uid": "19c0cf84-0858-4161-bac4-845c17c124f0",
  "exit_status": "success",
  "reason": "",
  "num_events": {
    "primary": 3
  }
}
```

docker-compose.yml
```diff
@@ -0,0 +1,28 @@
...
+        source: ./example_config.yml
+        target: /deploy/config.yml
+    environment:
+      - TILED_SINGLE_USER_API_KEY=$TILED_SINGLE_USER_API_KEY
+    ports:
+      - 8000:8000
+  mongo:
+    image: docker.io/mongo:6.0.4
+    ports:
+      - 27017:27017
```
dylanmcreynolds: This is a little picky, but...some IT departments will ding you for exposing a database port. How hard would it be to add a third service, based on python, for the example code that contains bluesky dependencies? Then it could mount the example file, and one could run it with `podman-compose exec example-container python write_and_read_back_data.py`. Then that port wouldn't exposed AND the user could run without building their own env/conda environment.

Too much?

tacaswell: bluesky-pods has exactly this (but with terrible bash scripts).

Might be simpler to also stuff a minimal kafka cluster into the pod and expose that as a write end poin.

danielballan: This occurred to me to. As databroker already exposes a `GET /documents/{path}` endpoint, the simplest thing may be to add a `POST /documents/{path}` endpoint.

For production deployments, Kafka is the way to go, but for the simplest sandbox, sticking to HTTP would be a nice option to have. This would require some databroker development work though, so I had in mind doing this in a separate PR.

danielballan: OK, this actually wasn't that hard to get working:

```
$ http POST :8000/api/v1/documents/?api_key=secret <<< '{"name": "start", "document": {}}'  # invalid document!
HTTP/1.1 400 Bad Request
content-length: 41
content-type: application/json
date: Wed, 01 Feb 2023 19:33:41 GMT
server: uvicorn
server-timing: app;dur=399.9
set-cookie: tiled_csrf=--kJ0uuUvSwisPoEBqdnOyVQBrxzAUI0V2_OafoxlG0; HttpOnly; Path=/; SameSite=lax
set-cookie: tiled_api_key=secret; HttpOnly; Path=/; SameSite=lax

{
    "detail": "'uid' is a required property"
}


$ http POST :8000/api/v1/documents/?api_key=secret <<< '{"name": "start", "document": {"uid": "xxx", "time": 0}}'
HTTP/1.1 200 OK
content-length: 4
content-type: application/json
date: Wed, 01 Feb 2023 19:34:01 GMT
server: uvicorn
server-timing: app;dur=4.2
set-cookie: tiled_csrf=gM0DQIZzH6ITYOvDMbxD5pR1umab_sF5v8VhM9faRbs; HttpOnly; Path=/; SameSite=lax
set-cookie: tiled_api_key=secret; HttpOnly; Path=/; SameSite=lax

null

$ http :8000/api/v1/documents/xxx?api_key=secret
HTTP/1.1 200 OK
content-type: application/x-ndjson
date: Wed, 01 Feb 2023 19:35:29 GMT
server: uvicorn
server-timing: app;dur=4.4
set-cookie: tiled_csrf=uIcO4GdHdzZovXRroG0JmyMjeFKPbGkat-mHA-yHOg4; HttpOnly; Path=/; SameSite=lax
set-cookie: tiled_api_key=secret; HttpOnly; Path=/; SameSite=lax
transfer-encoding: chunked

[
    "start",
    {
        "time": 0,
        "uid": "xxx"
    }
]
```

danielballan: As of commit 3856ac7, the database is not exposed:

```
telnet 127.0.0.1 27017
Trying 127.0.0.1...
telnet: Unable to connect to remote host: Connection refused
```

dylanmcreynolds: That's very cool.

dylanmcreynolds (approved):

danielballan:
In response to @dylanmcreynolds' suggestion above that we should avoid exposing the MongoDB to the host network, I have rolled https://github.com/bluesky/databroker/issues/735 into this. In retrospect I should have handled that first (in its own PR) before attempting the `docker-compose` work here. But I hope this is still reviewable.

* Added `POST /documents` to provide a way to get documents into MongoDB via the service
* Modified the structure returned by `GET /documents` for consistency and alignment with [a current conversation in event-model](https://github.com/bluesky/event-model/issues/235#issuecomment-1405307703)
* Added a method to the Python databroker client `post_document` and refactored `db.insert` to use it internally. This is expected to incur a small performance penalty compared to direct insertion into MongoDB, but I feel confident that it is the correct thing to do. Separately, in production at facility scale, bluesky-kafka or other means can be used. But the basic built-in behavior should use the service, not reach around to an exposed database port.

mrakitin (commented):
It worked for me, but first, I had to build the `databroker` docker image locally (with `docker build .`) and to then tag it as `databroker:latest`. Should these steps be added to the instructions?

Also, it would be great to add the corresponding CI workflow. I am not sure if you'd like to run it on every commit though.

dylanmcreynolds:
@mrakitin Oh, interesting. I think the docker-compose.yml should be modified.

Instead of:
```yaml
image: databroker
```

we should use:
```yaml
build:
    context: .
 ```

This will tell dockercompose to build on first run, or any time you start docker-compose with the `---build` flag. Then you won't need to explicitly build outside of the docker-compose environment.

**Comment by danielballan** [comment]
Thanks for the fast review, all.

Yes, I meant to add a GHA workflow, same as we have for Tiled, but I forgot. Thanks for catching that @mrakitin.

I intended to change `image` to point to a tagged and built image on the GH container registry. @dylanmcreynolds’ suggestion would be slower on first build but more dynamic in a way that seems overall useful.

**Comment by danielballan** [comment]
As for how often to build: for tiled we build on merges to `main` and on tags. We do not build on every commit. That seems to be the right “volume”. Most people will use tagged images but it will be possible to work from bleeding edge.

**Comment by danielballan** [comment]
I switch to

```yaml
build:
    context: .
```

and confirmed that it works as @dylanmcreynolds describes, in a clean environment. I also added a GH workflow. We'll find out once it's merged into `main` whether it works as expected. I copied it directly from the tiled repo and there is nothing repo-specific about it, so I expect it will work.

mrakitin:
Thanks, Dan! I think in my message I rather meant building the image and tagging without push to make sure the integration test works (i.e. use the added workflow + set up the environment & run `examples/write_and_read_back_data.py`).

PR #753, opened 2023-02-09
Bump tiled minimum versions and base image.

danielballan:
This is not expected to affect the library itself at all. But the `CMD` on the base docker image has changed in `v0.1.0a84`, and we want to propagate that changed into a tagged databroker image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ghcr.io/bluesky/tiled:v0.1.0a83 as base
+FROM ghcr.io/bluesky/tiled:v0.1.0a84 as base
 
 FROM base as builder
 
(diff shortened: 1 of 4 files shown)


danielballan:
I built locally and tested it like so:

```
$ TILED_SINGLE_USER_API_KEY=secret podman-compose up
```

```
$ python examples/write_and_read_back_data.py 
<Catalog {1, 2, 3}>
<BlueskyRun {'primary'} scan_id=3 uid='05bf4478' 2023-02-09 12:44>
```

We can give some thought to how to integrate this acceptance test with CI. I think it's a bit heavy to run on every commit, but it's something we should get in the habit of running on each release, similar to EPICS' `acctst`.

dylanmcreynolds:
It would be really nice if there were some way to parametrize this across requirements*.txt and dockerfile...that's a lot of files to know to update.

danielballan:
It's true. I don't know of a way to do this, unfortunately, but I am open to ideas.

PR #756, opened 2023-02-17
Restore Broker.named('temp') in v2.

danielballan:
`Broker.named('temp')` is a convenience for making a Broker backed by an in-memory store. We disabled it when we were drafting v2 (raised `NotImplementedError`) but now we should restore it.


mrakitin (approved):
Looks good! Thanks for the quick fix. It will be useful for our online tutorial (used [here](https://github.com/bluesky/tutorials/blob/246641d712b7b6be3edb03e205818aa65669a5f7/bluesky-tutorial-utils/bluesky_tutorial_utils/beamline_configuration.py#L32)).

PR #758, opened 2023-03-17
Use a more robust ALL sentinel.

danielballan:
Needs a Tiled release xref https://github.com/bluesky/tiled/pull/421

### DIFF
--- a/databroker/mongo_normalized.py
+++ b/databroker/mongo_normalized.py
@@ -22,7 +22,7 @@ import pymongo.errors
 import toolz.itertoolz
 import xarray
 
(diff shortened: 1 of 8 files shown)


tacaswell (approved):
pending turning the crank on tiled.

danielballan:
The necessary change to Tiled has been released in v0.1.0a86, and I have updated the minimum version pin here.

tacaswell:
We might want to pin setuptools back to avoid the namespace package issue: https://github.com/pypa/setuptools/pull/3434

danielballan:
Hmm. If we handle that in event-model that might settle the issue for all relevant repos.

PR #763, opened 2023-05-03
Added distinct to MongoAdapter in experimental code

jmaruland:
This includes the implementation of nested `search()` queries before a `distinct()` request

### DIFF
--- a/databroker/experimental/server_ext.py
+++ b/databroker/experimental/server_ext.py
@@ -7,6 +7,7 @@ import uuid
 import time
 from pathlib import Path
 from datetime import datetime, timezone
+from collections import Counter
 
 import dask.dataframe
 import numpy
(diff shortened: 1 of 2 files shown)


danielballan:
This does separate queries for getting the list of distinct values and their counts. Given the experimental and temporary status of this code, that is good enough for now. For the record---in case we want it later---I will link this Stack Overflow post describing how to obtain both in a single query: https://stackoverflow.com/a/33418582/1221924

jmaruland:
With the previous commit (https://github.com/bluesky/databroker/pull/763/commits/c345fc5b47b41387bb61d67a6579c6536f3d01ad), I improved the way specs is requested with distinct in the mongo database and it returns a similar behavior that we can see in the base code of tiled and that is also referenced in https://github.com/bluesky/tiled/pull/436

danielballan:
Power-cycled to re-run tests. This was failing due to an [upstream issue in vcrpy](https://github.com/kevin1024/vcrpy/issues/688) that seems to have been resolved by their latest release.

danielballan (approved):

PR #766, opened 2023-05-18
Added distinct to MongoAdapater class

jmaruland:
Distinct is applied over the start and stop collection.
This implementation follows the same idea and is compatible with the structure that tiled is using already.

distinct checks for keys in metadata, for structure families and for specs while it remembers previous search queries to apply the analysis over specific child nodes and not the root node.
Example: `c.search(Key("x")=="a").search(Key("y")=="b").distinct("z", structure_families=True, specs=True, counts=True)`

### DIFF
--- a/databroker/mongo_normalized.py
+++ b/databroker/mongo_normalized.py
@@ -40,7 +40,7 @@ from tiled.adapters.utils import (
     tree_repr,
     IndexersMixin,
 )
-from tiled.structures.core import Spec
+from tiled.structures.core import Spec, StructureFamily
 from tiled.utils import import_object, OneShotCachedMap, UNCHANGED
 
 from .common import BlueskyEventStreamMixin, BlueskyRunMixin, CatalogOfBlueskyRunsMixin
@@ -216,6 +216,9 @@ class DatasetMapAdapter(MapAdapter):
         return True
 
 
+BLUESKYRUN_SPEC = Spec("BlueskyRun", version="1")
+
+
 class BlueskyRun(MapAdapter, BlueskyRunMixin):
     def __init__(
         self,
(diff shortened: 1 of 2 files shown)


danielballan:
Power-cycled to re-run tests. This was failing due to an [upstream issue in vcrpy](https://github.com/kevin1024/vcrpy/issues/688) that seems to have been resolved by their latest release.

danielballan:
The test failure on 3.7 is a familiar flaky test in one of the disused v0 backends. Ignoring.

danielballan:
I tried this interactively. It seems to work as expected.

```py
In [1]: from bluesky import RunEngine

In [2]: from ophyd.sim import det, motor

In [3]: from databroker import temp

In [4]: RE = RunEngine()

In [5]: c = temp().v2
OBJECT CACHE: Will use up to 10_095_076_147 bytes (15% of total physical RAM)

In [6]: def post_document(name, doc):
   ...:     c.post_document(name, doc)
   ...: 

In [7]: from bluesky.plans import scan, count

In [8]: RE.subscribe(post_document)
Out[8]: 0

In [9]: RE(count([det]))
Out[9]: ('3e4543ef-3eee-4fe2-9afc-1da37b790b1c',)

In [10]: RE(scan([det], motor, -1, 1, 3))
Out[10]: ('eee74219-b060-4328-be05-d477580967a9',)

In [11]: RE(scan([det], motor, -1, 1, 3))
Out[11]: ('5853c836-3e6e-454d-b158-d46ee52f0077',)

In [12]: RE(scan([det], motor, -1, 1, 3))
Out[12]: ('a797af5d-c4de-46f9-964f-c4523ac382b6',)

In [13]: c.distinct("plan_name", counts=True)
Out[13]: 
{'metadata': {'start.plan_name': [{'value': 'count', 'count': 1},
   {'value': 'scan', 'count': 3}]},
 'structure_families': None,
 'specs': None}
```

PR #769, opened 2023-06-02
Bump tiled version in image.

danielballan:
Bump the tiled version. No specific motivation, just keeping the image current.

I also removed the vestigial `docker/` directory. (We now keep the `Dockerfile` in the repo root, and have for some time.)


danielballan:
CI is not relevant to this change. Merging.

PR #783, opened 2023-08-25
Bump tiled version

danielballan:


danielballan:
This picks up a critical fix to zstandard decoding, https://github.com/bluesky/tiled/pull/556.

PR #784, opened 2023-09-11
Update store-analysis-results.rst

subindevs:
Fix: replaced import bluesky_live.run_builder import 'module_name' with from bluesky_live.run_builder import 'module_name'. Because repeated import statements causes syntax error

## Description
In the current documentation there are two code lines 
`import bluesky_live.run_builder import build_simple_run`  and `import bluesky_live.run_builder import RunBuilder`, which give syntax error. It should be `frombluesky_live.run_builder import build_simple_run`  and `frombluesky_live.run_builder import RunBuilder`

## How Has This Been Tested?
Copy pasted the code in the example to python note book which gives syntax error.

### DIFF
--- a/docs/source/how-to/store-analysis-results.rst
+++ b/docs/source/how-to/store-analysis-results.rst
@@ -29,7 +29,7 @@ analysis results, we have to do some of that works ourselves.
 
    .. code:: python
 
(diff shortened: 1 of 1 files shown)


danielballan (approved):

danielballan:
Thank you!

PR #785, opened 2023-09-11
Fix compat with tiled API changes.

danielballan:
## Description

Update usage of `ArrayAdapter`.

## Motivation and Context

Tiled made backward-compatible changes in https://github.com/bluesky/tiled/pull/537. The databroker adapter was not fully updated to account for this. The issue went un-caught because it only occurs on a codepath where a databroker handler returns a Dask Array. This is not tested.

## How Has This Been Tested?

I tested this interactively by looking at two pieces of real beamline data:
* https://tiled-demo.blueskyproject.io/ui/browse/bmm/raw/5631b9e1-b07d-4455-ae85-36e9a84ab7dc/primary/data/I0
* https://tiled-demo.blueskyproject.io/ui/browse/fxi/raw/1b0b4d73-6d87-43ab-8d62-ed035c51b9b4/primary/data/Andor_image

...using current `main` (broken) and this PR branch (currently deployed, works).

There _should_ be a unit test but that would actually be a pretty big addition to the current Databroker test harness, and we are in a rush to land this before the resumption of operations.

### DIFF
--- a/databroker/assets/tests/test_mongo_details.py
+++ b/databroker/assets/tests/test_mongo_details.py
@@ -12,16 +12,6 @@ def test_double_sentinel(fs_mongo):
         install_sentinels(fs.config, fs.version)
 
 
-def test_index(fs_mongo):
-    fs = fs_mongo
-    fs._create_datum_index()
-    indx = fs._datum_col.index_information()
-
-    assert len(indx) == 3
-    index_fields = set(v['key'][0][0] for v in indx.values())
-    assert index_fields == {'_id', 'datum_id', 'resource'}
-
-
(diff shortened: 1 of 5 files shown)


danielballan:
An unrelated recently-merged PR, https://github.com/bluesky/databroker/pull/638, broke `main`. That PR was so old that that the CI determination got stale. A lesson there is to re-run CI on old PRs.

The last 3 commits here clean that up.

mrakitin (approved):
The fix looks reasonable to me and the interactive tests demonstrating the fix are sufficient for now.

AbbyGi (approved):

PR #786, opened 2023-09-28
Bump base image.

danielballan:
Bumping Tiled base image, to get support for Awkward and to get support for a confirmation message on all Authenticators.


danielballan:
This only touched the `Dockerfile`, and the relevant workflow passed so I'm merging without waiting for the others.

danielballan:
Actually, this will be the first run since the tiled v0.1.0a107 release. Let's see how it goes.

Confirmed that v0.1.0a107 was installed on CI: https://github.com/bluesky/databroker/actions/runs/6343660075/job/17232059200#step:6:703

PR #795, opened 2024-01-31
Support updating metadata from Tiled client 🧱

### LINKED ISSUE #792: Support updating metadata from Tiled client
1. Add an `update_metadata` method on `databroker.mongo_normalzed.BlueskyRun`. This will enable updating the RunStart document and RunStop document. Most importantly, it will enable fixing erroneous `data_session`.
2. Add an `update_metadata` method on `databroker.mongo_normalzed.BlueskyEventStream`.  This will enable fixing EventDescriptors. Most importantly, it will enable fixing erroneous shape, data type.

Kezzsim:
This adds the ability for the raw blueskyRun and BlueskyEventStream objects to update their metadata, putting the onus on clients that invoke it to preserve existing metadata. Currently behaves like a `PUT` request, overwriting the existing data unless the client performs a `copy.deepcopy()` of the extant metadata dict.

Here is the code for performing such an update, keep in mind that the client should now be refactored with friendlier facilities for performing this action including HTTP `PATCH` requests like `rfc6902`. Potentially this would manifest as 
- update_metadata (alias)
- patch_metadata
- put_metadata

```
import tiled
import copy
c = from_uri("http://localhost:8000",api_key="secret")
metadata = copy.deepcopy(dict(c[154705].metadata))
metadata["start"]["num_points"] = 2
c[154705].update_metadata(metadata=metadata)
```

Fix #792

### DIFF
--- a/databroker/mongo_normalized.py
+++ b/databroker/mongo_normalized.py
@@ -276,18 +276,19 @@ class BlueskyRun(MapAdapter, BlueskyRunMixin):
     async def update_metadata(self, metadata=None, specs=None):
         if("start" not in metadata):
              raise NotImplementedError('update_metadata method not implemented')
-        elif(specs is not None):
+        elif(specs is None):
             raise NotImplementedError('Updating of specs is not yet supported.')
         start = metadata["start"]
-        # stop = metadata["stop"]
+        stop = metadata["stop"] if "stop" in metadata else None
         try:
             schema_validators[DocumentNames.start].validate(start)
-            # schema_validators[DocumentNames.stop].validate(stop)
+            if (stop is not None):
+                schema_validators[DocumentNames.stop].validate(stop)
         except ValidationError as err:
             raise
         # Update start
         self._serializer.update("start", metadata["start"])
-        # self._serializer.update("stop", metadata["stop"])
+        self._serializer.update("stop", metadata["stop"])
 
     @property
     def filler(self):
--- a/databroker/mongo_normalized.py
+++ b/databroker/mongo_normalized.py
@@ -62,6 +62,10 @@ from .queries import (
 )
 from .server import router
 
+from jsonschema import ValidationError
+
+from event_model import DocumentNames, schema_validators
+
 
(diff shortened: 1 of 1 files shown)


databroker/mongo_normalized.py
```diff
@@ -261,6 +272,20 @@ def metadata(self):
             transformed["stop"] = self.transforms["stop"](self._metadata["stop"])
         metadata = dict(collections.ChainMap(transformed, self._metadata))
         return metadata
+    
+    async def update_metadata(self, metadata=None, specs=None):
+        # if(specs is not None or "start" not in metadata):
```
danielballan: As written this silently ignore any input to `specs`. The "ignores" part of this is fine for the foreseeable future, but the "silently" part could conceivably give someone a bad day.

Can we get `if specs is None` and some kind of "Updating specs is not yet supported" error here?

databroker/mongo_normalized.py
```diff
@@ -440,6 +465,21 @@ def metadata(self):
             ]
         metadata = dict(collections.ChainMap(transformed, self._metadata))
         return metadata
+    
+    async def update_metadata(self, metadata=None, specs=None):
```
danielballan, line 490: This one should be working with `metadata["descriptors"]` and, correspondingly:

```py
for descriptor in metadata["descriptors"]:
    self.serializer.update("descriptor", descriptor)
```

A prerequisite is the add support for updating `"descriptor"` documents in `suitcase.mongo_normalized`.

databroker/mongo_normalized.py
Kezzsim: resolved in 01a556cc89b9b019bbd5bec2e044d5af39605665
So long as you meant `if specs is not None` which would mean, "If specs is present, let the user know they cannot update it"

databroker/mongo_normalized.py
Kezzsim: Support added in 63ed99bffc409add78f86afcfb66318f66216acd
Changes just need to be merged in `suitcase-mongo` and then CI can be run again to pass all tests related to it.

danielballan (approved):

PR #796, opened 2024-02-01
Pass validate_shape through new_variation.

danielballan:
Hotfix for SRX


danielballan:
Tested on `scan_id` 152689, 152690, 152691.

PR #803, opened 2024-04-13
Update dependencies to make tests pass

padraic-shafer:
## Description
Tests are failing in GitHub CI and locally on my laptop.

### Updates
* [Create indexes for mongo-legacy backend](https://github.com/padraic-shafer/databroker/commit/81d9c6374b63314c9ad604f902c1306348304a59)
* [Pin glue-core to keep glue.qglue](https://github.com/padraic-shafer/databroker/commit/5be59a74906eeb2b8ac7d5bcdde1f25bfae0ce4f)
* [Add .structure() to mongo_normalized adapters](https://github.com/padraic-shafer/databroker/commit/995cc95c20879550a144beae7bbb4439ef5f61af)

### Bonus
* [Keep .env secrets out of project repo](https://github.com/padraic-shafer/databroker/commit/451e4239753568dde0646d1d3a89de303ef2f024) 

## Motivation and Context
Currently tests are failing; these changes enable test suite to pass without loss of functionality.

## How Has This Been Tested?
Tested locally on MacOS (M2) with fresh instances of Mongo:(5, 6, 7) available on port 27017 (pulled from official docker hub mongo images) and with tiled image from `databroker/docker-compose.yml`.

PR #804, opened 2024-04-13
Accept numpy.int64 as databroker catalog key

padraic-shafer:
## Description
Databroker will now accept a `numpy.int64` value as a key for accessing scans in a databroker catalog, ~~as long as that value can be safely converted to a python int (int32). Otherwise a `KeyError` is raised.~~ Behavior of using the int-valued key is unchanged by this PR: The key is treated as a `scan_id` value if positive, or the `N`th-most recent scan index if negative. Suitable unit tests have been added to verify this.

## Motivation and Context
Databroker is frequently used in conjunction with pandas. By default pandas imports integer columns as `numpy.int64`; databroker then raises an exception when someone tries to use the int64-valued `scan_id` as a catalog key.

Although type conversion could be done manually by the calling code, this is such a common scenario that it would improve out-of-the-box user experience to handle this case within databroker.

~~### Note: Stacked PR~~
~~This PR builds upon outstanding PR https://github.com/bluesky/databroker/pull/803. Please wait until https://github.com/bluesky/databroker/pull/803 has been accepted before merging this PR.~~

## How Has This Been Tested?
Passes [linter](https://github.com/padraic-shafer/databroker/actions/runs/8674817684) and [unit tests](https://github.com/padraic-shafer/databroker/actions/runs/8674817685) in GH actions of forked repo.

_EDITED to clarify the intention that the int64-valued key be treated as an int-valued key; other than that, databroker behaves no differently._

### DIFF
--- a/databroker/tests/test_broker.py
+++ b/databroker/tests/test_broker.py
@@ -261,6 +261,13 @@ def test_int64_indexing(db_empty, RE, hw):
     with does_not_raise():
         db[integer64_key]
 
+    # Key is Nth-last scan
+    integer64_key = np.int64(-1)
+    assert not isinstance(integer64_key, int)
+    assert isinstance(integer64_key, numbers.Integral)
+    with does_not_raise():
+        db[integer64_key]
+
     # Key is too large
     integer64_key = np.int64(2**33)
     assert not isinstance(integer64_key, int)
--- a/databroker/tests/test_broker.py
+++ b/databroker/tests/test_broker.py
@@ -279,7 +279,7 @@ def test_int64_indexing(db_empty, RE, hw, key, expected):
 def test_large_int_indexing(db_empty, RE, hw, key, expected):
     """Integer-valued catalog key can exceed 32-bit values"""
     db = db_empty
-    RE.md["scan_id"] = 2*33
+    RE.md["scan_id"] = 2**33 - 1
     RE.subscribe(db.insert)
 
     uids = []
--- a/databroker/tests/test_broker.py
+++ b/databroker/tests/test_broker.py
@@ -240,6 +240,7 @@ def test_indexing(db_empty, RE, hw):
 
 
 def test_int64_indexing(db_empty, RE, hw):
+    """numpy.int64 can be used as a catalog key, if it is int-convertable"""
     db = db_empty
     RE.subscribe(db.insert)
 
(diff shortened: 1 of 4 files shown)


padraic-shafer:
@ambarb I believe this fixes the issue that you observed

danielballan (approved):

padraic-shafer:
Rebased this PR onto main to simplify the PR review.  No other changes since previous review.

databroker/tests/test_broker.py
```diff
@@ -237,6 +239,29 @@ def test_indexing(db_empty, RE, hw):
         db[-11]
 
 
+@pytest.mark.parametrize(
+    "key, expected",
+    (
+        (np.int64(1), does_not_raise()),  # Key is a Scan ID
+        (np.int64(-1), does_not_raise()),  # Key is Nth-last scan
+        (np.int64(2**33), pytest.raises(KeyError)),  # Key is too large
```
prjemian: What about `-np.int64(2**33)`?

prjemian (approved):
I'm fine with this code.  Suggest adding one more test, with appropriate handling.

tacaswell:
I don't understand why we have to check the value of the int:

```
python
Python 3.11.8 (main, Feb 12 2024, 14:50:05) [GCC 13.2.1 20230801] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> int(2**456)
186070713419675363980626894819329160794532188335953423432061490990243657757029868371504908982723472783555205531204141550984858016925351936
>>>
```

Python's native int does not have a fixed bit length (it is one of the BigInt versions...but I don't remember off the top of my head how it works).

databroker/tests/test_broker.py
padraic-shafer: Great suggestion!  I'll add it.

padraic-shafer:
> Python's native int does not have a fixed bit length (it is one of the BigInt versions...but I don't remember off the top of my head how it works).

You make a good point.  I thought I had a good reason for this, but some changes in the approach were made since then.  I'll go back and try without size cutoff under the current scenario.

databroker/tests/test_broker.py
padraic-shafer: Added this test case.  No changes to databroker code were needed.  Although as @tacaswell we should be able to bypass checking the size of the int-valued variable.

padraic-shafer:
@tacaswell Of course you are right about the variable-length int in python 3.  This makes this PR vastly simpler.

I've added a separate test to distinguish between key values that are large vs. out-of-range for the catalog (as opposed to overflow of the binary int representation).

I'll go back and update the PR description above.

padraic-shafer:
After the latest updates: Net changes to databroker code are very minimal now.  Test cases have been expanded.

PR #802, opened 2024-03-11
Add support for metadata patching, replacing, merging

hyperrealist:
https://github.com/bluesky/tiled/pull/688 depends on these changes

PR #811, opened 2024-06-05
Update for tiled v0.1.0b3

danielballan:
This updates databroker for this backward-incompatible change in how metadata update works. The method was renamed from `update_metadata` to `replace_metadata` to (1) be unambiguous that the input will fully replace any existing metadata and (2) leave space for more efficient update/patch methods in the future.


danielballan:
I forgot that #802 was already queued up. Closing.

PR #814, opened 2024-06-20
Factor out `bluesky-tiled-plugins` package

### LINKED ISSUE #812: Refactor Tiled plugins into separate package
Per discussion with @tacaswell:

In this same same repo, make a separate `bluesky-tiled-plugins` package that contains the objects in `databroker.client` and their associated entrypoint declarations. These are the objects like `BlueskyRun` and `CatalogOfBlueskyRuns` that provide an optional, upgraded Python interface to Bluesky metadata/data in Tiled.

Thus, it will only be necessary to import `databroker` when:

- Importing `databroker.Broker` for backward-compatible user code
- Running a Tiled server with `databroker.mongo_normalized`

New user code can operate with just `tiled` and the tiny `bluesky-tiled-plugins`. Lightweight and easy to explain to new users.

danielballan:
This creates a second Python package in the databroker repository, `bluesky-tiled-plugins`, with the "special client" objects `BlueskyRun`, `BlueskyEventStream`, and `CatalogOfBlueskyRuns`, as well as the custom query objects `PartialUID`, `ScanID`, and `TimeRange`. Quoting the README:

> For a user wishing to connect to a running Tiled server and access Bluesky data,
this package, along with its dependency `tiled[client]`, is all they need.

> The databroker package is only required if the user wants to use the legacy
`databroker.Broker` API.

This means it is no longer necessary to install `databroker` in the client environment unless they have legacy `databroker.Broker` code.

To be clear, the _server_ environment still needs `databroker.mongo_normalized`, the Tiled Adapter for MongoDB with Bluesky document collections.

---

This is a backward-compatible change. Databroker now has a dependency on `bluesky-tiled-plugins` and has shim modules that expose the moved objects at the original locations within the `databroker` package.

Closes #812


danielballan:
I have registered the `publish-pypi.yml` workflow in this repository as a ["pending" trusted publisher](https://docs.pypi.org/trusted-publishers/creating-a-project-through-oidc/) such that merge this PR should _create_ the `bluesky-tiled-plugins` package on PyPI.

bluesky-tiled-plugins/bluesky_tiled_plugins/__init__.py
```diff
@@ -0,0 +1,3 @@
+from ._catalog_of_bluesky_runs import CatalogOfBlueskyRuns  # noqa: F401
```
danielballan: Does this work with entrypoints? I vaguely recall that entrypoints have to point to the actual definition, not an import.

bluesky-tiled-plugins/bluesky_tiled_plugins/_version.py
```diff
@@ -0,0 +1,17 @@
+# file generated by setuptools_scm
+# don't change, don't track in version control
```
danielballan: This means you, @danielballan.

danielballan:
One transient failure, unrelated to the change:

```
FAILED databroker/assets/tests/test_write.py::test_multi_write_fail[sqlite] - TimeoutError: cursor call timed out
```

PR #816, opened 2024-06-23
Add mongoquery as a server dep.

danielballan:
This addresses a bug caught in prod, where the databroker container image is missing this dependency. It comes up in situations when a `ScanID` where is mixed with other query types.


padraic-shafer (approved):
Good catch

PR #817, opened 2024-08-09
Add support for using dtype_numpy as in ophyd-async

jwlodek:
## Description
Add support for using dtype_numpy as in ophyd-async

## Motivation and Context
ophyd-async uses `dtype_numpy` as an equivalent for the old `dtype_str`


danielballan (approved):
Looks good, thanks @jwlodek!

PR #829, opened 2024-10-25
Chunking and Padding

### LINKED ISSUE #827: Chunking
Chunking has a major effect on read performance.  For example, a 61GB dataset took over 1 hour to load through the tiled server with the default chunking.  After spending some time optimizing the chunking we were able to load the dataset in 6 minutes.  One issue with the default chunking is that the chunks often span multiple files. This means that multiple files need to be read and joined to get 1 chunk.

It would be nice to improve the default chunking so that the default chunks don't span files, but this may be impossible because `resource` and `datum` documents can refer to multiple files or parts of files.  I'll investigate this a little bit and see if there is anything we can do.

There is currently a mechanism to override the default chunks at collection time.  You can write the `chunk_list` into the descriptor document which describes the shape of all of the chunks. For example: `((3, 3, 3, 1), (3, 3, 3, 1))`.  This can be done by updating the `describe` method of the Ophyd device.  The problem with this mechanism is that we do not know the shape of the data until the collection is completed (for example, a scan could end early), and therefore we cannot know the `chunk_list` at collection time.

So, I propose that we update this mechanism to accept both a `chunk_list` and `chunk_shape`.  Where `chunk_shape` is a tuple that describes the shape of one chunk. For example: `(3, 3, 3, 1)`.  It is not always possible to know the `chunk_shape` ahead of time (a scan could end early, for example).  To address this, it should also support `auto` and `-1` as described in the Dask docs: https://docs.dask.org/en/latest/array-chunks.html#automatic-chunking

One concern with `chunk_shape` is that it may not perfectly partition the data, which will raise an exception. So we may want some extra logic to avoid this.

Using `chunk_shape` addresses this problem for new Bluesky scans.  We may also need to update existing scans. Here is an example of how we can do that: https://github.com/NSLS-II-TES/profile_collection/blob/main/startup/39-export_data.py#L10

### LINKED ISSUE #828: Padding
Databroker is currently zero-padding by default.

The padding rules are written in the doc-string here:
https://github.com/bluesky/databroker/blob/2550bd8c053139b58a9b6d6ea3f4ad0b9c0c48e6/databroker/mongo_normalized.py#L2207-L2215

Note: The doc-string says NaN-pad but this `numpy.pad` call is doing zero-padding because the default pad value is 0.
https://github.com/bluesky/databroker/blob/2550bd8c053139b58a9b6d6ea3f4ad0b9c0c48e6/databroker/mongo_normalized.py#L2243

I think that implicit zero-padding and NaN-padding could affect scientific results without users realizing it. NaN-padding is probably better because users will have to decide how to handle invalid data on their end, but there could be alignment issues if edge-padding.

For example: 

- Maybe a beamline is scanning the energy while capturing image data from a detector. They are recording the energy and taking an image for 10 different energy values.  For some reason, the detector fails to write the 5th image. So there are 10 energy measurements, but only 9 images.  The `default_validator` edge-pads so it would add 1 image of NaNs in the last position so that the data matches the `expected_shape`.  This would mean that images after the 4th image won't be aligned correctly with the energy measurement.

Implicit "ragged" NaN-padding or replacing the missing image with a NaN array, might be safe.
Currently, there is no ragged padding in `default_validate_shape`, I think it makes sense to add it. At SRX we just saw a couple of examples where one of the files was smaller than all of the others.  In the context of the previous example, this would be like only half of an image was written for one of the measurements.

Another challenge with NaN-padding is that it won't work with integer dtypes.  

One way to handle this is for users to override the `default_validator`, as SRX has done here: 
https://github.com/NSLS2/tiled-site-config/blob/2c0e441474862b5e5187f9247296b80f3ba22087/srx_patches.py#L38
This way they are in control of the way padding is done and they can ensure it is done in a way that is right for them and doesn't affect their results.

gwbischof:
Fixes #827
Fixes #828

### DIFF
--- a/databroker/mongo_normalized.py
+++ b/databroker/mongo_normalized.py
@@ -160,7 +160,9 @@ def structure_from_descriptor(descriptor, sub_dict, max_seq_num, unicode_columns
         numpy_dtype = dtype.to_numpy_dtype()
         if "chunks" in field_metadata:
             # If the Event Descriptor tells us a preferred chunking, use that.
-            suggested_chunks = tuple(tuple(chunks) for chunks in field_metadata["chunks"])
+            suggested_chunks = [tuple(chunk) if isinstance(chunk, list)
+                                else chunk for chunk in field_metadata['chunks']]
+            print(suggested_chunks, shape)
         elif (0 in shape) or (numpy_dtype.itemsize == 0):
             # special case to avoid warning from dask
             suggested_chunks = shape
--- a/databroker/mongo_normalized.py
+++ b/databroker/mongo_normalized.py
@@ -2205,7 +2205,7 @@ class BadShapeMetadata(Exception):
     pass
 
 
-def default_validate_shape(key, data, expected_shape):
+def default_validate_shape(key, data, expected_shape, uid=None):
     """
     Check that data.shape == expected.shape.
 
--- a/databroker/mongo_normalized.py
+++ b/databroker/mongo_normalized.py
@@ -2212,7 +2212,7 @@ def default_validate_shape(key, data, expected_shape):
     * If number of dimensions differ, raise BadShapeMetadata
     * If any dimension differs by more than MAX_SIZE_DIFF, raise BadShapeMetadata.
     * If some dimensions are smaller than expected,, pad "right" edge of each
-      dimension that falls short with NaN.
+      dimension that falls short with zeros..
     """
     MAX_SIZE_DIFF = 2
     if data.shape == expected_shape:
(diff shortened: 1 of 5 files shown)


databroker/mongo_normalized.py
```diff
@@ -2241,6 +2242,11 @@ def default_validate_shape(key, data, expected_shape):
         else:  # margin == 0
             padding.append((0, 0))
     padded = numpy.pad(data, padding, "edge")
+
+    logger.warning(f"The data.shape: {data.shape} did not match the expected_shape: "
```
gwbischof, line 2250: Is there a way that I can add the run uid to this warning?
`default_validate_shape` doesn't have `self` so something like this wont work.
```
f"RunStart UID: {self._run.metadata()['start']['uid']!r}"
```

danielballan: The signature of `validate_metadata_shape` could be modified to accept the `uid` (last, and optionally, for back-compat).

danielballan (approved):

PR #830, opened 2024-11-07
Fix dtype deserialization

### LINKED ISSUE #820: Reading configuration of devices in a run doesn't work with tiled server
This issue seems somewhat similar to #745

## Expected Behavior
With databroker backed by intake I am able to access the configuration parameters of devices like this:
```python
db = databroker.catalog['xyz']
run = db[-1]
run.primary.config['det'].read()
```
This gives an xarray of device configuration parameters. I would like the same functionality using a databroker with a tiled server like this:

```python
db = from_uri("http://localhost:8000/api")
run = db[-1]
run.primary.config['det'].read()
```
## Current Behavior
I get the following error when I try to access the config data:
```
HTTPStatusError: Server error '500 Internal Server Error' for url 'http://localhost:8000/api/v1/metadata/df8df11a-28f6-4e43-af52-e9c5722d271a/primary/config/det'
For more information, server admin can search server logs for correlation ID None.
```

## Steps to Reproduce (for bugs)
```python
from bluesky import RunEngine

from bluesky.plans import scan
from ophyd.sim import det,motor

## Set up env


RE = RunEngine({})
db = from_uri("http://localhost:8000",api_key="secret")

def post_document(name,doc):
    db.post_document(name, doc)

RE.subscribe(post_document)

RE(scan([det],motor,1,2,10))


db[-1].primary.config['motor'].read() # this somehow returns config data

db[-1].primary.config['det'].read() # this fails
```

## Context
I want to use the data inside the configuration parameters to populate a Nexus File . Similarly to the Issue #745 the data is still accessible under:
```python
run.primary.data.metadata['descriptors'][0]['configuration']['det']['data']
```

## Your Environment
```
databroker==2.0.0b46
tiled==0.1.0b7
```

danielballan:
Closes #820

### DIFF
--- a/databroker/mongo_normalized.py
+++ b/databroker/mongo_normalized.py
@@ -77,7 +77,8 @@ def _try_descr(field_metadata):
     if descr:
         if len(descr) == 1 and descr[0][0] == "":
             return None
-        dtype = StructDtype.from_numpy_dtype(numpy.dtype(descr))
+        numpy_dtype = numpy.dtype([tuple(field) for field in descr])
+        dtype = StructDtype.from_numpy_dtype(numpy_dtype)
         if dtype.max_depth() > 1:
             raise RuntimeError(
                 "We can not yet cope with multiple nested structured dtypes.  "


Bilchreis:
Okay so not only the incoming dtype needs to be deserialized correctly, but in some cases also the structured arrays. So the changes needed are a little more extensive. 

1. Shape Validation (`mongo`):
https://github.com/danielballan/databroker/blob/6c7c26a8752709ee102e2551c2a12bd4d671a9f2/databroker/mongo_normalized.py#L932-L946

`numpy.asarray(item)`: the conversion of item to an ndarray has to be done with the appropriate numpy dtype,  from the `data_key['dtype_descr']` if it exists, and each structured array element has to be turned into a  tuple

2. Config xarray: 
https://github.com/danielballan/databroker/blob/6c7c26a8752709ee102e2551c2a12bd4d671a9f2/databroker/mongo_normalized.py#L1081

`columns[key] = numpy.array(column, dtype=numpy_dtype)`

if `numpy_dtype` is a structured array, the data in column needs to be turned into a list of lists of tuples. 

How would you like to proceed, sould I open my own PR or do you want to implement these changes?

PR #831, opened 2025-01-16
Address bit rot

danielballan:
- Bumped minimum Tiled version to latest
- Tiled AccesControlPolicy methods grew an additional required parameter, `path_parts`, and changed from sync to async.
- Tiled auth-related parameters changed.
- Python 3.8 is EOL. Python 3.12 was added. Python 3.13 was not yet added, as Tiled is not testing against it 3.13. (Last we checked some of our dependencies were still catching up.)

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -16,7 +16,7 @@ jobs:
         - 27017:27017
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11"]
+        python-version: ["3.9", "3.10", "3.11", "3.12"]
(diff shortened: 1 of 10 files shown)


danielballan:
Power-cycled because the Unit Tests workflow was disabled by GH due to many consecutive failures.

danielballan:
One transient failure on the last run:

```
FAILED databroker/assets/tests/test_fs.py::test_insert_funcs[column_hdf5-insert_syn_data] - TimeoutError: cursor call timed out
```

but previous runs had no problem here, so I'll plow ahead...

PR #835, opened 2025-02-13
Adjust to API change in FullText query.

danielballan:
The `case_insensitive` attribute was removed from `FullText` a little more than a year ago in https://github.com/bluesky/tiled/commit/eb79bdb8aea1dec2601c29344e22deedc9f87799#diff-39c6c64d38798eb3df5586ef8342fc2f7b947a43b8f762368f0ef4d2e8545a8d. This needs to be updated to match.


danielballan:
Needs unit test; this API change should have been caught by tests.

danielballan:
Ah, mongomock does not support `$text` queries, so it would be a lift to test this.

PR #837, opened 2025-02-20
feat: serialize named tuples as dicts

maffettone:
Take advantage of [dependent PR in Tiled](https://github.com/bluesky/tiled/pull/895) to serialize NamedTuple like objects as dictionaries. 

## Description

Add function to be consumed as default serialization. 

## Motivation and Context

This default functionality was removed from `orjson` in version 3.10, and causes problems with Bluesky use (e.g., hklpy).

### DIFF
--- a/bluesky-tiled-plugins/bluesky_tiled_plugins/catalog_of_bluesky_runs.py
+++ b/bluesky-tiled-plugins/bluesky_tiled_plugins/catalog_of_bluesky_runs.py
@@ -147,7 +147,15 @@ class CatalogOfBlueskyRuns(Container):
 
     def post_document(self, name, doc):
         link = self.item["links"]["self"].replace("/metadata", "/documents", 1)
-        response = self.context.http_client.post(
-            link, content=safe_json_dump({"name": name, "doc": doc}, default=self._namedtuple_asdict_default)
-        )
+        try:
+            response = self.context.http_client.post(
+                link, content=safe_json_dump({"name": name, "doc": doc}, default=self._namedtuple_asdict_default)
+            )
+        except TypeError as e:
+            if "got an unexpected keyword argument" in str(e):
+                response = self.context.http_client.post(
+                    link, content=safe_json_dump({"name": name, "doc": doc})
+                )
+            else:
+                raise
         handle_error(response)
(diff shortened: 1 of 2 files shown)


maffettone:
Tagging @ambarb for visibility.

bluesky-tiled-plugins/bluesky_tiled_plugins/catalog_of_bluesky_runs.py
```diff
@@ -138,9 +138,24 @@ def v1(self):
             self._v1 = Broker(self)
         return self._v1
 
+    @staticmethod
+    def _namedtuple_asdict_default(obj):
+        """Convert NamedTuple to dict before serialization."""
+        if isinstance(obj, tuple) and hasattr(obj, "_asdict"):  # NamedTuple check
+            return obj._asdict()
```
tacaswell, line 145: do dataclasses as well?

maffettone: Seemed like a great idea, but `orjson` handles data classes fine, but treats NamedTuples as ambiguous (is it an array or a dictionary).
Fun rabbit hole for the morning.

danielballan:
I appreciate the effort to isolate and target this fix. Unfortunately, as we look toward migrating the facility off of `mongo_normalized` during 2025, this fix would be short-lived.

At the moment, databroker registers a custom endpoint with the Tiled FastAPI Router, `POST /documents/{path}`, which you are customizing here, and `blueksy_tiled_plugins.CatalogOfBlueskyRuns.post_document` calls it. But in the future (and already at HEX) the `bluesky.callbacks.tiled_writer.TiledWriter` will be used, and it will send metadata from the start document at the vanilla `POST /metadata` endpoint.

Other options:
- Make Tiled tolerate `namedtuple` and perhaps other by default. (I don't _love_ this. I think orjson is right that there is no unambigous JSON serialization of `namedtuple`.)
- Merge this fix, and add tolerant JSON serialization into `TiledWriter` as well. Tiled would need a way to expose this in `tiled.client.container.Container.new()`.
- In the Device, use a serializable type.

maffettone:
Coming at this a little later than I'd like. On a deeper look into `tiled.client.container.Container` I think option 3 is the right approach. Anything else would have to change the signature of a `new` and a number of convenience methods that pass through to `new`.
If we are willing to accept the standard that `orjson` provides as correct (named tuple is ambiguous and one should be explicit for serialization), we should leave that up to the client and device. In this case `hklpy`. 

In other words, fewer changes would be required to `hklpy` than to a concerted effort of `tiled`, `bluesky`, and `databroker` simultaneously, so that seems like the correct route forward.

maffettone:
I'm comfortable closing this and the [Tiled PR](https://github.com/bluesky/tiled/pull/895) if in agreement.

PR #838, opened 2025-02-28
Fix access to timestamps in external data_keys

danielballan:
An event document containing a `datum_id` still has the associated timestamp inline. Don't try to fill it.

Unit test reproduces a bug found on the floor and fails, next commit fixes the failure.


danielballan:
The only error was one that is commonly flaky:

```
FAILED databroker/assets/tests/test_asset.py::test_bulk_datum_register_table[column_hdf5-dict] - TimeoutError: cursor call timed out
```

nmaytan (approved):

PR #839, opened 2025-03-05
fix imsave : change it with imwrite

skarakuzu:


skarakuzu:
@danielballan It turned out that there is only one case which used `imsave` .

danielballan:
Thanks! Superseded by PR with the same commit in https://github.com/danielballan/databroker/pull/14

PR #847, opened 2025-05-07
add:  root map dict arg to cli tool to extend

maffettone:
## Description

Adds a parser (so that typer can cope with dictionaries) and optional argument to pass a dict as a string to be evaluated.

## Motivation and Context

Root map made this unworkable for rsoxs.

## How Has This Been Tested?

- with correct json dict '{"key": "val"}'
- with incorrect dict
- with not dictionary
- with empty string

### DIFF
--- a/databroker/cli.py
+++ b/databroker/cli.py
@@ -1,4 +1,4 @@
-import ast
+import json
 from typing import List, Optional
 
 import tiled.client
@@ -27,12 +27,14 @@ def parse_dict_arg(arg):
     if arg is None or arg.strip() == '':
         return {}
 
-    result = ast.literal_eval(arg)
-    if isinstance(result, dict):
-        return result
-    else:
-        raise ValueError("Parsed value is not a dictionary.")
-
+    try:
+        result = json.loads(arg)
+        if isinstance(result, dict):
+            return result
+        else:
+            raise ValueError("Parsed value is not a dictionary.")
+    except json.JSONDecodeError as e:
+        raise ValueError(f"Invalid JSON string: {arg}") from e
 
 @admin_app.command("shape-fixer")
 def shape_fix(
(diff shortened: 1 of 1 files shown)


databroker/cli.py
```diff
@@ -17,6 +17,23 @@
...
+    """
+    Parse a string like '{"key1": "value1", "key2": 2}' into a dictionary.
+
+    Returns:
+        dict: The parsed dictionary or an empty dict if arg is None or empty.
+    """
+    if arg is None or arg.strip() == '':
+        return {}
+
+    result = ast.literal_eval(arg)
```
danielballan: Using a json parser is a bit safer and more mainstream here.

danielballan (commented):
Thank you! One small comment.

databroker/cli.py
```diff
@@ -67,6 +85,10 @@ def shape_fix(
         typer.echo(f"Limited to first {limit} BlueskyRuns only")
         items = items[:limit]
 
+    root_map = root_map if root_map is not None else {}
+    root_map = getattr(adapter, "root_map", {}) | root_map
```
danielballan, line 92: Should this be `or` instead of `|`?

```python
>>> None or 1
1
>>> None | 1
  File "<stdin>", line 1, in <module>
TypeError: unsupported operand type(s) for |: 'NoneType' and 'int'
```

maffettone: That should work as an [in-place `update` operation (or union)](https://peps.python.org/pep-0584/). The first line ensures you're working with dicts.

databroker/cli.py
maffettone: Good point. Squared away in https://github.com/bluesky/databroker/pull/847/commits/fb8c69f9698a7abe5831ffdc67fe43ce6ff17bad

genematx (approved):

databroker/cli.py
danielballan: Oh right, not used to that syntax yet. Cool.

PR #848, opened 2025-05-08
add: replace metadata signature for BlueskyEventStream

maffettone:
## Description

See below. 

## Motivation and Context

The `replace_metadata` signature was updated for `BlueskyRun`, but not for `BlueskyEventStream`. See [recent release](https://github.com/bluesky/databroker/compare/v2.0.0b60...v2.0.0b61) and 7f4c901.

Was found when trying to update chunking for RSOXS. 
```
  File "/opt/venv/lib/python3.12/site-packages/tiled/server/router.py", line 1480, in patch_metadata
    await entry.replace_metadata(
          ^^^^^^^^^^^^^^^^^^^^^^^
TypeError: BlueskyEventStream.replace_metadata() got an unexpected keyword argument 'drop_revision'
```

### DIFF
--- a/databroker/mongo_normalized.py
+++ b/databroker/mongo_normalized.py
@@ -502,7 +502,9 @@ class BlueskyEventStream(MapAdapter):
     def key(self):
         return self._metadata["descriptors"][0]["name"]
 
-    async def replace_metadata(self, metadata=None, specs=None):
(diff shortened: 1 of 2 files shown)


danielballan:
I caught this other one because it has test coverage; evidently this path does not. It would probably be pretty easy to extend the existing test.

maffettone:
@danielballan Good point, I extended the test to target the stream

PR #853, opened 2025-08-13
Adjust to changed tiled `get_entry` internals

danielballan:
Databroker extends tiled's HTTP API with a `/documents` endpoint. In tiled v0.1.0b30, `get_entry` was changed from a FastAPI Dependency to a normal async function. This PR updates to adjust for that, and bumps the minimum version of tiled.


danielballan:
Thanks, @genematx!

PR #856, opened 2025-08-20
Update metadata fixup

danielballan:
An xfail-ed tests caused us to miss a regression. When the `MongoAdapter` is configured with an access policy (`authz_shim`), updating metadata raised an error. This PR includes a unit test that reproduced the error:

```
FAILED databroker/tests/test_broker.py::test_update[mongo-tiled-False] - NotImplementedError: Updating access_blob on MongoDB-backed data is not supported.
FAILED databroker/tests/test_broker.py::test_update[mongo-tiled-True] - NotImplementedError: Updating access_blob on MongoDB-backed data is not supported.
```

and then a commit with a fix.

It also fixes the confusing mistake within the test itself, which seems to be the that caused us to xfail this test in the first place.

### DIFF
--- a/databroker/mongo_normalized.py
+++ b/databroker/mongo_normalized.py
@@ -300,7 +300,7 @@ class BlueskyRun(MapAdapter):
         return metadata
 
     async def replace_metadata(self, metadata=None, specs=None, access_blob=None, drop_revision=False):
-        if access_blob:
+        if access_blob and (access_blob != self.access_blob):
             raise NotImplementedError("Updating access_blob on MongoDB-backed data is not supported.")
         if drop_revision:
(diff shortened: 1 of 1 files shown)


danielballan:
Test failures...

```
FAILED databroker/tests/test_broker.py::test_fill_and_multiple_streams[sqlite-tiled-False] - KeyError: 'detfs2'
FAILED databroker/tests/test_broker.py::test_fill_and_multiple_streams[sqlite-tiled-True] - KeyError: 'detfs2'
= 2 failed, 765 passed, 92 skipped, 8 deselected, 47 xfailed, 26 xpassed, 60 warnings in 557.56s (0:09:17) =
```

are due to a known regression in bluesky, fixed in https://github.com/bluesky/bluesky/pull/1942 but not yet released.

PR #858, opened 2025-08-26
FIX: getting columns from "composite" table

genematx:
Allows a user to query columns from a table in "composite" container from the `BlueskyRun` container directly, e.g.
`c[123456].v2['primary/data/det']`; `c[123456].v2['primary/data']['det']` and `c[123456].v3['streams/primary/internal/det']` always worked as expected.


danielballan:
Power-cycled to re-run CI after #859 was merged to fix the Dockerfile issue unrelated to this PR.

danielballan:
One known-flaky test failure on one build:

```
FAILED databroker/assets/tests/test_fs.py::test_root[column_hdf5] - TimeoutError: cursor call timed out
```

Merging.

PR #869, opened 2025-10-29
Additional Minor Fixes in Consolidators and TiledWriter

genematx:
1. Explicitly declare allowed kwargs for different Consolidator types;
2. Fix parsing of structural dtypes in the descriptor;
3. Reference `spec` from Resource document in Datums to set necessary parameters in `RunNormalizer`;
4. Remove duplicate definitions of classes that can be imported from Tiled;
5. Refactor `get_adapter` (used only internally).

### DIFF
--- a/bluesky-tiled-plugins/bluesky_tiled_plugins/consolidators.py
+++ b/bluesky-tiled-plugins/bluesky_tiled_plugins/consolidators.py
@@ -1,50 +1,16 @@
 import collections
 import dataclasses
-import enum
 import os
 import re
 import warnings
-from typing import Any, Literal, Optional, Union, cast
+from typing import Literal, Union, cast
 
 import numpy as np
 from event_model.documents import EventDescriptor, StreamDatum, StreamResource
 from tiled.mimetypes import DEFAULT_ADAPTERS_BY_MIMETYPE
 from tiled.structures.array import ArrayStructure, BuiltinDtype, StructDtype
-
-
-class StructureFamily(str, enum.Enum):
-    array = "array"
-    awkward = "awkward"
-    container = "container"
-    sparse = "sparse"
-    table = "table"
-
-
-class Management(str, enum.Enum):
-    external = "external"
-    immutable = "immutable"
-    locked = "locked"
-    writable = "writable"
-
-
-@dataclasses.dataclass
-class Asset:
-    data_uri: str
-    is_directory: bool
-    parameter: Optional[str]
-    num: Optional[int] = None
-    id: Optional[int] = None
-
-
-@dataclasses.dataclass
-class DataSource:
-    structure_family: StructureFamily
-    structure: Any
-    id: Optional[int] = None
-    mimetype: Optional[str] = None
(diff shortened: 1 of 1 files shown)


bluesky-tiled-plugins/bluesky_tiled_plugins/tiled_writer.py
```diff
@@ -467,6 +471,11 @@ def stream_datum(self, doc: StreamDatum):
 
     def datum(self, doc: Datum):
         doc = copy.copy(doc)
+        # Mark the Datum document with the spec of the corresponding Resource, if known
```
danielballan, line 474: What does this achieve? Is this a performance optimization to avoid doing a chained lookup Datum -> Resource -> Spec?

genematx: In short, yes.
In a way, it is just a shortcut to make parsing `datum_kwargs` easier. Since kwargs are not supported in `StreamDatums`, but are a part of the old `Datum`, they need to be converted and attributed the `Resource` (`StreamResource`); this is done in `patch_datum` functions -- something that is not a part of the usual `MongoNormalized`, but is necessary to ingest the old data -- the caller can be passed to the `TiledWriter` and normally would be applied to the Datum doc immediately after it is received (without looking up the associated `Resource`). The meaning of `datum_kwargs` may in general depend on the spec, for example:

```
        if spec == "XIA_XMAP_HDF5":
            # ISSXIAHDF5HandlerLegacy
            if kwargs["data_type"] != "roi":
                raise ValueError("XIA_XMAP_HDF5 only supports data_type='roi'")
            roi_num = kwargs["roi_num"]  # ROI number, 0-indexed
            channel = kwargs["channel"]  # Channel number, 1-indexed
            kwargs["dataset"] = "/entry/data/data"
            kwargs["slice"] = f"(:, {channel-1}, {roi_num})"
            kwargs["squeeze"] = True

        elif kwargs["data_type"] == "image":
            kwargs["dataset"] = "/entry/data/data"

        elif kwargs["data_type"] == "spectrum":
            # databroker.assets.handlers.Xspress3HDF5Handler
            kwargs["dataset"] = "entry/instrument/detector/data"
            channel = kwargs["channel"]     # Assume that 'channel' is always present
            kwargs["slice"] = f"(:,{channel-1},:)"
            kwargs["squeeze"] = True
```
(I wouldn't be surprised if some specs assume that `channel` is numbered as 0-based and others as 1-based, for example).
I don't like passing spec "in-band" either, and I think, in principle, it is possible to do the lookup, but then this would complicate the `patch_datum` signature (which would require `spec`) and make it different from other `path_xxx` functions. Since `spec` in `datum_kwargs` is used only internally here and then dropped (along with the entire document), I thought it was okay.

genematx: First, I was doing this in a subclassed `RunNormalizer` for ISS and QAS, but then decided to move it to the main class, to make it simpler.

PR #873, opened 2025-10-30
Restructure

genematx:
Restructure the repository to separate reading `clients` and `writing` utilities.

### DIFF
--- a/databroker/document.py
+++ b/databroker/document.py
@@ -1,7 +1,7 @@
 # This module is now a back-compat shim. The objects are defined in a
 # separate package, bluesky_tiled_plugins, that resides in the same
 # git repository as databroker.
-from bluesky_tiled_plugins.document import (  # noqa: F401
+from bluesky_tiled_plugins.clients.document import (  # noqa: F401
     Datum,
     DatumPage,
     Descriptor,


bluesky-tiled-plugins/pyproject.toml
```diff
@@ -39,9 +39,9 @@ classifiers = [
 dynamic = ["version"]
 
 [project.entry-points."tiled.special_client"]
-CatalogOfBlueskyRuns = "bluesky_tiled_plugins.catalog_of_bluesky_runs:CatalogOfBlueskyRuns"
-BlueskyRun = "bluesky_tiled_plugins.bluesky_run:BlueskyRun"
-BlueskyEventStream = "bluesky_tiled_plugins.bluesky_event_stream:BlueskyEventStream"
+CatalogOfBlueskyRuns = "bluesky_tiled_plugins:CatalogOfBlueskyRuns"
```
danielballan: I have noticed that entrypoints must target the module where the object is _defined_. Targeting re-exported locations does not work. Perhaps a comment should be added here to note that.

genematx: It passed tests on my laptop, but maybe it's version specific. Fixed to make the paths absolute.

PR #875, opened 2025-11-04
Make validation optional for all data sources

genematx:
When writing to Tiled, a consolidator has an option to validate the data structure before putting the `data_source` (typically, this is faster than post-factum updates). However, this validation requires access to the files on the client-side. To avoid running into permission errors when the client may not have reading permissions on the server filesystem, this PR makes validation optional globally (as a parameter passed to TiledWriter) instead of relying on in-band parameters.


danielballan (approved):
This is much cleaner than injecting a `_validate` parameter into `sres_parameters`. :+1:

danielballan:
Moved to https://github.com/bluesky/bluesky-tiled-plugins/pull/2

PR #877, opened 2025-11-18
Raise an error if validation fails

genematx:
This ensures that if data validation in TiledWriter fails for any reason (e.g. missing external file), a specific error is raised. This  relates only to the validation verformed by TiledWriter itself (if `validate=True`, which is False by default).


danielballan:
This change was applied in the new repo where this code moved to. https://github.com/bluesky/bluesky-tiled-plugins/pull/1

PR #878, opened 2025-11-19
Deprecated databroker, and split bluesky-tiled-plugins into separate repo

danielballan:
- The bluesky-tiled-plugins package is [in its own repo](https://github.com/bluesky/bluesky-tiled-plugins).
- The databroker docs are pared down, to reduce maintenance cost.
- A deprecation notice, modeled closely on the text in https://github.com/bluesky/website-beta/pull/1, is added.

## Description

This removes bluesky-tiled-plugins and associated tooling configuration.
That package now lives in its own repo rather than sharing a repo
with databroker. (Note: It has _always_ been a separate package. That
has not changed. Only the location of the code changed.)

Also, this re-instates documentation-publishing, which has been missing.
The documentation has rotted _substantially_ as we prioritized resources
for Tiled deployment and deployment.

I have removed all tutorials, how-to guides, and explanations, which were
overwhelmingly broken and/or offering dated advice. I have retained
the API reference and added a deprecation notice in the README, which
also appears at the top of the index page for the docs.


tacaswell (approved):

padraic-shafer (approved):

mrakitin (approved):
Looks good to me.

PR #880, opened 2026-05-27
Use read_block when reading slices of the entire arrays

genematx:
Use `read_block` methods for reading slices of array data.

## Motivation and Context
When reading slices of big arrays, Tiled by default calls the `.read(slice=...)` method, which causes databroker to fetch data for the entire array, event hough most of it may be discarded during slicing. These changes allow the server to only fetch necessary chunks.

## How Has This Been Tested?
Utility functions are tested in Tiled. Full workflow is tested in a dev set-up with mounted MongoDB.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ghcr.io/bluesky/tiled:0.1.0-b36 as base
+FROM ghcr.io/bluesky/tiled:0.2.11 as base
 
 USER root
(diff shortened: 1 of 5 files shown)


danielballan (approved):
Tested against the production MongoDB

danielballan:
Oh, better bump minimum tiled version first!

danielballan (approved):

